PR #3, opened 2015-05-13
Fix assertion failure for 'effect E _ ->'

yallop:
Before:

```
# try perform E with effect E _ -> ();;
Fatal error: exception File "bytecomp/translcore.ml", line 948, characters 9-15: Assertion failed
```

After:

```
# try perform E with effect E _ -> ();;
- : unit = ()
```


kayceesrk:
Thanks!

PR #6, opened 2015-05-23
Fix error on IA32

zakki:
* minor heap size overflow in caml_init_domain
* caml_mem_round_up_pages type missmatch


kayceesrk:
Thanks!

PR #11, opened 2015-06-08
Make physical equality aware of promotion duplication.

### LINKED ISSUE #7: Unhandled raised when two domains use single function
The following code works as expected

```ocaml
effect Print : string -> unit

let run thunk =
  match thunk () with
  | () -> ()
  | effect (Print s) k -> print_endline s; flush stdout; continue k ()

let thunk () =
  let rec write x =
    Printf.printf "domain %d %s\n" (Domain.self ()) x;
    flush stdout;
    perform (Print x);
    Unix.sleep 1;
    write x in

  (* Same as [write]. *)
  let rec write2 x =
    Printf.printf "domain %d %s\n" (Domain.self ()) x;
    flush stdout;
    perform (Print x);
    Unix.sleep 1;
    write2 x in

  Domain.spawn (fun () -> run (fun () -> write "first"));
  write2 "second"

let () = run thunk
```

but when I remove `write2` and replace `write2 "second"` by `write "second"` then the exception `Unhandled` is raised when performing effect `Print "second"`.

kayceesrk:
Promoting an object duplicates the object in the major heap and uses write barriers to ensure that the copies are in sync. However, physical equality breaks due to this duplication. Physical equality is essential for pattern matching polymorphic and extensible variants. See #7 for example. The fix ensures that the major heap object is always used when comparing promoted objects.

This fix is certainly not ideal since it adds a branch to every integer equality! We might perhaps add a special instruction for physical equality. This PR fixes #7.

### DIFF
--- a/byterun/interp.c
+++ b/byterun/interp.c
@@ -1061,8 +1061,20 @@ value caml_interprete(code_t prog, asize_t prog_size)
       accu = (value)((((intnat) accu - 1) >> Long_val(*sp++)) | 1); Next;
 
 #define Integer_comparison(typ,opname,tst) \
-    Instruct(opname): \
-      accu = Val_int((typ) accu tst (typ) *sp++); Next;
+    Instruct(opname): { \
+      if (Is_long (accu) || Is_long(*sp)) { \
+        accu = Val_int((typ) accu tst (typ) *sp++); \
+        Next; \
+      } \
+      if (Is_promoted_hd(Hd_val(accu)) && Is_young(accu)) \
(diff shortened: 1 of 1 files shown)


bluddy:
Yeah we definitely need a new primitive for physical equality.

stedolan:
Looks good. Will add for now, revert if we decide to go for promotion via eagar evacuation of objects from the local heap.

kayceesrk:
Thanks! We really should do eager evacuation.

Issue #13, opened 2015-06-13
Segmentation fault when assigning function to reference in different thread


radekm:
The following program

```ocaml
effect Yield : unit

let () =
  let kk = ref (fun () -> failwith "continuation") in
  begin match
    print_endline "0";
    perform Yield;
    print_endline "1";
    Unix.sleep 1;
    perform Yield;
    print_endline "2"
  with
    | () -> print_endline "done"
    | effect Yield k ->
        print_endline "before kk assig";
        kk := (fun () -> continue k ());
        print_endline "after kk assig"
  end;
  print_endline "before spawn";
  Domain.spawn (fun () -> !kk ());
  print_endline "after spawn";
  Unix.sleep 20;
  print_endline "after sleep"
```

always ends with segmentation fault. The output is usually
```
0
before kk assig
after kk assig
before spawn
1
after spawn
before kk assig
Segmentation fault
```

~~It seems that the problem occurs when performing `Yield` after `Unix.sleep 1`.~~ What helps is replacing `Domain.spawn (fun () -> !kk ())` by `!kk ()`.

radekm:
Also note that performing an effect in a different thread doesn't cause the problem. Eg. the following program doesn't end with segmentation fault:

```ocaml

effect Yield : unit

let () =
  begin match perform Yield with
    | () -> ()
    | effect Yield _ -> ()
  end;
  Domain.spawn
    (fun () ->
      match perform Yield with
        | () -> ()
        | effect Yield _ -> print_endline "yield");
  Domain.spawn (fun () -> perform Yield);
  Domain.spawn (fun () -> perform Yield);
  Unix.sleep 1
```

radekm:
The segmentation fault happens when assigning to `kk` - when executing `kk := (fun () -> continue k ())`. But when I assign value which doesn't contain `k` (eg. `(fun () -> continue (failwith "") ())`) then there is no segmentation fault.

kayceesrk:
Fixed by #14.

PR #14, opened 2015-07-10
Fix stack promotion.

kayceesrk:
As a part of promoting the stack, objects pointed to by the stack should also be promoted. This commit fixes a bug that ignores the latter if the stack is on a minor heap; `caml_scan_dirty_stack` only scans dirty stack whereas stacks in a minor heap are never marked as dirty. The fix always scans the stack, and also marks it clean since the stack no longer points to a minor heap.

The unconditional stack scanning can be optimized by only scanning the stack if the stack is in the minor heap or in the major heap and is marked as dirty.

### DIFF
--- a/byterun/fiber.c
+++ b/byterun/fiber.c
@@ -458,6 +458,14 @@ void caml_clean_stack(value stack)
   }
 }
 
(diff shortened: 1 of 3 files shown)


kayceesrk:
Also, this fixes #13.

PR #18, opened 2015-07-20
Scan dark stack

kayceesrk:
Since parts of the major heap may be marked between major collections, it is necessary to ensure that stacks which happen to be marked but are run after marking are scanned again to handle subsequent mutations to the stack. Otherwise, there is a chance of failing to mark live objects on live stacks.

Of course, we might end up doing duplicate work with this solution, re-scanning unchanged parts of the stack. Is it worth optimising this to avoid duplicate work a la native code [1].

[1] https://github.com/kayceesrk/ocaml/blob/effects/asmrun/roots.c#L211

### DIFF
--- a/byterun/fiber.c
+++ b/byterun/fiber.c
@@ -7,6 +7,7 @@
 #include "alloc.h"
 #include "platform.h"
 #include "fix_code.h"
+#include "shared_heap.h"
 
 #ifdef NATIVE_CODE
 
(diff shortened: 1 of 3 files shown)


byterun/fiber.c
```diff
@@ -73,6 +74,10 @@ static value save_stack ()
   Assert(caml_stack_high == Stack_high(old_stack));
   Assert(caml_extern_sp == caml_stack_high + Stack_sp(old_stack));
   dirty_stack(old_stack);
+  if (!Is_minor(old_stack) && caml_is_marked(old_stack)) {
+    //TODO: Only scan mutated part of the stack
+    caml_scan_stack(&caml_darken, old_stack);
+  }
```
stedolan: I don't understand why this is necessary. Shouldn't the call to `dirty_stack` above suffice? The minor GC scans all dirty stacks and calls caml_darken, regardless of whether they have been marked (see `clean_stacks` in `minor_gc.c`).

kayceesrk:
Right. The reason I'd added this fix was because this [1] `verify_heap` assertion was failing. Guess it doesn't count for much.

[1] https://github.com/ocamllabs/ocaml-multicore/blob/master/byterun/shared_heap.c#L455

PR #22, opened 2015-08-13
Reset realloc'ed stack to avoid useless scanning.

kayceesrk:
When a stack is realloc’ed, it is worth resetting the old stack contents to avoid space leaks. Since the stack is being realloc’ed, it is currently running. Hence, no references to this stack from other heap objects are “useful”; such a reference cannot be used to resume the continuation represented by this stack, which would violate the one-shotness. If the program carries around such references, it would constitute a space leak. Resetting the stack avoids the need for the GC to scan the old copy of the realloc'ed stack.

### DIFF
--- a/byterun/fiber.c
+++ b/byterun/fiber.c
@@ -381,6 +381,14 @@ void caml_realloc_stack(asize_t required_space, value* saved_vals, int nsaved)
   }
 
   load_stack(new_stack);
+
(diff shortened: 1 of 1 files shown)


stedolan:
Looks good, but you should also delete the now-redundant line 379.

kayceesrk:
Done.

PR #23, opened 2015-08-13
Darken dirty stack objects after emptying minor heap.

kayceesrk:
Assume that a dirty stack is already marked and it contains objects in
the minor heap. If the stack was darkened before emptying the minor
heap, then the minor heap objects on the stack will never be marked.
Darkening those objects after emptying minor heap ensures that those
objects will be marked.


stedolan:
This looks much nicer than what was there.

PR #31, opened 2015-10-01
Bvars

stedolan:
Implementation of bvars, used to represent continuations.

Here's a test for mutual exclusion:

https://gist.github.com/stedolan/4ec23cbdc2473de2adbd

Fixes a bug in fibers - due to use of caml_modify, the old continuation representation doesn't have proper mutual exclusion.

### DIFF
--- a/bytecomp/translcore.ml
+++ b/bytecomp/translcore.ml
@@ -1179,6 +1179,9 @@ and transl_match e arg pat_expr_list exn_pat_expr_list partial =
     static_catch [transl_exp arg] [val_id]
       (Matching.for_function e.exp_loc None (Lvar val_id) cases partial)
 
+and prim_bvar_create =
+  Pccall { prim_name = "caml_bvar_create"; prim_arity = 1; prim_alloc = true;
+           prim_native_name = ""; prim_native_float = false }
 
 and transl_handler e body val_caselist exn_caselist eff_caselist =
   let val_fun =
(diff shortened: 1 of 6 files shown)


kayceesrk:
The semantics of BVars, which transfers them on take or put, makes it expensive to use them for produce-consumer style programs (every operation transfers the bvar between domains), but reasonable for ping-ponging (write is local, read is remote) between domains. Is this choice deliberate? By splitting the take/put and transfer (share or steal) as two independent operations (bvar remains with a domain unless it is transferred), one could optimize for both cases.

stedolan:
On Thu, Oct 1, 2015 at 9:44 PM, KC Sivaramakrishnan <
notifications@github.com> wrote:

> The semantics of BVars, which transfers them on take or put, makes it
> expensive to use them for produce-consumer style programs (every operation
> transfers the bvar between domains), but reasonable for ping-ponging (write
> is local, read is remote) between domains. Is this choice deliberate? By
> splitting the take/put and transfer (share or steal) as two independent
> operations (bvar remains with a domain unless it is transferred), one could
> optimize for both cases.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ocamllabs/ocaml-multicore/pull/31#issuecomment-144842734>
> .
>
The semantics are mostly to match what's currently done for fibers, rather
than a separate conscious choice. Fibers can currently be resumed from any
domain, without an explicit transfer of ownership, so works the same way.

I think this is probably the right way: OCaml code should probably not be
aware of domains except as a performance optimisation.

kayceesrk:
This PR has been superseded by #32?

stedolan:
Yes, closing.

PR #32, opened 2015-10-06
Effects refactor

stedolan:
**Simplify instruction set**

Replace HANDLE, CONTINUE, DISCONTINUE, CONTINUETERM and DISCONTINUETERM with two instructions RESUME and RESUMETERM.

RESUME/RESUMETERM take an extra argument, a closure to be applied on the target stack. For "continue", the closure is `fun x -> x`, for "discontinue" it's `fun e -> raise e`, and handlers are compiled to a call to the new primitive "caml_alloc_stack" and a use of RESUME/RESUMETERM.

Remove the FINISH opcode, and instead check for fiber termination in the RET instruction. This saves 3 words per fiber, since the extra stack frame used for FINISH is removed.

**Optimised delegation**

Add a DELEGATETERM opcode, used to implement the default case of effect matching.

Effects are forwarded to the next handler by reversing the parent pointers. That is, when handler A performs and effect, delegated by A's parent B to B's parent C, the parent pointer of the performer (A) is set to the delegator (B).

This accumulates a singly-linked list of delegators, which is reversed by RESUME when the continuation is invoked. So, delegators are not context-switched back into after delegating an effect: control passes directly back to the original performer.


kayceesrk:
Reviewing this PR. Native code build hasn't been ported to the refactored effects. It might be useful to do this before the merge.

stedolan:
Native code now has effect stubs.

kayceesrk:
Excellent. I still haven't finished going through the entire PR closely. Will do it by the end of the week.

asmcomp/selectgen.ml
```diff
@@ -288,6 +292,9 @@ method select_operation op args =
   | (Cfloatofint, _) -> (Ifloatofint, args)
   | (Cintoffloat, _) -> (Iintoffloat, args)
   | (Ccheckbound _, _) -> self#select_arith Icheckbound args
+  | (Cperform, _) -> (Iperform, args)
+  | (Cresume, _) -> (Iresume_ind, args)
```
kayceesrk: The distinction between `Iresume_ind` and `Itail_resume_ind` is never made. Should this be similar to `Icall_ind` at tail position [0]?

[0] https://github.com/ocamllabs/ocaml-multicore/blob/effects-refactor/asmcomp/selectgen.ml#L728-L744

asmcomp/mach.mli
```diff
@@ -43,6 +43,10 @@ type operation =
   | Icall_imm of string
   | Itailcall_ind
   | Itailcall_imm of string
+  | Iperform
+  | Itail_delegate
```
kayceesrk: `delegate` is always at tail position. So should `Itail_delegate` be just `Idelegate`? Even otherwise, `Itaildelegate` is more consistent with the naming convention for calls.

asmcomp/printmach.ml
```diff
@@ -111,6 +111,10 @@ let operation op arg ppf res =
   | Icall_imm lbl -> fprintf ppf "call \"%s\" %a" lbl regs arg
   | Itailcall_ind -> fprintf ppf "tailcall %a" regs arg
   | Itailcall_imm lbl -> fprintf ppf "tailcall \"%s\" %a" lbl regs arg
+  | Iperform -> fprintf ppf "perform %a" regs arg
+  | Itail_delegate -> fprintf ppf "tail_delegate %a" regs arg
```
kayceesrk: `tail_delegate` should just be `delegate`.

stdlib/effects.ml
```diff
@@ -1,3 +1,4 @@
+(*
```
kayceesrk: This file should be removed.

stdlib/effects.mli
```diff
@@ -1,3 +1,5 @@
+(*
```
kayceesrk: This file should be removed.

kayceesrk:
I've gone through the patch closely. Other than the comments (which should perhaps be moved to the issues?) the merge looks ok to me.

asmcomp/selectgen.ml
```diff
@@ -42,6 +42,10 @@ let oper_result_type = function
   | Cintoffloat -> typ_int
   | Craise _ -> typ_void
   | Ccheckbound _ -> typ_void
+  | Cperform -> typ_addr
```
kayceesrk: Is the result type of `Cperform`, `Cresume` and `Cdelegate` always typ_addr? Should they be parameterized with the result type similar to `Capply`?

lpw25: It looks to me like having a parameterized result type like Capply might allow slightly better code generation when the result of these operations is not used because it can be turned into `typ_void`, but I'm not very familiar with this part of the compiler so I could be wrong.

I'm also surprised that the results are `typ_addr` rather than `typ_val` since I would have thought they returned valid OCaml values that should be made GC roots.

stedolan: `typ_addr` is correct here, but will need changing when we merge to current trunk.

Multicore is currently based on the 4.02.1 code. 4.02.1 has the following `machtype_components`:
 - `Addr`: general values, integers or pointers
 - `Int`: tagged integers (subtype of `Addr`)
 - `Float`: unboxed floats

A `machtype` is an array of the above, so `typ_addr = [| Addr |]`, etc. `typ_void` is `[| |]`, for operations that produce no result, not even unit.

It transpired that `Addr` was used for intermediate values, like derived pointers into the middle of an array, which are unsafe to expose to the GC. So, recently, `Addr` was split into:
 - `Val`: general values, integers or pointers
 - `Addr`: derived pointers which must not be live by the next safepoint.

So, `typ_addr` is correct now, but should change to `typ_val` when we merge. The naming is unfortunate.

asmcomp/selectgen.ml
stedolan: Yes, and it will be. I haven't implemented tail-call optimisation yet. (The _ind version is always correct, but the tail_ind version is faster / uses less space)

PR #35, opened 2016-04-10
Move based promotion

kayceesrk:
Move-based promotion scheme for promoting objects from minor to major heaps.


kayceesrk:
@stedolan Ready for review.

byterun/caml/shared_heap.h
```diff
@@ -22,8 +22,10 @@ intnat caml_sweep(struct caml_heap_state*, intnat);
 /* must be called during STW */
 void caml_cycle_heap_stw(void);
 
-/* must be called on each domain 
+/* must be called on each domain
    (after caml_cycle_heap_stw) */
 void caml_cycle_heap(struct caml_heap_state*);
 
+int is_garbage (value);
```
stedolan: I don't really like this function, because it's not in general valid to check whether a value is garbage. The only correct use of this is `Assert(!is_garbage(...))`. How about adding `Check_hd(hd)` to `misc.h` instead, which checks that `hd != 0` and `hd` is not garbage?

byterun/interp.c
```diff
@@ -686,6 +687,8 @@ value caml_interprete(code_t prog, asize_t prog_size)
         Init_field(block, 0, accu);
         for (i = 1; i < wosize; i++) Init_field(block, i, *sp++);
       } else {
+        /* XXX KC: Is setup/restore needed? See comment in caml_alloc_shr in
+         * CLOSURE and CLOSUREREC. */
```
kayceesrk: Removed the Setup_ & Restore_after_gc.

byterun/major_gc.c
```diff
@@ -111,12 +116,16 @@ static intnat mark(value initial, intnat budget) {
...
+    // caml_gc_log ("mark: v=0x%lx hd=0x%lx tag=%d sz=%lu",
+    //             v, hd_v, Tag_val(v), Wosize_val(v));
     if (Tag_hd (hd_v) == Stack_tag) {
+      // caml_gc_log ("mark: stack=%p", (value*)v);
       caml_scan_stack(&caml_darken, v);
     } else if (Tag_hd (hd_v) < No_scan_tag) {
       int i;
       for (i = 0; i < Wosize_hd(hd_v); i++) {
-        value child = Field(v, i);
+        value child = FieldImm(v, i);
```
kayceesrk: Done.

byterun/memory.c
```diff
@@ -216,39 +208,6 @@ struct write_fault_req {
...
-    struct domain* owner = caml_owner_of_shared_block(obj);
-    struct write_fault_req req = {obj, field, val};
-    caml_gc_log("Write fault to domain [%02d]", owner->id);
-    caml_domain_rpc(owner, &handle_write_fault, &req);
-    shared_heap_write_barrier(obj, field, val);
-  }
-}
-
-
 struct cas_fault_req {
```
kayceesrk: Done.

byterun/memory.c
```diff
@@ -381,16 +290,11 @@ CAMLprim value caml_bvar_take(value bv)
 
 CAMLprim value caml_bvar_put(value bv, value v)
 {
-  /* bvar operations need to operate on the promoted copy */
```
kayceesrk: :-)

byterun/minor_gc.c
```diff
@@ -232,228 +97,440 @@ static value alloc_shared(mlsize_t wosize, tag_t tag)
...
-        }else{
+        } else {
           v = f;                        /* Follow the forwarding */
           goto tail_call;               /*  then oldify. */
         }
       }
     }
-  }else{
+  } else {
+    /* XXX KC: Does this operation race with other domain's oldifying operation */
```
kayceesrk: 👍

byterun/minor_gc.c
```diff
@@ -232,228 +97,440 @@ static value alloc_shared(mlsize_t wosize, tag_t tag)
...
+        Assert (Is_block(vnew) && !Is_minor(vnew));
+        Assert (Hd_val(vnew));
         if (Tag_hd(hd) == Infix_tag) { Assert(Tag_val(vnew) == Infix_tag); }
-        rewritten += caml_atomic_cas_field(r->obj, r->field, v, vnew);
+        if (__sync_bool_compare_and_swap (*r,v,vnew)) ++rewritten;
+        caml_darken(vnew,0);
       }
+      //XXX KC: Could I move the darkening into the conditional above as
+      //caml_darken(vnew,0)?
+      //caml_darken (**r,*r);
```
kayceesrk: 👍

byterun/caml/shared_heap.h
kayceesrk: The value `global.GARBAGE` is local to `shared_heap.c`. `is_garbage` is now only defined under DEBUG mode. Hopefully, this makes it more clear where `is_garbage` should be used.

lpw25:
One thing that I don't see is changes to when the compiler generates a read barrier. IIRC copy-based promotion required a read barrier on all mutable reads, whereas move-based promotion only requires a read barrier on mutable reads which may involve pointers.

lpw25:
Similarly to the above, I think that there are some uses of `Field` lying around which could now be changed to `FieldImm` since the values are always integers.

stedolan:
> One thing that I don't see is changes to when the compiler generates a read barrier. IIRC copy-based promotion required a read barrier on all mutable reads, whereas move-based promotion only requires a read barrier on mutable reads which may involve pointers.

That is true. I think this optimisation should be a separate pull rq, though.

> Similarly to the above, I think that there are some uses of Field lying around which could now be changed to FieldImm since the values are always integers.

Yes, there's quite a few changes to the C API that need to be made. I'm going to submit those as a separate pull rq.

bluddy:
So now, when a domain needs a mutable object in another domain's minor heap, that object is moved to the major heap, rather than copied and synchronized?

stedolan:
On Thu, Nov 17, 2016 at 3:57 PM, Yotam Barnoy <notifications@github.com>
wrote:

> So now, when a domain needs a mutable object in another domain's minor
> heap, that object is moved to the major heap, rather than copied and
> synchronized?
>
Yep!

PR #47, opened 2016-05-17
Native code backend for multicore

kayceesrk:
This PR adds nativecode support. I've ported the nativecode changes from `ocaml-effects`. All the thread local variables accessed by asm code are based off the domain tls area. To make the code uniform between byte and native, the thread local variables `caml_current_stack`, `caml_stack_high` and `caml_stack_threshold` have been removed. They are now accessed as `caml_domain_state->current_stack`, etc. All the tests in `./native-tests` directory succeed!


kayceesrk:
@stedolan @lpw25 I've tested the native backend on small benchmarks (chameneos-redux, reagents). The GC works fine under multiple cores, except for a single heisenbug that I've been trying to debug all week [0]. I can't seem to make the program deadlock in the native code versions of either move-based promotion [1] or promotion with minor GC [2] variants. This leads me to think that the deadlock bug has to do with barriers in copy-based promotion. It would be good to get this PR reviewed and merged since the move-based GC variants are what we want in the long run.

[0] https://github.com/ocamllabs/ocaml-multicore/issues/51
[1] https://github.com/ocamllabs/ocaml-multicore/tree/move-native
[2] https://github.com/ocamllabs/ocaml-multicore/tree/promote-with-minor-gc-native

lpw25:
It looks good to me. If Stephen's happy with it go ahead and merge.

stedolan:
This looks good! I'm afraid I broke the merge by merging the delegate -> reperform change, but happy to merge once it's possible again.

Do backtraces work at the moment? There are some odd-looking `caml_fatal_errors` in backtrace.c.

I'm happy to merge it at the moment, but I'd like to remove the "slop space" soon. It's used for two things, as far as I can see: signal handler stack space and `mcount` (for profiling).

Signal handlers should use a separate stack, allocated with `sigaltstack` at startup. It turns out that `mcount` is a short assembly wrapper that calls the C function `mcount_internal` passing the current PC and return address (see `sysdeps/x86_64/_mcount.S` in glibc source), so we can handle `mcount` as a standard C call (switching to the system stack) if we call `mcount_internal` instead.

As a future optimisation, a noalloc extcall can be made more efficient by instead of saving and restoring `%r10` to the OCaml stack, using a register preserved by C calls (e.g. `%r13`) and marking it `destroyed_at_oper` in `proc.ml`

kayceesrk:
> I'm afraid I broke the merge by merging the delegate -> reperform change, but happy to merge once it's possible again.

No worries. I'll merge upstream edits next.

> Do backtraces work at the moment? There are some odd-looking caml_fatal_errors in backtrace.c.

I've deliberately broken backtraces. Haven't thought about how to fix this in a backwards compatible manner. My plan is to fix backtraces such that we get a complete backtrace through handlers and nested calls to C.

I've made issues for the suggested improvements: https://github.com/ocamllabs/ocaml-multicore/issues/54 & https://github.com/ocamllabs/ocaml-multicore/issues/55.

PR #58, opened 2016-08-22
Sys-threads support for Multicore OCaml

kayceesrk:
This PR adds systhreads support for Multicore OCaml. While it seems unnecessary that systhreads should be required under Multicore, several packages (including lwt) depend on thread support. The aim is to provide backwards compatible implementation of systhreads in Multicore.

## Design

Every systhread is associated with a domain. No two systhreads that belong to the same domain  run OCaml code simultaneously; each domain is associated with a master lock that needs to be taken before a systhread may run. systhreads that belong to two different domains may run OCaml code in parallel. Therefore, with just a single domain, the behavior is backwards compatible with vanilla OCaml behavior. *Systhreads are untested in the presence of multiple domains.*

## Other improvements

This PR also makes other improvements. The domain local states remembered set and mark stack are moved to the end-of-minor-heap area.


kayceesrk:
I've tested this PR on lwt, and it seems to work fine. I am merging this PR. We can revisit if a better design for systhreads for multicore comes up.

PR #65, opened 2016-09-21
Make TLS work on OSX

bluddy:
OSX has almost no support for TLS. As such, I'm using pthread's TLS support instead and hiding it via macros.

Unfortunately, pthread TLS storage is slower than just a pointer access, requiring function calls and spin-locks, and this might mean some local caching of TLS values may be desirable. This local caching will almost certainly be optimized away by the C compiler, whereas beta reduction/CSE optimization on the pthread calls is unlikely to happen.

The current PR is a work in progress, and currently doesn't compile, failing at caml_local_roots, which is a TLS variable I have yet to deal with.


bluddy:
There's a problem with making caml_local_roots TLS under this model. `domain.c` line 196 saves the address of `caml_local_roots` into `d->state.local_roots`. This obviously doesn't work if `caml_local_roots` is a value stored in pthreads' TLS table.

In general, the programming model needs to incorporate the fact that to properly support OSX as it currently is (and OSX is one of the chief platforms), TLS access must be considered to have a non-negligible cost, and TLS values don't reside in normal memory.

Not sure what to do about this issue right now -- looks like I need to dive deeper into the logic.

bluddy:
Having looked at the assembly produced on OS X by calling `pthread_getspecific` and `pthread_setspecific`, it doesn't look too bad. There's overhead for calling the function, but the actual implementation seems to be written in assembly and therefore quite optimized. This means it should be fine for replacing __thread.

However, the code still cannot assume that this data has an 'address' -- that needs to be addressed.

bluddy:
OK so I've managed to modify local_roots and make it use pthreads' TLS. Now I'm getting weird errors: "Serializing closures is broken". 

@kayceesrk, is there any path (in ocamlrun perhaps) that doesn't start with startup.c in either byterun or asmrun? Because that's where I'm initializing the pthread TLS keys.

kayceesrk:
@bluddy Great! but also, wat? Perhaps @stedolan knows whats up. There was a mention of serialization being broken under multicore.

Thanks for continuing looking into this issue.

bluddy:
@kayceesrk I can get past this error by removing '-bin-annot' on the 'boot/ocamlrun boot/ocamlc ... list.ml' invocation, but then I fail elsewhere. Looks like the interpreter gets corrupted. I'm going to try to debug with lldb.

bluddy:
@kayceesrk, @stedolan: Running the debug runtime under valgrind, I've got [this output](https://gist.github.com/bluddy/141475f1ee8912268269dcc6dd136bb7). Seems like the '''set address range perms''' isn't an issue, but are all these faults normal? Eventually I have an invalid read, but it's not obvious why that happens.

bluddy:
@kayceesrk, any idea how to debug this? I've tried `lldb` with `../byterun/ocamlrund ../boot/ocamlc ...` but it's a little hard to debug the interpreter this way since you're always running the same code. I end up using a lot of conditional breakpoints, which `lldb` doesn't seem to be doing very well with. Is there a preferred method to debug the interpreter?

bluddy:
I added some debug printout statements, and this is what I get:

```
[-1] Opening bytecode executable ../byterun/ocamlrund
[-1] Not a bytecode executable
[-1] Opening bytecode executable ../boot/ocamlc
[-1] Initial stack limit: 8192k bytes
cur_instr = 59, pc = 2b6, accu = 1, sp[0] = 0, sp[1] = 0
...
cur_instr = 1b, pc = 4e, accu = 7e9, sp[0] = 7e9, sp[1] = 20008cd60
cur_instr = 4e, pc = 5e, accu = 10ca1b710, sp[0] = 7e9, sp[1] = 7e9
[00] Write fault to domain [00]
[00] Handling write fault for domain [00] on 0x10ca1b710(0)
cur_instr = 5e, pc = 7d, accu = 1, sp[0] = 7e9, sp[1] = 20008cd60
cur_instr = 68, pc = 6d, accu = 1, sp[0] = 10cb54641, sp[1] = fffffffffffffe8d
...
cur_instr = 1b, pc = 21, accu = 10cd14550, sp[0] = 20008c688, sp[1] = 10cb2b459
cur_instr = 21, pc = 10, accu = 10ca1c230, sp[0] = 10cd14550, sp[1] = 20008c688
cur_instr = 0, pc = 38, accu = 10ca1c230, sp[0] = 10cd14550, sp[1] = 10cb2b42d
cur_instr = 38, pc = 2a, accu = 10cd14550, sp[0] = 10cd14550, sp[1] = 10cb2b42d
cur_instr = 21, pc = b, accu = 10ca1fd50, sp[0] = 10cd14550, sp[1] = 10cd14550
cur_instr = 0, pc = 62, accu = 10ca1fd50, sp[0] = 10cd14550, sp[1] = 10cb2b51d
cur_instr = 62, pc = 59, accu = 10cd14550, sp[0] = 10cd14550, sp[1] = 10cb2b51d
[00] ref_table threshold crossed
[00] Minor collection starting
[00] CAS fault to domain [00]
[00] Handling CAS fault for domain [00]
[00] CAS fault to domain [00]
[00] Handling CAS fault for domain [00]
...
[00] Minor collection completed: 1037 of 1499 kb live, 29942 pointers rewritten
[00] Pooled memory: 221184 alloced, 95499 free, 1129 fragmentation
[00] Large memory: 48361 alloced, 0 free, 46 fragmentation
[00] Major slice: 142791 alloc, 749652 work, 0 sweep, 148965 mark (35943 blocks)
[00] Contributing 256 heaps to GC
[00] Pooled memory: 221184 alloced, 95499 free, 1129 fragmentation
[00] Large memory: 48361 alloced, 0 free, 46 fragmentation
[00] Finished marking major heap. Marked 35943 blocks
[00] GC cycle completed (heap cycled)
[00] Cycling heap [00]
...
cur_instr = 44, pc = f, accu = 20008c670, sp[0] = 2001ff6d0, sp[1] = 1
cur_instr = f, pc = 43, accu = d700d7d7d700d6d7, sp[0] = 2001ff6d0, sp[1] = 1
cur_instr = 43, pc = 3e, accu = 20008c670, sp[0] = d700d7d7d700d6d7, sp[1] = 2001ff6d0
cur_instr = 3e, pc = 4, accu = d700d7d7d700d6d7, sp[0] = d700d7d7d700d6d7, sp[1] = 2001ff6d0
```

That long number, d700d..., is what eventually causes a segmentation fault. It is a result of a 0x44 instruction (GETFIELD1), which suggests that memory is being corrupted by the GC, as it happens after collection.

I wonder if this is an LLVM vs. GCC thing.

bluddy:
More progress on debugging:

There's definitely a bug in the GC that occurs here. Before GC is triggered, `accu` is 0x20008c688, which points to valid data. After GC, the accumulator is changed to 0x2008c670, but that area in memory has only garbage. I'll have to step through the GC code to figure out what goes wrong.

kayceesrk:
@bluddy Am I correct in understanding that errors that you see are during `make world.opt`? `d700d7d7d700d6d7` is magic value that is written into the heap during debug runs and represents uninitialized memory [1]. This value should never be read. Indicates a bug in the GC.

[1]: https://github.com/ocamllabs/ocaml-multicore/blob/052a12d881e5b58b5a53d4d5f29367d0c8aaf597/byterun/caml/misc.h#L124-L153

bluddy:
Ah thanks. That makes sense, since I've been seeing a lot of memory with this value and wondered what it was.

This happens in either `make world` or `make world.opt`, in the bootstrapping phase. The bytecode runtime has been built, and it's going through and using `boot/ocamlc` to build the stdlib, and failing on specific files after hitting a GC.

Currently I'm not sure which one of these is the problem:
a. My conversion is buggy. I'm not sure how, since it's a pretty simple transformation and I've gone through it and checked for errors, though I might have missed something.
b. There's an bug in the multicore code specifically to do with `clang`.
c. There's a bug in the multicore code specifically to do with `OSX`.
c. `clang` causes a pre-existing universal bug to show itself.
d. My conversion causes a pre-existing universal bug to show itself.
e. `OSX` causes a pre-existing universal bug to show itself.

I've wanted to run a Linux VM and see if it happens there when run on clang, and then try my transformation and see what happens. I should do that.

bluddy:
OK so the problem had to do with my code. To resolve the problem with caml_local_roots where `domain` contained a double-pointer to the local-roots (to be synchronized with the list head), I changed it to a single-pointer, not realizing that this was an issue. `domain` didn't track the changes to the TLS caml_local_roots.

To fix this, I moved caml_local_roots into `domain_state`. It's an extra pointer dereference to get to caml_local_roots, but I don't see how this is avoidable in the general case.

I'm now hitting other issues later in compilation.

bluddy:
@kayceesrk, I've gotten to the point that both native and bytecode versions compile. I've moved all `extern __thread` vars to the domain_state, and I've created a corresponding `byte_domain_state.tbl` for `extern __thread` vars that only exist in the bytecode interpreter.

More thoughts:

1. All `static __thread` vars have been kept, but they probably should be moved as well. I can do that in a PR after this one is committed.
2. I found some hints that the native code is supposed access domain_state stuff via ocaml_young_ptr which is `r15` in the OCaml code... is that right? How would that work? `r15` would be synchronized in any OCaml context switch, but the C code has no way to access ocaml_young_ptr except as a global, which means that any context switch in the C code will give you the wrong caml_young_ptr.
3. From reading up some more, it seems that accessing TLS vars is universally expensive (on all systems). Do we want to consider passing the domain_state var as a first argument in the C code?

bluddy:
ping @kayceesrk -- any thoughts?

kayceesrk:
I am currently away until mid-nov and shall take a close look then. Perhaps
@stedolan can answer some of the questions and review the PR.

bluddy:
Is there any interest in making OSX work? I'm happy to update the branch if anyone's interested.

DemiMarie:
There is another option: hijack an existing TLS key that is meant for something else.  That's what GHC does.

kayceesrk:
@DemiMarie 

> There is another option: hijack an existing TLS key that is meant for something else. That's what GHC does.

Could you explain what you are suggesting this as an alternative to? More info on the GHC idea (or a link to the code) would be quite useful.

stedolan:
This causes a segfault whenever a new domain is created, since `caml_local_roots` is accessed in `domain_thread_func` before domain state is initialised.

kayceesrk:
@stedolan we need tests in the testsuite that create multiple domains :-|

stedolan:
> @stedolan we need tests in the testsuite that create multiple domains :-|

There are! I learned about this when tests/parallel segfaulted.

kayceesrk:
looks like we don't run these tests in CI: https://github.com/ocamllabs/ocaml-multicore/blob/master/.travis-ci.sh#L29

stedolan:
> looks like we don't run these tests in CI: https://github.com/ocamllabs/ocaml-multicore/blob/master/.travis-ci.sh#L29

Hmm. The testsuite's not in great shape at the moment. I've started fixing it at #110.

PR #70, opened 2016-11-18
Moving tests from native-tests to the testsuite.

kayceesrk:
Introduces tests for effects into the testsuite. Now

```
make one DIR=tests/effects
```
runs the tests for effects. There are additional tests in callbacks.


kayceesrk:
Merging the PR since it is non-intrusive. @stedolan Now there is a place for you to add your tests. :-)

PR #72, opened 2016-11-18
Alloc API changes (1/2)

stedolan:
Add the convenience functions `caml_alloc_<n>`, and simplify how `Alloc_small` works.

### DIFF
--- a/byterun/alloc.c
+++ b/byterun/alloc.c
@@ -25,9 +25,6 @@
 #include "caml/fiber.h"
 #include "caml/domain.h"
 
-#define Setup_for_gc
-#define Restore_after_gc
-
 CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
 {
   value result;
@@ -38,7 +35,7 @@ CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
   if (wosize == 0){
     result = Atom (tag);
   }else if (wosize <= Max_young_wosize){
-    Alloc_small (result, wosize, tag);
+    Alloc_small (result, wosize, tag, { caml_handle_gc_interrupt(); });
     if (tag < No_scan_tag){
       for (i = 0; i < wosize; i++) Init_field (result, i, Val_unit);
     }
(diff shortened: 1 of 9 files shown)


byterun/alloc.c
```diff
@@ -53,14 +51,116 @@ CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
...
 
+static inline void enter_gc_preserving_vals(mlsize_t wosize, value* vals)
+{
+  CAMLparam0();
+  /* Copy the values to be preserved to a different array.
+     The original vals array never escapes, generating better code in
+     the fast path. */
+  CAMLlocalN(vals_copy, wosize);
+  mlsize_t i;
+  for (i = 0; i < wosize; i++) vals_copy[i] = vals[i];
```
kayceesrk: Why is this copy back and forth required? Can we do anything to avoid this?

byterun/alloc.c
```diff
@@ -53,14 +51,116 @@ CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
...
+                               value e)
+{
+  value v[5] = {a, b, c, d, e};
+  return do_alloc_small(5, tag, v);
+}
+
+CAMLexport value caml_alloc_6 (tag_t tag, value a, value b, value c, value d,
+                               value e, value f)
+{
+  value v[6] = {a, b, c, d, e, f};
```
kayceesrk: Can this and similar allocations in `caml_alloc_N` be `CAMLlocalN`? Since the very next thing we do is `do_alloc_small()` -> `enter_gc_preserving_calls()` which copies the values from this array to `CAMLlocalN` structure.

stedolan: It was originally written like that, but this way generates much better code at `-O2` (tested gcc and clang).

Since the `v` array does not escape, the C compiler removes the stack allocation after inlining `do_alloc_small` and `enter_gc_preserving_vals`. Taking a copy of the array in `enter_gc_preserving_vals` ensures that the `v` array does not escape, since the address is taken (via `caml_local_roots`) of a different array. This approach pushes the work of setting up `caml_local_roots` to the case where the GC is actually invoked.

kayceesrk: Great. The changes look good to me. 👍 Feel free to merge when ready.

PR #80, opened 2016-12-13
Alloc API change (3/4)

stedolan:
(the expected number of pull requests keeps increasing, but I'm pretty sure the next one will be the last...)

Remove most uses of `Field`. This pull request handles the large number of fairly-easy cases. The next one will contain some trickier ones that require more careful review (places where the current code relies on subtle GC invariants).

The changes to win32unix and win32graph are untested.

### DIFF
--- a/byterun/alloc.c
+++ b/byterun/alloc.c
@@ -46,11 +46,13 @@ CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
         Op_val(result)[i] = init_val;
       }
     }
+
+    if (tag == Stack_tag) Stack_sp(result) = 0;
   }else{
     result = caml_alloc_shr (wosize, tag);
-    if (tag == Stack_tag) Stack_sp(result) = 0;
-    result = caml_check_urgent_gc (result);
+    result = caml_check_urgent_gc(result);
   }
+
   return result;
 }
 
@@ -226,8 +228,8 @@ CAMLexport int caml_convert_flag_list(value list, const int *flags)
   int res;
   res = 0;
   while (list != Val_int(0)) {
-    res |= flags[Int_val(Field(list, 0))];
-    list = Field(list, 1);
+    res |= flags[Int_field(list, 0)];
+    list = Field_imm(list, 1);
   }
   return res;
 }
@@ -252,6 +254,8 @@ CAMLprim value caml_alloc_dummy_float (value size)
 
 CAMLprim value caml_update_dummy(value dummy, value newval)
 {
+  CAMLparam2(dummy, newval);
+  CAMLlocal1(x);
   mlsize_t size, i;
   tag_t tag;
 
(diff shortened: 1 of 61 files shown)


byterun/alloc.c
```diff
@@ -46,11 +46,13 @@ CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
         Op_val(result)[i] = init_val;
       }
     }
+
+    if (tag == Stack_tag) Stack_sp(result) = 0;
```
kayceesrk: what's the reasoning behind this change?

byterun/array.c
```diff
@@ -35,9 +35,12 @@ CAMLexport int caml_is_double_array(value array)
 
 CAMLprim value caml_array_get_addr(value array, value index)
 {
+  CAMLparam2(array, index);
+  CAMLlocal1(x);
   intnat idx = Long_val(index);
   if (idx < 0 || idx >= Wosize_val(array)) caml_array_bound_error();
-  return Field(array, idx);
+  caml_read_field(array, idx, &x);
```
kayceesrk: Do we need `CAMLParam`/`CAMLlocal` in this function at all? The variables are not live across this safe point.

byterun/array.c
```diff
@@ -100,10 +103,13 @@ CAMLprim value caml_array_unsafe_get_float(value array, value index)
...
 CAMLprim value caml_array_unsafe_get(value array, value index)
 {
+  CAMLparam2(array, index);
+  CAMLlocal1(x);
   if (Tag_val(array) == Double_array_tag)
-    return caml_array_unsafe_get_float(array, index);
+    x = caml_array_unsafe_get_float(array, index);
   else
-    return Field(array, Long_val(index));
+    caml_read_field(array, Long_val(index), &x);
```
kayceesrk: Same comments as `caml_array_get_addr`. It should be worthwhile optimising for the array access.

kayceesrk:
I've skimmed through the rest of the commits. Have you run this PR through the unit tests from the testsuite? Some of them are broken now (backtraces, finalizers, etc), but there are some useful ones in there.

byterun/alloc.c
stedolan: `caml_alloc` used to initialise `Stack_sp(result)` only in the too-big-for-minor-heap case, which seems like a bug. This commit always initialises `Stack_sp`, in `caml_alloc` for small allocations and in `caml_alloc_shr` for large ones.

stedolan:
Your suggested changes to `array.c` are correct, but I wasn't terribly concerned with performance there: in native code array access is inlined, so these functions only matter with bytecode.

The testsuite is no more broken with this patch than in master, but neither are in good shape. Debugging some stuff that really shouldn't be failing in `tests/typing-gadts` at the moment, but I don't think it has anything to do with this patch (it also fails on master).

byterun/alloc.c
kayceesrk: That makes sense.

kayceesrk:
Looks good to me.

PR #82, opened 2016-12-20
Moved macro 'Assert' from private zone into public space in misc.h

dhil:
I believe the macro 'Assert' was wrongly defined inside a private zone in misc.h. I have moved the macro into public space.


kayceesrk:
This looks good to me. Thanks!

PR #109, opened 2017-03-06
Atomics library

stedolan:
Very basic atomic primitives.

Unoptimised implementation in C with heavyweight fences around every operation. Platform-specific fast paths to come later.


lpw25:
I'm not sure about the name `Atomics`. It implies that ordinary `ref`s doen't have atomic read and write, whereas IIUC they do have atomic read and write just not synchronising.

Also, I prefer `compare_and_swap` to `compare_and_set`, but maybe that's just me.

stedolan:
> I'm not sure about the name Atomics. It implies that ordinary refs doen't have atomic read and write, whereas IIUC they do have atomic read and write just not synchronising.

True, but `Atomic` is a fairly standard name for this concept. Also, only `Atomic.t` has atomic read-modify-write operations.

> Also, I prefer compare_and_swap to compare_and_set, but maybe that's just me.

To me, those are slightly different operations (or more accurately, different APIs to the same operation). `compare_and_set` returns a boolean indicating whether the operation succeeded, while `compare_and_swap` returns the value that was previously in the location (and you do the comparison with `old` yourself to know whether it has been updated). (GCC calls these `__sync_bool_compare_and_swap` and `__sync_val_compare_and_swap`, but I find `set` / `swap` clearer)

byterun/memory.c
```diff
@@ -97,6 +97,49 @@ CAMLexport int caml_atomic_cas_field (value obj, int field, value oldval, value
...
+  return Val_unit;
+}
+
+CAMLprim value caml_atomic_cas (value ref, value oldv, value newv)
+{
+  value* p = Op_val(ref);
+  if (Is_young(ref)) {
+    if (*p == oldv) {
+      *p = newv;
+      write_barrier(ref, 0, newv);
```
kayceesrk: Is it safe to call `write_barrier` after the value has been updated? The current write barrier assumes that `write_barrier` is invoked before the update: https://github.com/ocamllabs/ocaml-multicore/blob/master/byterun/memory.c#L33. 
Since the write barrier only expects to read the old value in the case of young ref, we should be able to safely swap this line with the previous line. 

I believe `caml_atomic_cas_field` might also be incorrect due to this reason and needs a similar change.

byterun/memory.c
```diff
@@ -97,6 +97,49 @@ CAMLexport int caml_atomic_cas_field (value obj, int field, value oldval, value
...
+      return Val_int(1);
+    } else {
+      return Val_int(0);
+    }
+  } else {
+    int r = __sync_bool_compare_and_swap(p, oldv, newv);
+    if (r) write_barrier(ref, 0, newv);
```
kayceesrk: in order to handle this case, once we move to yuasa (which darkens the old value), we might need to change the write barrier to explicitly accept the old value as an argument.

kayceesrk:
We should also provide `compare_and_swap` which @fondation451 needs for his k-cas algorithm.

byterun/memory.c
stedolan: Nice catch! I think passing the old value explicitly to the write barrier is a good idea. That way, `write_barrier` touches no heap memory and can be put wherever.

byterun/memory.c
stedolan: `caml_atomic_cas_field` now fixed on master, and this pullrq is updated to pass `oldv` to `write_barrier`.

kayceesrk:
The comments in the atomic primitives [0] suggest that `__sync_bool_compare_and_swap` is weak (no spurious failures) while `__sync_val_compare_and_swap` is strong. Is this true on all architectures? C++ seems to provide both strong and weak version of the `bool` returning versions but does not provide `val` returning versions [1]. 

[0] https://github.com/ocamllabs/ocaml-multicore/blob/74c472bb4dbdaa8e3e5b3e3cff44ab76054a5abb/byterun/caml/platform.h#L113-L122
[1] http://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange

stedolan:
> The comments in the atomic primitives [0] suggest that __sync_bool_compare_and_swap is weak

I think that comment is just wrong, and __sync_foo is strong.

kayceesrk:
We should merge this PR. I'd like to start programming with this interface to gain more experience. Porting reagents would be a good test for the interface.

c-cube:
Sorry to intrude, but I'm wondering: why does this have to be in the compiler, and not as a separate library? I would think Mutexes would be sufficient for a first release of multicore?

lpw25:
> why does this have to be in the compiler, and not as a separate library?

I believe that the compiler needs to know about these operations in order to implement them efficiently.

> I would think Mutexes would be sufficient for a first release of multicore?

That would preclude any lock-free data structures (e.g. Reagents).

c-cube:
Thanks for the explanations.

stedolan:
It's not just efficiency. The semantics of shared references are a fundamental part of the language, and the interactions between references and atomic variables are the most important part of that, so it's not something that it makes sense to do outside the compiler. See [the memory model](https://github.com/ocamllabs/ocaml-multicore/wiki/Memory-model) for details.

PR #113, opened 2017-03-13
Use stateful scanning_action for minor GC and promotion.

stedolan:


byterun/minor_gc.c
```diff
@@ -198,47 +200,43 @@ static void oldify_one (value v, value *p, int promote_stack)
...
+    v = st->todo_list;                 /* Get the head. */
     Assert (Hd_val (v) == 0);             /* It must be forwarded. */
     new_v = Op_val (v)[0];                /* Follow forward pointer. */
     if (Tag_val(new_v) == Stack_tag) {
-      oldify_todo_list = Op_val (v)[1];   /* Remove from list (stack) */
+      st->todo_list = Op_val (v)[1];   /* Remove from list (stack) */
       //caml_gc_log ("oldify_mopup: caml_scan_stack start old=%p new=%p",
       //             (value*)v, (value*)new_v);
-      caml_scan_stack(caml_oldify_one, 0, new_v);
+      caml_scan_stack(oldify_one, st, new_v);
```
stedolan: This is a slight change in behaviour, although one I don't think it's possible to observe. Here, the stack will be scanned with `should_promote_stacks = 0`, if called from `caml_promote`, when previously it had `promote_stack = 1`. Is this OK?

PR #120, opened 2017-04-20
ARM64 backend

kayceesrk:
**DO NOT MERGE YET**

All the `effects` test cases work! There are some failing test cases, which i'm looking into currently. 

- [x] callback/test1.native

CC @dbuenzli @buzzheavyyear


byterun/signals.c
```diff
@@ -76,7 +76,7 @@ void caml_record_signal(int signal_number)
 static caml_root caml_signal_handlers;
 
 void caml_init_signal_handling() {
-  caml_signal_handlers = caml_create_root(caml_alloc(NSIG, 0));
```
avsm: Is this a separate change that needs to go into the x86 version too?

kayceesrk: `caml_alloc` may trigger a GC, but at this point GC is not fully initialized. In particular, the frame tables haven't been inited. By allocating in the shared heap, we ensure that GC cannot be triggered due to a failed allocation. This case was triggered on aarch64 but the fix applies to both architectures.

kayceesrk:
```
Summary:
  466 test(s) passed
    0 test(s) failed
   15 test(s) disabled
    0 unexpected error(s)
```

PR #122, opened 2017-05-01
Inlined mutable reads

kayceesrk:
Inlines the mutable reads. Currently, mutable reads are always read through an external call `caml_read_barrier(value base, int offset)`, which checks whether `obj = base[offset]` is a foreign pointer. If so, the read barrier is invoked which promotes `obj`. In multicore, external calls are expensive since we need to switch from OCaml to C stack, perform the call and switch back to the original OCaml stack. Moreover, if the external call allocates, then extra context information needs to be saved and restored during context switch. However, we expect majority of reads of mutable locations to not cause a read fault. Hence, it is beneficial to inline the check such that the fast path of mutable reads is done inline. This PR does this.

The crux of the idea is from the original [multicore OCaml proposal](https://github.com/ocamllabs/compiler-hacking/wiki/OCaml-Multicore-Runtime#read-barrier-implementation)! We add new cmm and mach instructions `Cloadmut` and `Iloadmut` for mutable loads. 

We reserve two registers for evaluating this instruction (`x0` and `x1` on arm64; `%rax` and `%rdx` on amd64) in order to perform the intermediate computation. Moreover, we ensure that the argument and result registers don't overlap in order to ensure that we have access to the original arguments. This is achieved by adding suitable interference equations. 

In the case of a read fault, we might invoke the GC. Hence, mutable loads need to record the live registers in the frame table. We emit calls to `caml_read_barrier` in this case similar to how GC is invoked on allocation failure.

- [x] amd64
- [x] arm64


kayceesrk:
**Todo**
- [ ] on amd64, do not destroy `%rcx` for temporary. Save `%rcx` on system stack. See the trick in https://github.com/ocamllabs/ocaml-multicore/blob/master/asmrun/amd64.S#L530-L552

Issue #45, opened 2016-04-14
Implement default handlers


kayceesrk:
Default handlers have been implemented for bytecode in ocaml-effects: https://github.com/lpw25/ocaml-effects/tree/default-handlers

kayceesrk:
@dhil you had mentioned that you would like to implement default handlers for multicore. The relevant commits from `ecaml-multicore` are https://github.com/lpw25/ocaml/commit/f098d35e3b7e2b8ed58549236745dfe23ac19a99 and https://github.com/lpw25/ocaml/commit/bd0e22d9517315e4a384dcba49a0bc48818cc820.

dhil:
Yes. I will look into this during the weekend. 

I do not have sufficient privileges on Github to assign the task to myself.

dhil:
I suppose I will have port the syntax and typing changes as well, starting from [lpw25/ocaml@`2f33aad`](https://github.com/lpw25/ocaml/commit/2f33aad1d44b8906646536bdef98938a762fab19).

kayceesrk:
> I do not have sufficient privileges on Github to assign the task to myself.

Done.

dhil:
Yesterday, when we discussed the semantics of default handlers it became clear that we need to take care. After our discussion I went and had a look at my notes. From my notes I have compiled a list of questions.

## Synopsis
A default handler provides a *default* implementation of an effectful operation. An invocation of an operation will be handled by its default if no other handler for that operation is in scope. If a handler for the operation is in scope of the invocation then it overloads the default implementation.

## Proposed syntax 
The proposed syntax is closely aligned with Leo's implementation in ecaml.
At the top level: A default handler is declared along with an operation declaration using the `with function` syntax, e.g. 
```ocaml
effect Foo : int -> int with function
                             | Foo 42 -> 42
                             | Foo i -> 42 - i
```
This declares an operation `Foo` with a default handler (or default implementation) which interprets `perform (Foo i)` as either the constant `42` or the difference `42 - i` depending on the value of `i`.

For modules we have two levels of operation declarations: signature and structure declarations. For the former the proposed syntax is
```ocaml
module type M = sig
     effect Foo : int -> int with function
end
```
This module expose a single operation `Foo` which has a default handler. By contract, any implementation of `M` must provide a default handler for the operation `Foo`.

For structures we adopt the top level syntax, i.e.
```ocaml
module N = struct
   effect Foo : int -> int with function
                                 | Foo 42 -> 42
[...]

PR #123, opened 2017-05-01
Default handlers

dhil:
## Syntax
The syntax for default handlers is adopted from @lpw25 's ecaml-multicore.
```ocaml
effect E : dom -> cod with function E x_1 -> M_1 | ... | E x_N -> M_N
```
This defines an effectful operation `E` with domain `dom` and codomain `cod` along with a default handler which comprises `N` clauses using the standard OCaml `function` syntax.

The syntax for declaring an effectful operation without a default handler remains unchanged, i.e.
```ocaml
effect T : dom -> cod
```

## Semantics

**Operational** In contrast to a regular effect handler, a default handler does not expose the continuation of the performed operation. The main idea is that the body of every clause in a default handler is implicitly wrapped in the continuation. Operationally, a default handler can be viewed as a regular effect handler which encloses the program, i.e.
```ocaml
effect E : dom -> cod 
  with function
  | E x_1 -> M_1
  ...
  | E x_N -> M_N

M'
==
match M' with
| x -> x
| effect (E x_1) k -> try let x = M_1 in continue k x with e -> discontinue k e
| ...
| effect (E x_N) k -> let x = M_N in continue k x  with e -> discontinue k e
```

A default handler for an operation `E` is invoked whenever there is no enclosing handler for `E`. For example
```ocaml
effect E : int -> int with function E i -> i + 1
let y = match perform (E 41) with | x -> x
```
binds `y` to `42`, while
 ```ocaml
effect E : int -> int with function E i -> i + 1
let y = match perform (E 41) with 
   | x -> x
   | effect (E i) k -> continue k (i-1)
```
binds `y` to `40`, because `E` is handled locally.

**Typing** The typing of default handlers is somewhat easier than regular handlers. Every clause body of a default handler must have the same type as the codomain of the operation it is handling.  

**No default handler** Invoking an effect without a default handler raises the `Unhandled` exception at the invocation site, allowing a local exception handler to handle the exception. For example, the following function returns `true`
```ocaml
effect E : unit
let f () = match perform E with 
     | _ -> false
     | exception Unhandled -> true
```

**Effect handling context** A default handler runs outside the effect handling context of the local abstract computation. For example, consider the following program
```ocaml
effect E : unit with function E -> print_endline "E top level"
effect T : unit with function T -> perform E; print_endline "T top level"

let _ = 
  match perform E; perform T with
  | x -> x
  | effect T k -> print_endline "T local"; perform T
```
The invocation of `E` reaches the default handler at the top level, while the first invocation of `T` reaches the local handler which itself invokes `T`, this invocation reaches the top level default handler for `T`. Thus the program prints
```
E top level
T local
E top level
T top level
```
**Recursive handling** Following @lpw25 's "the more recursion, the merrier" suggestion, a default effect handler can recursively perform the effect it is handling, e.g.
```ocaml
effect E : unit with function E -> perform E
```
This default handler is well typed. The performed operation `E` is handled by the same default handler, thus this handler diverges. 

## Implementation
The memory layout of an effectful operation has changed slightly.
Before an effectful operation would be a heap declared object with a tag + two fields, e.g.
```ocaml
effect E : unit
```
would be represented as

| Tag              | Name |  Constant   |
| -----------  | :------: | --------- |
| 248              | "E"        |  0            |    

Now this representation has been extended with two additional fields

| Tag              | Name |  Constant   |  Default handler | Delegate handler |
| -----------  | :------: | --------- | ---------------- | ------------- |
| 248              | "E"        |  0            |  default/number | delegate/number

The default handler component contains the identifier for the default handler. Similarly, the reperformer component contains the identifier for the "delegate" handler which is a special handler that is invoked if the operation `E` has been `reperform`ed. This handler ensures that if `default` returns a value or causes an exception the result gets bubbled downwards the stack correctly. This greatly simplifies both the byte code interpreter and native code implementations of default handlers. This is a design adapted from @lpw25.

Concretely, the following declaration 
```ocaml
effect E : int -> int with function E i -> i + 1
```
produces
```
(letrec
  (E/1028
     (caml_set_oo_id (makemutable 248 "E" 0 default/1032 delegate/1033))
    default/1032
      (function param/1034
        (if (== (field_imm 0 param/1034) E/1028) (field_imm 1 param/1034)
          (raise
            (makeblock 0 (global Match_failure/18g) [0: "//toplevel//" 1 22]))))
    delegate/1033
      (function eff/1036 k/1037
        (catch
          (try (exit -2 (apply default/1032 eff/1036)) with exn/1038
            (resume k/1037 (function x/1041 (raise x/1041)) exn/1038))
         with (-2 v/1039) (resume k/1037 (function x/1040 x/1040) v/1039))))
  (apply (field_imm 1 (global Toploop!)) "E/1028" E/1028))
```

Another simplification is that every effect has a default handler that is even if the programmer omits a default handler for an effectful operation, since the compiler will generate a default handler which raises the `Unhandled` exception. For example, the declaration
```ocaml
effect E : int -> int
```
produces the following lambda code
```
(letrec
  (E/1018
     (caml_set_oo_id (makemutable 248 "E" 0 default/1019 delegate/1020))
    default/1019 (function param/1021 (raise (global Unhandled/30g)))
    delegate/1020
      (function eff/1022 k/1023
        (catch
          (try (exit -1 (apply default/1019 eff/1022)) with exn/1024
            (resume k/1023 (function x/1027 (raise x/1027)) exn/1024))
         with (-1 v/1025) (resume k/1023 (function x/1026 x/1026) v/1025))))
  (apply (field_imm 1 (global Toploop!)) "E/1018" E/1018))
```
This makes the representation of operations at runtime uniform.

## Error messages
Mistyping the operation in the pattern-matching clause produces a helpful error message
```ocaml
# effect E : unit with function T -> ();;
Error: This pattern matches effect T, but the expected effect was E.
```
If the pattern is not on the form `E [arg]` then an error message is produced, for example
```ocaml
# effect E : unit with function x -> ();;
Error: Expected definition of the form: E ... -> ...
```
I think this error message is a bit obscure and may be improved.

## Remaining issues

* [x] Exhaustive pattern matching check
* [x] ~~Arm64 native backend~~ @kayceesrk has agreed to implement this.

## Thanks
Thanks to @lpw25, @kayceesrk, and @stedolan for helpful guidance.


stedolan:
Since this apparently fixes tests/effects/test6.ml, this pull request should delete [these lines from `testsuite/disabled`](https://github.com/ocamllabs/ocaml-multicore/blob/master/testsuite/disabled#L21-L22)

dhil:
@stedolan and I discussed the "default" implementation of default handlers which currently is
```ocaml
let default = function | _ -> raise Unhandled
```
Note that the default handler does not have access to the continuation.

@stedolan suggested that it maybe ought to be
```ocaml
let default k = function | _ -> discontinue k Unhandled
```
That is passing in the continuation and immediately discontinue it with `Unhandled`.

I discussed this with @lpw25 and he had a rather good argument for why we should stick to the former design. Since the default handler does not take a continuation it means in programs that do not use handlers it is very cheap to perform an effect. It amounts to a single projection and a function call.

Whereas if we were to pass in the continuation then we would have to allocate a stack. Consequently, every time a default handler performs an effect we would have to allocate another stack. Thus the cost of default handlers increases significantly.

dhil:
I realised that the syntax allows one to implement a default handler on a signature item
```ocaml
module type M = sig 
  effect E : unit with function E -> ()
end
```
I am not sure what it is supposed to mean. To me it seems somewhat nonsensical. Also in issue #45 we agreed that we would not brother with default handler declarations on signature items.

So in my latest commit I have fixed this such that the above produces an error
```ocaml
# module type M = sig effect E : unit with function E -> () end;;
Error: An effect declaration in a signature cannot have a default handler.
```

stedolan:
Looks good! Sorry it's taken me a while to get around to reading this. I have a couple of questions / comments:

  - What's the extra `bool` argument to `Tpat_construct` for? I can't see where it's used. Is `bool` the right way of expressing this?

  - This patch seems to add default handlers to "extension constructors". While extension constructors are used for effects, they're also used for any open variant declared with `type t = ..` (the effect syntax is shorthand for `type 'a eff += ...`). I think the current patch adds a default handler to all extension constructors, but it should only do that for effects.

It might also be good to get @lpw25 to have a look at this, since he's more familiar with the code.

kayceesrk:
We shouldn't merge this PR yet since it breaks native compilation. Effect declarations now refer to default handler closures, which can't be serialized due to broken marshalling (https://github.com/ocamllabs/ocaml-multicore/issues/75). Hence, compilation of `.cmx` files are broken. We should also add support for Arm64 backend.

dhil:
@kayceesrk I agree, we should wait until closures marshalling has been fixed.

@stedolan the extra bool field is actually @lpw25's idea. This fixes the false positive exhaustivity warnings that were being generated by default handlers. A default handler is essentially a `match` on the operation, however because operations themselves are implemented as constructors of the extensible variant `eff` pattern matching on operation constructors always requires `| _ -> assert false` case. The typing of a default handler permit it only to match on the constructor of the operation it handles, so we ought only to check whether the constructor pattern is total and we ought not check whether the match is exhaustive for the variant `eff`. Thus the bool field is meant to indicate whether to short-circuit the latter check for the constructor. It is used in [`parmatch.ml#L625`](https://github.com/dhil/ocaml-multicore/blob/1a9e595dae98344b60fa478f7f994c0bc5576cb5/typing/parmatch.ml#L625).

I discussed design alternatives with @lpw25, but we eventually arrived at his original approach. If we had refutable patterns, I think it would be nice to out your suggestion which was to have the compiler add two extra clauses to the match, i.e.
```ocaml
match op with
| E pat -> M
| E _ -> .
| _ -> assert false
```

stedolan:
OK, that makes sense. Would you mind changing the type, though? It's not at all obvious what `true` means here, and I think a new datatype with more meaningful case names would help.

dhil:
@stedolan Sure, I will be happy to revisit it once I am back home from Japan.

dhil:
@stedolan I have implemented what you suggested and I have synchronised this PR with the master branch. 

Following your suggestion I have introduced a regular data type `constructor_completeness = Check | Complete` which is intended to tell the exhaustivity checker whether a given constructor pattern is "assumed" to be exhaustive. Ideally, I would like to use a polymorphic variant for this, but I have not seen any use of polymorphic variants in the compiler itself. I do not know whether this is due to some unwritten policy. If it is not a problem, then I would like to change the code to use polymorphic variants as it would get rid of some boilerplate code.

The merge with the master branch introduced a small regression in the assembly code in `amd64.S` which was due to KC's dwarf PR #134 which claims labels 118 and 119 which I happened to be using for default handlers too. I simply bumped my use of 118 and 119 to 120 and 121 respectively. I am not aware of any numbering scheme for labels.

kayceesrk:
I am for closing this PR for now. Parts of this work will have to be redone with effect typing, but cleanly separating the contribution isn't worth the effort right now.

dhil:
> Parts of this work will have to be redone with effect typing

I seem to recall that we have already implemented. It is definitely features on the project plan for the branch.

> I am for closing this PR for now.

OK. Let's close it.

PR #127, opened 2017-05-13
Fix CFI in amd64.

kayceesrk:
Gdb now reports sane backtraces at "every" instruction. The intended behaviour is that the backtrace stops at 

  - top of fiber stack or
  - bottom of C frame at external calls (`caml_c_call`) or
  - `caml_start_program` when setting up the main OCaml program or callbacks. 

Being sane at "every" instruction ensures that code that might unwind the stack, such as profilers do not segfault due to corrupted frames at the bottom of the stack. gproftools/cpu-profiler now works on multicore!

The main trick is to ensure that the return address of the last frame is `NULL`. Debuggers and other unwinding code treats them as a signal to stop unwinding. We do this by carefully constructing CFI information such that they point to a known location which is guaranteed to be NULL. See `OCAMLCFAXXX` and `SYSCFAXXX`.

As a part of the goal to get sane cfi at every instruction, `SWITCH_OCAML_TO_C` and `SWITCH_C_TO_OCAML` are a few instructions longer to avoid clobbering `%rsp` during switching and losing access to CFA.

### DIFF
--- a/asmcomp/amd64/emit.mlp
+++ b/asmcomp/amd64/emit.mlp
@@ -473,14 +473,19 @@ let emit_instr fallthrough i =
           `	{locate_domain_state ()}, %r15\n`;
           `	movq	{domain_field Domainstate.Domain_young_ptr}(%r15), %r15\n`;
         end else begin
-          ` movq  %rsp, %r10\n`;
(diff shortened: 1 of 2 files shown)


stedolan:
Nice!

PR #130, opened 2017-05-16
Domain.join implementation

edwintorok:
`Domain.spawn` now returns the `pthread_t` wrapped in a custom block.
`Domain.join` simply calls `pthread_join`.

pause had to be removed to avoid the thread hanging indefinetely.


stedolan:
Looks good! The call to `pause` is from a time when domains referred to each other's thread local variables (!), a poor design now fixed, so it can go away.

You do need some means of handling `Domain.join` being called multiple times, or in parallel by different domains - `pthread_join` has undefined behaviour if called twice. I'm not sure whether subsequent calls to `Domain.join` should be no-ops or raise an exception, but it should do something well-defined.

byterun/domain.c
```diff
@@ -324,16 +336,31 @@ CAMLprim value caml_domain_spawn(value callback)
...
+CAMLprim value caml_ml_domain_join(value domain)
+{
+    CAMLparam1 (domain);
+    pthread_t *th = (pthread_t*) Data_custom_val(domain);
+    int err;
+
+    if (!th)
+        caml_invalid_argument("No thread");
+    caml_gc_log("Domain joining");
+    err = pthread_join(*th, NULL);
```
stedolan: This is a slow operation, so it needs to be in a blocking section (surrounded by `caml_enter_blocking_section` and `caml_leave_blocking_section`). Because `*th` can move, the dereference must be outside the blocking section. (The easiest way to do this is to write `pthread_t th = *(...)` instead of `pthread_t *th = (...)`)

edwintorok:
Thanks, I have added a mutex protected flag to detect multiple joins and exit early by raising an exception. Since I'm now storing a mutex inside the custom block I added another allocation (and a finalizer), I'm worried that OCaml's GC could move the custom block with the mutex in it which wouldn't be well defined.

Not sure whether I should just use `caml_plat_lock/unlock` instead of `With_mutex`, the mutex is only held for a short period of time to check a flag.

stedolan:
> Thanks, I have added a mutex protected flag to detect multiple joins and exit early by raising an exception. Since I'm now storing a mutex inside the custom block I added another allocation (and a finalizer), I'm worried that OCaml's GC could move the custom block with the mutex in it which wouldn't be well defined.

You're right, you do need a means to prevent the mutex moving. The standard idiom here is to allocate the `struct domain_thread` yourself, using `caml_stat_alloc`, and then only store a pointer to that structure in the custom block. That way, the GC can move the custom block but won't move the structure.

> Not sure whether I should just use caml_plat_lock/unlock instead of With_mutex, the mutex is only held for a short period of time to check a flag.

`caml_plat_lock/unlock` is fine. `With_mutex` includes some magic to ensure that the lock is released if an exception is thrown while the mutex is held, which is necessary for some hairy code in `io.c`. It's not needed here.

edwintorok:
Thanks, pushed a fixup commit.

byterun/domain.c
```diff
@@ -300,19 +299,61 @@ static void* domain_thread_func(void* v) {
...
+  *dp = d = caml_stat_alloc(sizeof(*d));
+  caml_plat_mutex_init(&d->m);
+  d->joinable = 1;
 
   caml_plat_event_init(&p.ev);
 
   p.callback = caml_create_root(caml_promote(&domain_self->state, callback));
 
-  err = pthread_create(&th, 0, domain_thread_func, (void*)&p);
+  err = pthread_create(&d->th, 0, domain_thread_func, (void*)&p);
```
stedolan: If this returns an error, then `d->th` has not been initialised. The finaliser will run, and since `joinable=true`, it will try to `pthread_detach(d->th)`.

I think `joinable` should be initialised to `0`, and only set to `1` if `pthread_create` succeeds.

edwintorok: Good point, pushed a fixup for that.

byterun/domain.c
```diff
@@ -324,16 +366,43 @@ CAMLprim value caml_domain_spawn(value callback)
...
+    CAMLparam1 (domain);
+    struct domain_thread **dp = Domainthreadptr_val(domain);
+    struct domain_thread *d;
+    int err = 0;
+    pthread_t th;
+
+    Assert(dp);
+    d = *dp;
+    Assert(d);
+    th = d->th;
```
stedolan: Here, `d->th` is accessed without holding the mutex `d->m`. It's not wrong, because `d->th` is readonly except in the finaliser, with which no code can race. However, the locking setup isn't obvious.

You should do one of:
  - Move this line after `caml_plat_lock` below so that `d->m` protects all accesses to `d`, or
  - Add a comment in the definition of `struct domain_thread` explaining that the mutex protects only `joinable` but not `th`, and that `th` is immutable once initialised.

edwintorok: I've added a comment, I think it is clearer that way. Otherwise it could be confusing why copying the `pthread_t` needed the lock, but operating on it (with `pthread_join`) didn't.

stedolan:
Sorry for the delay, I've been quite busy the last while. This version looks good, thanks!

PR #135, opened 2017-05-28
[WIP] Fix Continuation patterns are too strict

vramana:
```ocaml
effect E : int
let run f =
  match f () with
  | s ->
     print_string s
  | effect E (k : (int, string) continuation) ->
     continue k 42
```
Currently fails with 
```
Error: This pattern matches values of type (int, string) continuation
       but a pattern was expected which matches values of type
         (effect#0, unit) continuation
       Type int is not compatible with type effect#0 
```

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -3564,9 +3564,33 @@ and type_cases ?in_function env ty_arg ty_res ?conts partial_flag loc caselist =
     match conts with
     | None -> List.map (fun (pat, env) -> (pat, env, None)) pat_env_list
     | Some conts ->
-        List.map2
-          (fun (pat, env) cont -> (pat, env, cont))
-          pat_env_list conts
+        let ty = newvar () in
+        (* remember original level *)
+        (* Create a fake abstract type declaration for effect type. *)
+        let level = get_current_level () in
+        let decl = {
+          type_params = [];
+          type_arity = 0;
+          type_kind = Type_abstract;
+          type_private = Public;
+          type_manifest = None;
+          type_variance = [];
+          type_newtype_level = Some (level, level);
+          type_loc = loc;
+          type_attributes = [];
+        }
+        in
+        Ident.set_current_time ty.level;
+        let name = Ctype.get_new_abstract_name "effect" in
+        let (id, new_env) = Env.enter_type name decl env in
+        Ctype.init_def(Ident.current_time());
+        let ty_eff = newgenty (Tconstr (Path.Pident id,[],ref Mnil)) in
+        let ty_cont = Predef.type_continuation ty_eff ty_res in
+        let pat_env_cont_listt = List.map2
+            (fun (pat, env) cont -> let (envv, _) = env in
+              (pat, env, type_continuation_pat envv ty_cont cont))
+          pat_env_list conts in
+        pat_env_cont_listt
   in
   let cases =
     List.map2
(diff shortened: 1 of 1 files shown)


vramana:
@stedolan @kayceesrk I don't know how to move forward. My guess is these lines https://github.com/ocamllabs/ocaml-multicore/blob/master/typing/typecore.ml#L3637-L3640 have to adjusted. But I am not sure about that.

Are they helper functions to print out types? As the type checker runs, I want to see what types it's seeing.  Also any tips on how to debug type checker will also be helpful.

lpw25:
>  My guess is these lines ...

Actually the issue is more involved that that. Essentially `type_continuation_pat` is called with the correct types but in the wrong environment -- it needs the equation `effect#0 = int` which comes from matching on `E`.

So instead of calling `type_continuation_pat` and then `type_cases` it needs to call `type_continuation_pat` from within `type_cases`. If you look [here](https://github.com/ocamllabs/ocaml-multicore/blob/master/typing/typecore.ml#L3547) you have the continuation and the environment with the equations, so calling `type_continuation_pat` there should work.

lpw25:
> Are they helper functions to print out types?

You can use `Printtyp.raw_type_expr`. There is also a forward reference to that in `Btype.print_raw` which is useful if the file you are in is earlier that `Printtyp` in the dependency graph.

typing/typecore.ml
```diff
@@ -3543,9 +3564,33 @@ and type_cases ?in_function env ty_arg ty_res ?conts partial_flag loc caselist =
...
+        in
+        Ident.set_current_time ty.level;
+        let name = Ctype.get_new_abstract_name "effect" in
+        let (id, new_env) = Env.enter_type name decl env in
+        Ctype.init_def(Ident.current_time());
+        let ty_eff = newgenty (Tconstr (Path.Pident id,[],ref Mnil)) in
+        let ty_cont = Predef.type_continuation ty_eff ty_res in
+        let pat_env_cont_listt = List.map2
+            (fun (pat, env) cont -> let (envv, _) = env in
+              (pat, env, type_continuation_pat envv ty_cont cont))
```
vramana: @lpw25 How do I construct `ty_cont` that's passed to  here? I just tried to copy how it's done in `type_effect_cases`.

This is the current error message I am getting

→ OCAMLRUNPARAM=b  ./ocaml-multicore/boot/ocamlrun ./ocaml-multicore/ocamlc -nostdlib -I ./ocaml-multicore/stdlib -dtypedtree p.ml
File "p.ml", line 9, characters 16-18:
Error: This expression has type int but an expression was expected of type
         effect#1

lpw25: You should probably construct it where it was constructed before and then pass it into `type_cases` as part of the `conts` parameter.

vramana: Okay. It will try that.

vramana:
@lpw25 It compiles 😄  I have passed `conts` and `ty_cont` through the same argument but I couldn't find a better argument name so I left it as `conts` in `type_cases`.

typing/typecore.ml
```diff
@@ -908,7 +908,28 @@ let type_continuation_pat env expected_ty sp =
           Types.val_loc = loc; val_attributes = []; }
       in
         Some (id, desc)
-  | Ppat_extension ext ->
+  | Ppat_constraint ({ppat_desc=Ppat_var name; ppat_loc=lloc},
```
vramana: @lpw25 Here, I have restricted `ppat_desc` to just `Ppat_var`. Am I missing any other valid cases?

lpw25: `Ppat_constraint ({ppat_desc=Ppat_any; }, _)` should probably also be allowed.

vramana:
I have commented out the case with constraint in my test file, so it compiled. I just noticed it today. This is the error that I am getting now.
```
File "p.ml", line 7, characters 14-15:
Error: This pattern matches values of type (effect#0, string) continuation
       but a pattern was expected which matches values of type
         (effect#0, unit) continuation
       Type string is not compatible with type unit 
```

lpw25:
That's the correct error message

typing/typecore.ml
```diff
@@ -908,6 +908,31 @@ let type_continuation_pat env expected_ty sp =
...
+      let ty, expected_ty' =
+        if separate then begin
+          end_def();
+          generalize_structure ty;
+          instance env ty, instance env ty
+        end else ty, ty
+      in
+      unify_pat_types lloc env ty expected_ty;
+      pattern_force := force :: !pattern_force;
+      (match ddesc with
```
stedolan: Instead of a nested match, would it be better for this to be a recursive call to `type_continuation_pat`?

vramana: Yeah, That should work on think.

vramana:
@stedolan @lpw25 Can you review the PR?

PR #139, opened 2017-07-06
Allow type constraints in continuation patterns

### LINKED ISSUE #126: Continuation patterns are too strict
Currently, in the syntax for handling effects `effect Epat Kpat`, only variables and `_` are allowed as the pattern to match a continuation.

I think e.g. `(k : (unit, int) continuation)` should also be allowed, since it can be useful to be explicit about the type of the continuation.

(This might be a good first task for someone who wants to explore the type checker)

Here's an example program that should work but doesn't:
```
effect E : int
let run f =
  match f () with
  | s ->
     print_string s
  | effect E (k : (int, string) continuation) ->
     continue k 42
```

little-arhat:
Allow using `(k : (int, unit) continuation)` patterns in effect handlers. Hopefully fixes #126.


Drup:
Just FYI, this is not the right way to implement such feature. You should mostly never pattern match on type expressions. Instead, you should build an expression `('a, 'b) continuation` and unify it with the provided constraint. This allow to handle code like:

```
type t = (int, float) continuation
(* ... *)
  | effect E (k : t)
(* ... *)
```

It also removes the need for a special error (instead, you get a standard unification error).

kayceesrk:
@lpw25 thoughts on this PR?

lpw25:
I should point out that the code is not just too restrictive, it is also incorrect because the expected type is ignored in the constraint case.

`type_continuation_constraint` should look more like:

```ocaml
let type_continuation_constraint env loc expected_ty sty =
  let cty = Typetexp.transl_simple_type env false sty in
  unify_pat_types loc env cty expected_ty
```

kayceesrk:
Closing this issue as OCaml 5.00 will not have syntax for effect handlers.

Issue #137, opened 2017-07-04
Crash during GC


alexeyshch:
[This program](https://gist.github.com/alexeyshch/0ac467fa66fa4a2cd8a5e6489fe84548) crashes with SIGSEGV after a fraction of a second.  If Gc.minor call is removed from Scheduler.enqueue, then it works fine.  

Backtrace:
```
#0  0x000000000046a695 in mark_normalise ()
#1  0x000000000046ad57 in caml_darken ()
#2  0x0000000000468bba in caml_scan_stack ()
#3  0x000000000046ae96 in mark ()
#4  0x000000000046b12b in caml_major_collection_slice ()
#5  0x000000000046bd0a in caml_minor_collection ()
#6  0x00000000004638ee in caml_gc_minor ()
#7  <signal handler called>
#8  0x0000000000469c4e in caml_c_call ()
#9  0x0000000000407e71 in camlMctest__enqueue_1036 () at mctest.ml:21
#10 0x0000000000407a6a in camlMctest__fun_1154 () at mctest.ml:49
#11 <signal handler called>
#12 0x0000000000469e77 in caml_start_program ()
#13 0x0000000000464ebf in caml_callback_exn ()
#14 0x0000000000465107 in caml_callback ()
#15 0x00000000004672d3 in domain_thread_func ()
#16 0x00007ffff76c20a4 in start_thread (arg=0x7ffff7310700)
    at pthread_create.c:309
#17 0x00007ffff73f6ccd in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

Compiled with `ocamlfind ocamlopt -g -linkpkg -package "unix lockfree" mctest.ml -o mctest`

kayceesrk:
Thanks for the report. I am looking into this.

bluddy:
In bytecode mode, the program runs and prints, but doesn't terminate. Is that expected behavior?

kayceesrk:
The program terminates in native-code in non-buggy runs. I'd expect the program to terminate in bytecode mode as well.

bluddy:
Well, I'm getting the segfault in native mode, and it prints everything and hangs in bytecode mode. This is on OSX.

alexeyshch:
For me in Linux it doesn't terminate in native mode.  You can make segfaults rare by changing the frequency of yields e.g. by setting these numbers of iterations in the `worker` function:
```ocaml
      for i = 1 to 100 do
[...]

PR #141, opened 2017-07-18
Introduce synchronization for marking fibers.

kayceesrk:
Race between a mutator racing to context switch to a fiber and a GC
thread on a different domain racing to darken the stack is unsafe. This
is because the mutator may modify the stack as the GC is walking it.
This commit introduces necessary synchronization.

Before marking a clean fiber (one that does not have pointers into any
minor heap), its status is updated to scanning. If a mutator finds that
the target fiber of a context switch is currently being scanned, it spin
waits until the scanning is complete.

If the GC thread finds a dirty fiber, and is owned by the current
domain, then it goes ahead and scans it. This is safe because a different
domain context switching to such a fiber will have to request the domain
that owns the fiber to promote it.

If a GC thread finds a dirty fiber and the fiber is owned by a different
domain, the fiber is skipped. Such a fiber will be darkend at the end of
minor GC by the domain that owns it.


kayceesrk:
Fixes #137.

PR #143, opened 2017-07-27
Collect detailed statistics

kayceesrk:
This PR adds support for

1. Collecting detailed statistics on program execution
2. Displaying the stats at the end of program execution. 

A new configure time flag `-enable-stats` has been added which turns on detailed statistics collection. The programs are expected to be a bit slower with detailed statistics enabled. But the statistics collected are counts and sizes, which should be unaffected by slower running time. The statistics collected are number of allocations, mutable/immutable loads/stores, various kinds of calls (tail/nontail indirect/immediate), external calls (noalloc, alloc, alloc with stack arguments) and stack overflow checks.

A new `OCAMLRUNPARAM` argument `S` has been added, which when set displays the GC stats, and the detailed stats above (if detailed stats had been enabled at compile time). A sample output for compiling a simple program is shown below:

```
$ OCAMLRUNPARAM="S" ocamlopt.opt hello.ml
**** GC stats ****
Minor words:            257209
Promoted words:         56492
Major words:            0
Minor collections:      1
Major collections:      1

**** Other stats ****
Allocations:            50049

Loads:                  493161
Mutable:                55157 (11.18%)
Immutable:              438004 (88.82%)

Stores:                 44746
Mutable:                37827 (84.54%)
Immutable:              6919 (15.46%)

External calls:         54037
NoAlloc:                30235 (55.95%)
Alloc:                  23802 (44.05%)
Alloc + stack args:     0 (0.00%)

Calls:                  154931
Imm tail:               44379 (28.64%)
Ind tail:               23415 (15.11%)
Imm non-tail:           79357 (51.22%)
Ind non-tail:           7780 (5.02%)

Stackoverflow checks:   88833 (57.34%)
```


byterun/caml/memory.h
```diff
@@ -79,6 +79,12 @@ color_t caml_allocation_color (void *hp);
 #define DEBUG_clear(result, wosize)
 #endif
 
+#ifdef COLLECT_STATS
+#define Count_alloc dom_st->allocations++
+#else
+#define Count_alloc while(0)
```
stedolan: I get what's going on here, but at first it looked like some magic was happening. `#define Count_alloc (void)0` or just `#define Count_alloc` are more idiomatic no-ops.

byterun/domain.c
```diff
@@ -737,6 +738,80 @@ void caml_sample_gc_stats(struct gc_stats* buf)
...
+#ifdef COLLECT_STATS
+  struct detailed_stats ds;
+  caml_domain_state* st;
+  uint64 total;
+#endif
+
+  caml_gc_stat(Val_unit);
+  caml_sample_gc_stats(&s);
+  fprintf(stderr,"**** GC stats ****\n");
+  fprintf(stderr, "Minor words:\t\t%llu\n", s.minor_words);
```
stedolan: I wrote a comment about the portability of `%llu`, but the portable version is really ugly, and `%llu` is widespread enough these days that the current code is probably fine. If you care, here's the comment:

`%llu` is slightly dodgy for `uint64`. C99 specifies `ll` for `long long`, which is generally but not necessarily 64 bits. Historically, MSVC demands `I64` (e.g. `%I64u`) for 64-bit numbers, but I think it supports `ll` now as well.

The portable way to do this uses `ARCH_INT64_PRINTF_FORMAT` from config.h:

    fprintf(stderr, "Minor words:\t\t%"ARCH_INT64_PRINTF_FORMAT"u\n", s.minor_words);

(note the quotes and the `u`)

stedolan:
Nice! These numbers will be interesting.

Also, I'm now very eager to do the %r14 globals fix - the stats code is unnecessarily fiddly without it.

byterun/domain.c
kayceesrk: I did find the `ARCH_INT64_PRINTF_FORMAT` specifier. In my initial reading, I missed seeing the `u` and incorrectly assumed that it was printing signed integers. I'll leave this as `llu` for now.

byterun/caml/memory.h
kayceesrk: Changed to `#define Count_alloc`.

Issue #118, opened 2017-04-07
Implement `Domain.wait` and `Domain.notify` mechanisms


kayceesrk:
*TODO: describe the API.*

Related to #117.

kayceesrk:
Fixed by #144.

Issue #117, opened 2017-04-06
Domain.wait/notify for Multi-domain Coordination


dlreeves:
I work on the Hack type checker which is written in OCaml. I'm doing some initial investigation into multicore OCaml to see if it would be a suitable replacement down the line for our current method of parallelization via spawning multiple processes. We have a master process that farms out work to several worker processes. The worker processes essentially wait for jobs to execute by using unix select.

When translating this architecture to the multi-domain world one of the first things I encountered was trying to determine how to have a worker efficiently wait for jobs to execute. Ideally I would like to write something like:

```
while true do
  if job_queue_is_empty then
    Domain.wait ()
  else
    (* do work *)
done
```

Then the master could let a worker know there are jobs ready to execute by calling `Domain.notify domain_id`.

Is there already a way of achieving this level of coordination between domains? If not are there any plans to add such primitives?

kayceesrk:
Hi @dlreeves. `Domain.wait` and `Domain.notify` seem like an abstraction that we should provide in multicore. For now, a polling mechanism such as [0] should do the trick. I have added a separate issue to track this feature #118. 

For the job queues, I'd suggest https://github.com/kayceesrk/lockfree which provides a lock-free queue implementation. You can install this library using [multicore-opam remote](https://github.com/ocamllabs/multicore-opam).

[0] https://github.com/kayceesrk/kcas/blob/master/src/kcas_backoff.ml

stedolan:
We're definitely committed to providing a bunch of synchronisation primitives, including both low-level domain-based ones (mainly useful for projects like Hack that already have a threading model and just want to port it to multicore), and higher-level task-parallel ones.

Having said that, it probably won't be exactly `Domain.wait` and `Domain.notify`. With `wait`/`notify`-style interfaces, it's very difficult to get rid of the race condition between `job_queue_is_empty` and `Domain.wait ()` (what if the notification arrives just before the wait begins?). But there will definitely be a way of implementing this pattern.

kayceesrk:
@dlreeves, @dhil is interested in looking at parallelizing Hack using multicore. I remember you mentioning that you had clear info about where changes were needed. If you have some notes, could you share them?

kayceesrk:
Fixed by #144. See [domain.mli](https://github.com/ocamllabs/ocaml-multicore/blob/master/stdlib/domain.mli#L29).

Issue #102, opened 2017-01-20
Expose support for work-stealing


kayceesrk:
When it comes to load balancing work between multiple domains, asking nicely is better than stealing [0]. Specifically, it is better to keep the per-domain deques private to the domain and let the domains request other domains for work. The advantage of private deques is that operations on the deque do not require any synchronization. Since stealing should be a rare event, this design improves performance.

The runtime system has an RPC mechanism implemented over nursery allocation checks; by setting the allocation limit pointer to the start of the nursery, the allocation fails at which point it runs the RPC request. The same mechanism can be used to steal work from other domains. How best to expose this mechanism to the user?

## References

[0] [Scheduling Parallel Programs by Work Stealing with Private Deques](http://www.chargueraud.org/research/2013/ppopp/full.pdf)
[1] [Latency-Hiding Work Stealing](https://13a75b74-a-62cb3a1a-s-sites.googlegroups.com/site/umutacar/publications/spaa2016-full.pdf?attachauth=ANoY7coJchzhhFfB9j97fA4OiM0cjXiLoNf43PTvdjT7hEAGY4d6zZoE5UavrGWvrMqkRGdhW3qDSKvutWk2DbTAeVsFwA4YelJEjuGo5DAzmC0amtQjvoBBxs-OwUtRdFwRn6UQVvIBgkOMVxFCnrAORlvjGsLCU9Wef78phqJvQ_76-44WZkfJv9DRkBddWB6HfMu9bOy_biXPJy0YTiqCsVuckYwojg1m18LUiaoIr-IFWR3jYmI%3D&attredirects=0)
[2] [A Work-Efficient Algorithm for Parallel Unordered Depth-First Search](https://13a75b74-a-62cb3a1a-s-sites.googlegroups.com/site/umutacar/publications/sc2015-full.pdf?attachauth=ANoY7cqXEXVX4H-l3YvA0PqQygLMLj3BWXqCcBdQj2-2WMWjpbTMjJO7XDCjoP_obSwLKKlEI-VXCJsNRsqk-4-W9nuSjNUDZy73PfcHkr345ZLiG8hEMIuDA_z4bkG9lG2mDdw2xeZudVU-mlx-MuFbI2FKVzbisbmBr2rkB1wm_jwCYldKeFR3TaSdLfOizPQ5ziG2DJBEgzKVFjt6RlRQMqqnKvsrYeF-lwn4t-MejK-Mqf_cZjs%3D&attredirects=0)

[...]

Issue #100, opened 2017-01-20
Implement Systhreads in terms of domains


kayceesrk:
The systhread implementation is currently only safe when there is a single domain. The implementation aims to be backwards compatible with vanilla OCaml. Eventually, systhreads should be deprecated, with possibly a backwards compatible implementation over domains.

lpw25:
I don't think that systhreads should be deprecated. I think having multiple threads inside a single domain is the right way to implement asynchronous versions of synchronous system operations. It is probably useful in other situations as well.

Of course it should be made clear in documentation that domains are how to get parallelism. Since the name "threads" is pretty firmly connected to parallelism, it may even be necessary to rename the threads library to get the point across, but the functionality should still be available.

lpw25:
I think it would be useful to have a vaguely concrete implementation
plan about how to do this, and also how to handle asynchronous effects
and preemption. So here is my attempt to outline a design that I think
will work.

The aims of this design are:

1. Support the existing `Thread`s API
2. Support using "asynchronous" effects to do preemptive schedulers and
   handle things like signals
3. Remove from OCaml any genuine asynchronous exceptions -- this makes
   reasoning about resource management much simpler/possible.

Perhaps, those aren't actually the aims we need, but they are some
things that I'm most interested in, so that's what I've started from.

## Threads

The first step is to add a concept of "thread" to the multicore
runtime. These are not the current heavyweight threads tied to system
threads. They are a lightweight concept based on the existing
continuations.

A thread that is not currently running is represented by just a
continuation. Although continuations/stacks would need to be extended to
include information about which asynchronous events (e.g. signals,
alarms for preemptive schedulers) can be handled and how they should be
handled.

Unlike an ordinary continuation created when an effect is performed,
this continuation may be at a location that is not "exception-safe". By
"exception-safe" I mean a place where an effect or exception can safely
be performed/raised. As such, this continuation should never be
discontinued, it may only be continued. The continuation is always at a
point that expects `()` so there is essentially just a simple resume
operation on threads that takes no other arguments.

At any time a domain may have multiple threads scheduled to run. If there
are multiple threads then they are scheduled preemptively, using similar
heuristics to the current system threads implementation for when to
switch threads.

My reasoning for including another concept alongside effect
continuations is that when multiple threads are created using the
`Thread`s API we need to run them concurrently even when there is no
user-level scheduler provided. You can imagine that there is a built-in
user-level scheduler surrounding the whole program and the threads API
[...]

PR #144, opened 2017-07-28
Refactor domain interrupts and synchronisation

stedolan:
Massive refactor of domain synchronisation logic.

The new features is that domain termination now works, and domains can yield without spinning.

At domain termination, the minor heap is emptied, and active shared heap pools are moved into a global list, so that some other domain will pick them up and start sweeping them.

Yielding uses the new primitives yield, interrupt and critical sections, and is implemented via condition variables rather than atomics and spinning (so that yielding does actually sleep).


stedolan:
The only test failure is `Domain.join`, which is disabled in this code. (There's now enough support to implement `Domain.join` in OCaml, but I haven't done that yet)

kayceesrk:
With this PR, as I undestand, these mechanisms can be implemented in OCaml rather than C: https://github.com/ocamllabs/ocaml-multicore/issues/100, https://github.com/ocamllabs/ocaml-multicore/issues/118, https://github.com/ocamllabs/ocaml-multicore/issues/117, https://github.com/ocamllabs/ocaml-multicore/issues/102, https://github.com/ocamllabs/ocaml-multicore/issues/73. These are unnecessary: https://github.com/ocamllabs/ocaml-multicore/issues/90. @stedolan, would be useful if you implement https://github.com/ocamllabs/ocaml-multicore/issues/100 as an example.

stedolan:
> With this PR, as I undestand, these mechanisms can be implemented in OCaml rather than C: #100, #118, #117, #102, #73. These are unnecessary: #90. @stedolan, would be useful if you implement #100 as an example.

Yes to all except #73. We can't support preemption until it's possible to make a continuation in an arbitrary possibly-callback context, which requires doing #146 first.

kayceesrk:
`parallel/join` test fails. But merging now as we will reimplement join in OCaml.

Issue #67, opened 2016-11-01
[yuasa barrier] Avoid mutator blocking on stack scanning


kayceesrk:
Under yuasa (deletion) barrier design, the all the fiber stacks need to be marked at the start of the GC cycle. We can perform this incrementally by scanning the non-black fiber stack on context switching to it. However, this scanning operation still blocks the mutator and can add noticeable  latency for large stacks.

This proposal [0] for Golang obviates STW stack rescanning by enhancing the write barrier to take into account whether the current stack is black. This design avoids the non-incremental stack scanning in the current design by paying instead for slower write barrier. 

The main advantage with the scan-on-context-switch design is that non-concurrent OCaml code does not pay any additional cost (no need to check for colour of the current stack in the write barrier). I would think this to be important for performance backwards compatibility. The disadvantage is that context switching may incur a stack scan. This seems like a reasonable choice since we expect fewer context switches than writes. The stack scanning on context switching is only needed once per major GC cycle, that too if the stack hasn’t already been scanned by the GC.

Since the change to get new write barrier design is minimal, it would be worthwhile to implement and benchmark this scheme.

[0] https://github.com/golang/proposal/blob/master/design/17503-eliminate-rescan.md

stedolan:
On Tue, Nov 1, 2016 at 8:53 AM, KC Sivaramakrishnan <
notifications@github.com> wrote:

> Under yuasa (deletion) barrier design, the all the fiber stacks need to be
> marked at the start of the GC cycle. We can perform this incrementally by
> scanning the non-black fiber stack on context switching to it. However,
> this scanning operation still blocks the mutator and can add noticeable
> latency for large stacks.
>
> This proposal [0,1] for Golang obviates STW stack rescanning by enhancing
> the write barrier to take into account whether the current stack is black.
> This design avoids the non-incremental stack scanning in the current design
> by paying instead for slower write barrier.
>
> The main advantage with the scan-on-context-switch design is that
> non-concurrent OCaml code does not pay any additional cost (no need to
> check for colour of the current stack in the write barrier). I would think
> this to be important for performance backwards compatibility. The
> disadvantage is that context switching may incur a stack scan. This seems
> like a reasonable choice since we expect fewer context switches than
> writes. The stack scanning on context switching is only needed once per
> major GC cycle, that too if the stack hasn’t already been scanned by the GC.
>
I agree. I'm happy with scan-on-context-switch. If latency becomes an
[...]

Issue #41, opened 2016-04-14
Switch to Yuasa barrier from Dijkstra barrier


kayceesrk:
This eliminates the small stop-the-world phase at the end of a major collection to finish the marking work. Stop-the-world phase is only required to flip the colors at the end of collection.

bobot:
Funnily I opened an issue in the ocaml bugtracker for going in the other way http://caml.inria.fr/mantis/view.php?id=7303 . Because Dijkstra barrier is better for weak pointers, the GC doesn't have to mark them. But indeed that adds a small stop-the-world phase.

kayceesrk:
Yuasa has a subtle race condition in the presence of fibers. For a good background on the different barrier types, see [0]. 

Yuasa barrier is a deletion barrier that marks the object that is overwritten. This is implemented in the write barrier. What does this mean for GC roots (registers and stack)? Unlike objects on the heap, register updates and stack pops are not protected by write barrier. Hence, yuasa implementations typically mark the objects in the registers and the stack at the start of the GC cycle. 

The above strategy works for vanilla OCaml which only has a single stack. Under multicore OCaml, we have many stacks in the form of fibers. If we were to naively apply the strategy of scanning all the roots at the start of the GC cycle, we would end up with a long STW phase at the start of the GC cycle in order to scan all of the fibers. This design also mandates that we can easily access all of the fibers on the heap without scanning the entire heap; we would need some auxiliary data structure of maintaining a list of available fibers, which needs to be appropriately GCed when fibers die. This design hence is undesirable. 

Instead, what multicore OCaml does is that it lazily scans the fibers before a context switch. That is, before switching to the fiber, we mark all objects on the fiber (essentially making the fiber black). 
However, this scheme works only if we have a single GC thread. Consider the case where multiple GC threads concurrently mark the heap, while mutator threads are running OCaml code. In this case, it is possible that a GC thread reaches the fiber as a part of the marking process, marks the fiber grey and adds it to the GC thread's mark stack. Now it is unsafe to switch control to this fiber due to the risk of unsafe concurrent access by the GC thread which scans the stack and a mutator thread which runs on the same stack.

[...]

Issue #20, opened 2015-07-28
Quiescent phase in major GC


kayceesrk:
Avoid starting a new major GC immediately after the previous one ends. Implement quiescent phase for major GC. See https://github.com/ocamllabs/ocaml-multicore/commit/23e5bf0249500484e2ece5595d423ca03acbb6b6#commitcomment-12341271.

kayceesrk:
Some more details. At the end of a major GC phase, the domains synchronize and agree to the end of the current GC cycle. At this point, the meaning of GC color bits are flipped such at all objects which are `UNMARKED` become `GARBAGE` and `MARKED` become `UNMARKED`. After this under Yuasa barrier, new objects are allocated MARKED. As such newly allocated objects which have become unreachable survive a major GC (`MARKED` -> `UNMARKED` in the current cycle, `UNMARKED` -> `GARBGE` in the next one). This might be undesirable. 

With quiescence, a distinction is made between GC being `Idle` and `Marking`. After the end of a major GC cycle, the new GC starts off in an `Idle` phase where new objects are allocated `UNMARKED`. At some point, the domains must agree to move to `Marking` phase where new objects are allocated `MARKED`. 

The advantage of this scheme is that any object that was allocated in the `Idle` phase that has become unreachable in this cycle is discarded as GARBAGE in the same cycle. The disadvantage is that this required global synchronization of all domains to move from `Idle` to `Marking` phases.

kayceesrk:
As it happens, there isn't a long `Idle` phase in the vanilla OCaml GC. The `Idle` phase is only used to capture the state between the end of a minor GC cycle and before the roots are marked for the new cycle. The distinction between `Idle` and `Marking` is necessary since objects must be allocated as `White`/`Unmarked` in the `Idle` phase. 

In the multicore GC, we scan the roots immediately after the end of the major GC cycle, before the mutators are run. The major slices are not synchronized in the multicore GC. If the different domains were allowed to independently move from `Idle` to `Marking` phase, it is possible that a domain in `Idle` phase might witness the objects marked by a different domain. The invariants become very subtle at this point. Hence, at the end of major cycle, when all the domain synchronize on a barrier, It is simpler to scan the roots on all the domains before the mutator is allowed to run. Importantly, it is not necessary to synchronize the domains at the end of marking the roots - no additional global synchronizations are required.

PR #147, opened 2017-08-02
Switching to Yuasa (deletion) barrier from Dijkstra (insertion) barrier

kayceesrk:


kayceesrk:
The PR relates to #67,  #41 and #20. This PR covers #41 and #67. #20 will not be implemented.

PR #148, opened 2017-08-03
New Domain API with better low-level synchronisation primitives.

stedolan:


stdlib/domain.ml
```diff
@@ -1,11 +1,90 @@
...
+| Finished of ('a, exn) result
+| Joined
+
+type 'a t =
+  { domain : Raw.t; state : 'a state Atomic.t }
+
+exception Retry
+let rec spin f =
+  try f () with Retry ->
+     (* fixme: spin more gently *)
```
stedolan: Yes. The next feature I want to add is a couple of primitives for sleeping a defined amount of time or until an interrupt arrives.

stdlib/domain.mli
```diff
@@ -1,17 +1,83 @@
...
+  (** [critical_section f] runs [f], but blocks notifications until
+      [f] returns. See [notify] below. *)
+
+  val notify : id -> unit
+  (** If the domain [d] is within a critical section (i.e. is evaluating
+      [critical_section f]), then [notify d] waits for this critical
+      section to complete before returning.
+      Otherwise, [notify d] does nothing. In general, when [notify d]
+      returns, all of [d]'s critical sections which began before
+      [notify] was called have ended. *)
```
stedolan: No, the second wait in domain a does nothing, and domain b does not return until the critical section has ended. `notify` affects critical sections, not calls to `wait` - once a critical section is interrupted, it stays that way, and `wait` becomes a no-op.

stdlib/domain.ml
```diff
@@ -1,11 +1,90 @@
...
+  external self : unit -> t
+    = "caml_ml_domain_id"
+end
+
+module Sync = struct
+  let critical_section f =
+    Raw.critical_adjust (+1);
+    match f () with
+    | x -> Raw.critical_adjust (-1); x
+    | exception ex -> Raw.critical_adjust (-1); raise ex
```
stedolan: Hmm. I'm relying on the idea that effects that act like exceptions are generally followed by actual exceptions (or will be, once FiberDeath is implemented). It's probably a bad idea to perform effects inside critical sections anyway, though.

kayceesrk: Ok. That makes sense. Effect handlers that return immediately should be fine. If you drop the continuation, then the system will make progress after GC. The only tricky case would be when the continuation is stashed away in data structure for later resumption. At this point, any notifications will block until the continuation resumes. There is a risk of deadlock here.

stdlib/domain.mli
kayceesrk: This is quite subtle and differs from conventional wisdom. Would be useful to include the example in the comment and add a test case such that we don't silently break this guarantee in the future.

stdlib/domain.mli
```diff
@@ -1,17 +1,83 @@
...
+            if success then () else await_completion ()
+
+          let signal_completion waiting_domain =
+            Atomic.set success true;
+            notify waiting_domain
+
+      Semantically, the primitives are similar to having a single
+      monitor (or mutex + condition variable) per domain. That is,
+      [critical_section f] acquires the current domain's mutex and
+      runs [f], [wait f] waits on the current domain's condition
```
stedolan: fixed

stdlib/domain.mli
```diff
@@ -1,17 +1,83 @@
...
+      variable (releasing the mutex during the wait), and [notify d]
+      acquires domain [d]'s mutex and signals its condition variable.
+      The only difference from standard monitors is that [notify d]
+      waits for the critical sections that were in progress to complete.
+
+      However, the actual implementation is somewhat different. In
+      particular, [critical_section f] is cheaper than acquiring a mutex,
+      and performs no atomic operations *)
```
stedolan: fixed

stdlib/domain.mli
```diff
@@ -1,17 +1,83 @@
...
+
+  val critical_section : (unit -> 'a) -> 'a
+  (** [critical_section f] runs [f], but blocks notifications until
+      [f] returns. See [notify] below. *)
+
+  val notify : id -> unit
+  (** If the domain [d] is within a critical section (i.e. is evaluating
+      [critical_section f]), then [notify d] waits for this critical
+      section to complete before returning.
```
stedolan: I think the new version is clearer

stedolan:
OK, tests added, so I think this pullrq is done. (I'll add the timing / spinning stuff as a separate pull request).

PR #149, opened 2017-08-03
Fix a subtle bug in domain termination.

stedolan:
The old domain termination logic was roughly:
  - turn off incoming interrupts
  - empty minor heap / mark stack
  - quit

This had a subtle race condition: after incoming interrupts are
turned off, this domain no longer participates in STW heap cycling.
So, should a major GC cycle end just after we turn off incoming
interrupts, the GC colours can flip as we're emptying the mark
stack, causing segfaults / corruption.

The fix is to have domain termination logic like this:
   loop {
     - empty minor heap / mark stack
     - handle incoming interrupts
   } until no more incoming interrupts
   - turn off incoming interrupts
   - quit

Previously, the `interrupt.ml` test failed in debug builds about 50% of the time (in optimised builds much less frequently, because they do fewer checks). This version has passed ~130 test runs in a debug build with no failures.

(Also, we should run the testsuite in debug mode to catch this sort of thing better)

### DIFF
--- a/byterun/domain.c
+++ b/byterun/domain.c
@@ -52,7 +52,6 @@ int caml_send_interrupt(struct interruptor* self,
 void caml_handle_incoming_interrupts(struct interruptor* self);
 void caml_yield_until_interrupted(struct interruptor* self);
 void caml_start_interruptor(struct interruptor* self);
-void caml_stop_interruptor(struct interruptor* self);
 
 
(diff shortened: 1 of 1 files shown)


stedolan:
The test failure is in join.ml, since #148 isn't merged yet. `interrupts.ml` passed

PR #151, opened 2017-08-08
amd64.S simplification

stedolan:
- move the stack check on callbacks out of assembly into callback.c
  - fixed a bug in amd64.S - old Stack_parent not being registered as a root (test callback/nested_fibers now segfaults in native).
  - moved some saving / restoring of context into callback.c instead of amd64.S


asmrun/amd64.S
```diff
@@ -973,9 +946,18 @@ FUNCTION(G(caml_raise_exception))
 CFI_STARTPROC
         movq    C_ARG_1, %r15                /* young_ptr */
         movq    C_ARG_2, %rax
+LBL(caml_raise_exception):
```
kayceesrk: This looks like a very sensible change 👍

PR #152, opened 2017-08-14
Timing primitives

stedolan:


byterun/caml/platform.h
```diff
@@ -169,6 +169,8 @@ typedef struct { pthread_cond_t cond; caml_plat_mutex* mutex; } caml_plat_cond;
 #define CAML_PLAT_COND_INITIALIZER(m) { PTHREAD_COND_INITIALIZER, m }
 void caml_plat_cond_init(caml_plat_cond*, caml_plat_mutex*);
 void caml_plat_wait(caml_plat_cond*);
+/* returns 1 if the timer expires */
+int caml_plat_timedwait(caml_plat_cond*, int64);
```
kayceesrk: what is this int64 argument? interval or a future timestamp?

byterun/domain.c
```diff
@@ -921,3 +919,39 @@ CAMLprim value caml_ml_domain_interrupt(value domain)
...
 }
+
+CAMLprim value caml_ml_domain_ticks(value unused)
+{
+  return caml_copy_int64(caml_time_counter() - startup_timestamp);
+}
+
+CAMLprim value caml_ml_domain_yield_until(value t)
+{
+  int64 ts = Int64_val(t) + startup_timestamp;
```
kayceesrk: `Assert (Int64_val(t) >= 0)`?

stdlib/domain.ml
```diff
@@ -5,22 +5,39 @@ module Raw = struct
     = "caml_ml_domain_critical_section"
   external interrupt : t -> unit
     = "caml_ml_domain_interrupt"
-  external wait : unit -> unit = "caml_ml_domain_yield"
+  external wait : unit -> unit
+    = "caml_ml_domain_yield"
+  type timeout_or_notified = Timeout | Notified
+  external wait_until : int64 -> timeout_or_notified
```
kayceesrk: Raise exception on negative values of int64?

kayceesrk:
lgtm otherwise.

dbuenzli:
Is there a strong need for that ? This increases the porting surface and can be a bit difficult to bootstrap on baremetal.

stedolan:
> Is there a strong need for that ? This increases the porting surface and can be a bit difficult to bootstrap on baremetal.

Yeah, we need timing primitives for several reasons (implementing `Thread.delay`, exponential backoff in synchronisation algorithms, benchmarking...).

The simplest approach to bare-metal porting is to use busy-waiting by spinning on the cycle counter register, although it would be worth using real system idling primitives to avoid pegging all cores to 100% all the time.

dbuenzli:
> Yeah, we need timing primitives for several reasons (implementing Thread.delay, exponential backoff in synchronisation algorithms, benchmarking...)

But does it really need to be in the core system ?  Can't this be implemented as libraries via asynchronous effects ?
 
> The simplest approach to bare-metal porting is to use busy-waiting by spinning on the cycle counter register, although it would be worth using real system idling primitives to avoid pegging all cores to 100% all the time.

The problem is that you need to know about mmio registers of the hardware clocks before doing this. But for discovering this you would already like to run the runtime system if you want to program as much as possible in OCaml. You can maybe devise a scheme were you start by busy waiting and when you get your hand on a clock you patch the fun via a global ref but it's quite ugly.

stedolan:
> But does it really need to be in the core system ? Can't this be implemented as libraries via effects ?

Yes, it could, but it means that concurrent algorithms using timing (e.g. exponential backoff mutexes) will have to manage an extra dependency. I think timing has about the same status as Random - it could well be in a library, but it causes less pain to have it always available.

If it does turn out to be seriously awkward to set up clocks before starting the runtime system, then we should revisit this (The API for low-level synchronisation stuff needs at least one reorganisation before multicore gets widely used, so I don't think changing it will be hard)

dbuenzli:
> Yes, it could, but it means that concurrent algorithms using timing (e.g. exponential backoff mutexes) will have to manage an extra dependency. I think timing has about the same status as Random - it could well be in a library, but it causes less pain to have it always available.

I disagree, being able to control timing primitives is extremely useful for testing. In my opinion you are adding things here that belong to the `Unix` library.


> If it does turn out to be seriously awkward to set up clocks before starting the runtime system,

Well I can already tell you that this will be the case with this PR.

And by the way do you seriously think that falling back on `gettimeofday` is a good idea ? I think I prefer no time than a time that pretends to be monotonic but which is not.  Besides AFAIK `POSIX_TIMER` is not available on macOS you should use `mach_absolute_time`.

stedolan:
> being able to control timing primitives is extremely useful for testing.

I agree, but again that's just the same for Random.

> In my opinion you are adding things here that belong to the Unix library.

Good point, maybe Unix is a better place for these things. In general, I'm unsure where in the otherlibs / stdlib split the new parallelism primitives should end up.

> And by the way do you seriously think that falling back on gettimeofday is a good idea ? I think I prefer no time than a time that pretends to be monotonic but which is not. Besides AFAIK POSIX_TIMER is not available on macOS you should use mach_absolute_time.

These timer values are used in `wait_until`, which passes them to `pthread_cond_timedwait`. OS X does not support `pthread_condattr_setclock`, so the timeout values are [based on what gettimeofday returns](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/pthread_cond_timedwait.3.html). OS X does seem to have some other clock sources available (as well as mach_absolute_time, there's `host_get_clock_service(..., SYSTEM_CLOCK, ...)` and others), but I'm not sure how to use them to do an interruptible sleep.

dbuenzli:
> I agree, but again that's just the same for Random.

I never said I agreed with you on entropy ;-)

> OS X does not support pthread_condattr_setclock, so the timeout values are based on what gettimeofday returns.

Fair enough but you also expose that value via `timer_ticks` which somehow suggests you have a monotonic clock there. But this should really not used as such if you have `gettimeofday` behind it.

> In general, I'm unsure where in the otherlibs / stdlib split the new parallelism primitives should end up.

As far as the runtime system is needed I would strive for the minimal needed, this won't only help for porting to exotic environment but also help for general C & other language runtimes interaction.

> but I'm not sure how to use them to do an interruptible sleep.

I guess you meant *timed* interruptible sleep. So maybe since platforms already make it difficult to use these abstractions in a cross platform it's better to simply have a (non timed) interruptible wait and layer timed ones on top of it...

PR #154, opened 2017-08-20
Work stealing for GC marking

kayceesrk:


kayceesrk:
`interrupt.ml` fails in CI, but not locally. @stedolan any ideas?

kayceesrk:
Finally turned green!

stedolan:
Nice!

I was thinking about an alternative means of detecting termination based on a single global counter of domains doing marking work (like you suggested last week), and I now think that the races with domain creation, write barrier and stealing can be avoided. Here's an attempt, hole-poking appreciated:

A global counter `N` represents the number of domains that have marking work to do. `N` may both increase and decrease, but:
  - once `N` reaches zero it does not change again until the GC cycle completes.
  - if `N` is zero, then there is no more marking work to do this cycle.

To maintain these invariants, we need to ensure that if the last bit of work is transferred from domain A to domain B, then the `N++` representing domain B's receipt of work happens before the `N--` representing `A`'s loss of work.

`caml_darken` (called on the old object during the write barrier) should do this:
```
if header.gc == UNMARKED:
  if mark stack empty:
    N++
  header = MARKED
```
By ensuring that `N++` is done before the header becomes `MARKED`, we avoid the race with the write barrier moving work from one domain to another - if another domain thinks it's finished marking by noticing the newly MARKED header, it will also notice the `N++`.

Before a domain starts work-stealing, it should do `N--`. If the steal is successful, `N++` should happen in the steal handler, to ensure that it happens before the victim domain continues marking and possibly runs out of work.

When a new domain is created, its mark stack will be empty, so it will only do `N++` if the mark stack becomes nonempty via the write barrier or stealing.

kayceesrk:
Refactored the termination detection based on above.

bluddy:
Sorry to butt in to this discussion -- I'm not up to date on the intricacies of the multicore GC at this point. I just want to point out that a single atomic accessed by every domain is going to be somewhat expensive to synchronize (proportional to number of domains), and making it part of every write barrier call will multiply this expense. Don't know if you guys considered that.

byterun/caml/mlvalues.h
```diff
@@ -25,6 +25,12 @@ typedef intnat value;
 typedef int32 opcode_t;
 typedef opcode_t * code_t;
 
+/* atomic_uintnat is declared as a struct type to prevent accidental
+   accesses that don't use the primitives of platform.h */
+typedef struct atomic_uintnat {
+  volatile uintnat val;
+} atomic_uintnat;
```
bluddy: I know it's just a move in this commit, but you may want to consider padding this with an architecture-specific cache line size minus sizeof(uintnat), and to [align to 64 bytes](https://stackoverflow.com/questions/7281699/aligning-to-cache-line-and-knowing-the-cache-line-size). Don't want atomics sharing cache lines, especially with other atomics.

kayceesrk:
> a single atomic accessed by every domain is going to be somewhat expensive to synchronize (proportional to number of domains), and making it part of every write barrier call will multiply this expense

The atomic variable is accessed only if we are not in the middle of marking and the mark stack is empty. In particular, the atomic access is protected by a conditional check on a non-atomic variable: https://github.com/ocamllabs/ocaml-multicore/pull/154/files#diff-7f93bb02eaacc4afd8a538643d1a2426R451. The conditional check is expected to fail in the common case.

byterun/caml/mlvalues.h
kayceesrk: @stedolan thoughts on this?

testsuite/tests/parallel/wait.ml
```diff
@@ -1,5 +1,5 @@
 let rec await r =
-  if Atomic.get r then () else await r
+  if Atomic.get r then () else (Atomic.make (); await r)
```
stedolan: `Domain.Sync.cpu_relax ()` (just merged, now available on master)

byterun/caml/mlvalues.h
stedolan: Putting padding around at least some atomics is a good idea (just created #155). I'd like to see some benchmark showing that it makes a difference, though.

I'm not sure what the point of aligning to 64 bytes is, though. I think you just need to ensure that atomics don't share cache lines accidentally. The default alignment of `uintnat` (either 4 or 8 bytes) ensures that they don't straddle cache line boundaries.

stedolan:
This looks good to me!

byterun/caml/mlvalues.h
bluddy: See [here at the bottom of the page](https://software.intel.com/en-us/articles/coding-for-performance-data-alignment-and-structures) and [here](https://stackoverflow.com/questions/24102832/atomic-increment-is-flushing-other-cache-lines-code-and-results-provided) for some possible implications.

Alignment makes sure the atomic starts on a cache line -- otherwise the padding just extends from one cache line to the next, and doesn't prevent false sharing before the atomic. You don't want false sharing between atomics, and you don't really want atomics to share cache lines with other data either, since every update of an atomic forces cache line dumping by all other cores, which will hurt your cache performance for non-atomic data.

Issue #129, opened 2017-05-16
Error handling during domain creation


stedolan:
If a new domain can't allocate enough memory to start, then `Domain.spawn` should raise an exception. Currently, it kills the whole runtime.

rmdouglas:
I've looked further into implementing a solution to this issue, and to me it looks like there's 2 real options:
 - Create a "dummy" exception handler on the stack (setsigjmp for bytecode, create a c_stack and exn_handler for native code).
- Make sure that the operations called from create_domain cannot raise an ocaml exception (by, for example, calling caml_stat_alloc_noexn instead of caml_alloc_stat).

To me the first is more complicated (bytecode is simple enough). Apart from a call to caml_create_root (which calls caml_alloc_shr, which has no '_noexn' counterpart) the other functions call caml_stat_alloc and should be amenable to switching to caml_stat_alloc_noexn and indicating an error (via returning NULL, or an error value, etc).

Does anyone have any further thoughts or opinions on the options I can see, or has better ideas for a solution?

rmdouglas:
Another option could be to somehow install the parent domain's exception handler(s) in the child domain while it is initializing, though I don't know if that would be simpler than installing "dummy" exception handlers, nor how it might cause issues due to the exception crossing domain boundaries.

kayceesrk:
Fixed by #264.

PR #163, opened 2017-11-24
[WIP] Raise exception if alloc fails

Abhiroop:
Trying to fix https://github.com/ocamllabs/ocaml-multicore/issues/129 
Questions:
1. Is it possible to add tests to test the raising of exception? Can someone guide me on this?
2. Does raising an exception suffice or does it require any other graceful handling?

### DIFF
--- a/byterun/domain.c
+++ b/byterun/domain.c
@@ -198,6 +198,7 @@ static void create_domain(uintnat initial_minor_heap_size) {
     domain_state->young_start = domain_state->young_end =
       domain_state->young_ptr = 0;
     domain_state->remembered_set = caml_alloc_remembered_set();
+    if (domain_state->remembered_set == NULL) caml_raise_out_of_memory;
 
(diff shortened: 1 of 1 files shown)


stedolan:
Apologies this took so long to get to!

This approach isn't the right one - `caml_alloc_remembered_set` can't return `NULL`, because it calls `caml_stat_alloc` which raises an exception if the allocation fails. Normal memory allocation functions in the OCaml runtime never return null, so these checks aren't generally necessary.

The issue in #129 is that exceptions that occur on the new domain's thread are treated as exceptions at the top-level of a domain (which they are), and terminate the system. Arguably exceptions that occur before Domain.spawn has returned should actually be routed back to the spawning domain. (This might possibly be the case already, haven't look at that code in a while).

ctk21:
Closing this PR as #264 which fixes the issue has been merged.

PR #168, opened 2018-04-03
Install .tbl files.

yallop:
The `.tbl` files are [included by installed headers](https://github.com/ocamllabs/ocaml-multicore/blob/9749637d6642b48f330483a73c41c336713fbc47/byterun/caml/domain_state.h#L17), so they need to be installed, too.


stedolan:
Good catch, thanks.

The testsuite runs out of the build directory, so it doesn't catch this. We really need some sanity testing for the actual install.

stedolan:
The build failure was unrelated - there's not enough tolerance in the timing-sensitive `parallel/timing.ml` test.

stedolan:
Fix for the timing issue in 48e8ae8

PR #170, opened 2018-04-06
Merge 4.06.1

kayceesrk:


kayceesrk:
`tests/parallel/timing.ml` fails on OSX and is known to fail. Merging. Multicore catches up to trunk!

PR #173, opened 2018-04-12
Windows port

dra27:
There isn't an issue for Windows support at the moment, so here's a pull request to track it in instead. It's obviously not yet intended for merging.

There's a squashed commit of changes from #172 and this backports https://github.com/ocaml/ocaml/pull/1721.

All comments and contributions welcome!

### DIFF
--- a/byterun/win32.c
+++ b/byterun/win32.c
@@ -857,7 +857,7 @@ CAMLexport inline wchar_t* caml_stat_strdup_to_utf16(const char *s)
   int retcode;
 
   retcode = win_multi_byte_to_wide_char(s, -1, NULL, 0);
-  ws = malloc(retcode * sizeof(*ws));
+  ws = caml_stat_alloc_noexc(retcode * sizeof(*ws));
   win_multi_byte_to_wide_char(s, -1, ws, retcode);
 
   return ws;
--- a/byterun/win32.c
+++ b/byterun/win32.c
@@ -41,6 +41,8 @@
 #include "caml/osdeps.h"
 #include "caml/signals.h"
 #include "caml/sys.h"
+#include "caml/startup_aux.h"
+#include "caml/platform.h"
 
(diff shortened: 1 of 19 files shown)


byterun/win32.c
```diff
@@ -914,3 +916,22 @@ void caml_restore_win32_terminal(void)
...
+  QueryPerformanceFrequency(&frequency);
+}
+
+int64_t caml_time_counter(void)
+{
+  LARGE_INTEGER now;
+  /* Windows 2000 is no longer supported, so this function always succeeds */
+  QueryPerformanceCounter(&now);
+  now.QuadPart *= 1000000000L;
+  return (now.QuadPart * 1000000000L) / frequency.QuadPart;
```
stedolan: This doesn't look right. First, it does the sec->nanosecs conversion twice. Second, even if only done once, I think the multiplication will overflow after less than a day.

kayceesrk:
PR seems to have broken native compilation. See travis logs: https://travis-ci.org/ocamllabs/ocaml-multicore/jobs/365689997

> Minor_heap_align_bits misconfigured for this platform

byterun/win32.c
dra27: Yeah, I seem to have turned my brain off there.

byterun/win32.c
```diff
@@ -925,13 +925,14 @@ void caml_init_os_params(void)
   GetSystemInfo(&si);
   caml_sys_pagesize = si.dwPageSize;
   QueryPerformanceFrequency(&frequency);
+  /* Convert the frequency to the duration of 1 tick in ns */
+  frequency.QuadPart = 1000000000L / frequency.QuadPart;
```
dra27, line 929: @stedolan 128bit arithmetic will be bad - is converting the tick frequency to ns acceptable?

PR #178, opened 2018-04-30
Install .tbl files

### LINKED ISSUE #177: Can't 'opam install' owl
[See errors here](https://gist.github.com/ohad/4e8e0b2464f419cdc2cf8394adb06290).

Happy to provide more information!

yallop:
Follow-up to #168.  This should fix #177.


stedolan:
Thanks! (For anyone following along: the fix in #168 was to `Makefile.common`, which should have been deleted in favour of `Makefile`, but hung around doing nothing because of a merge issue)

PR #180, opened 2018-05-08
Simplify C exception handling by removing system_exnptr_offset.

stedolan:
It was only used to adjust the stack in caml_raise_exception,
but that function immediately calls SWITCH_C_TO_OCAML,
which resets the stack pointer.


stedolan:
This breaks arm64. I'm going to do a few more rounds of x86-specific hacking, then fix arm.

PR #198, opened 2018-07-10
Remove dependency on shared heap from frametables.

stedolan:
Currently, frametables are allocated on the shared GC heap, to ensure that they get freed. This leads to a couple of "how did that ever work" circularities during domain initialisation.

With this patch, frametables are now statically allocated. Deallocation of old frametables is done by a check during stack scanning, and no frametables are deallocated until after the end of the GC cycle in which they were replaced. (This is effectively RCU).


asmrun/frame_descriptors.c
```diff
@@ -81,48 +98,70 @@ static link *cons(intnat *frametable, link *tl) {
...
 
   caml_plat_unlock(&descr_mutex);
 }
 
-frame_descr* caml_find_frame_descr(uintnat pc)
+caml_frame_descrs caml_get_frame_descrs()
+{
+  struct frametable_version *ft =
+    (struct frametable_version*)atomic_load_acq(&current_frametable);
+  Assert(ft);
```
kayceesrk: `CAMLassert` instead of `Assert`?

asmrun/frame_descriptors.c
```diff
@@ -1,8 +1,12 @@
 #define CAML_INTERNALS
 
 #include "frame_descriptors.h"
-#include "caml/stack.h"
 #include "caml/platform.h"
+#include "caml/major_gc.h" /* for caml_major_cycles_completed */
+#include <stddef.h>
```
kayceesrk: Why do you need this?

stedolan: for `NULL`

asmrun/frame_descriptors.c
stedolan: Fixed

asmrun/frame_descriptors.c
kayceesrk: cool.

stedolan:
The Travis failure is real, but it's a rare nondeterministic bug and seems to have nothing to do with this commit (I can reproduce it on master). Debugging atm, I'll make a new PR when I figure it out.

PR #199, opened 2018-07-10
Run tests under debug runtime

stedolan:


stedolan:
This was failing because `caml_alloc` initialises fields to Debug_uninit_minor in debug mode, instead of Val_unit. That's changed now.

PR #200, opened 2018-07-10
Fix a race condition in parallel/join.ml.

stedolan:
The test used multiple `incr` from different threads with no sync.
Instead, use an array of separate flags.

Also, squash a warning by using an exception other than Failure
when matching on a particular string.

### DIFF
--- a/byterun/domain.c
+++ b/byterun/domain.c
@@ -827,9 +827,7 @@ static void domain_terminate() {
     Caml_state->critical_section_nesting = 0;
     acknowledge_all_pending_interrupts();
   }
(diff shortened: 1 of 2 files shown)


stedolan:
Debugging the test failure in #198 led to two separate bugs: a data race in the test itself (fixed by this patch) and a more serious segfault that I'm still debugging.

stedolan:
Segfault now fixed! It was a bug in domain termination, occurring nondeterministically about once per 3k executions (on a loaded laptop doing builds in another tab). With the fix, it's passed 30k executions.

PR #201, opened 2018-07-10
Add CAMLnoalloc to assert that the GC must not be invoked in a block.

stedolan:


stedolan:
This now depends on #199

PR #202, opened 2018-07-23
Incremental marking

stedolan:
Changes the major GC mark stack to contain object intervals instead of whole objects.

Incremental marking makes `Array.init (10_000_000) (fun _ -> ref 42)` much faster.

Stacks are now scanned as soon as they are darkened.


byterun/major_gc.c
```diff
@@ -41,6 +41,17 @@
 */
 #define MARK_STACK_SIZE (1 << 14)
 
+typedef struct {
+  value block;
+  uintnat offset;
+  uintnat end;
```
kayceesrk: Why do you want `end` field? Isn't this always `Wosize_val(block)`. You will save a word per entry if you remove it.

stedolan: There's a bunch of optimisations to be done here, that I'm leaving for a later patch. I do want to make mark_entry smaller, but I'd prefer to remove `block` and keep `offset` and `end` (as pointers). That way, we could implement steal-half-the-frontier workstealing easily.

kayceesrk: ok. Removing block seems like an immediate win. lgtm otherwise.

stedolan: `block` is there at the moment solely because some code in mark stack overflow handling needs it. I'd like to refactor that to not use block (with the bitmap trick we discussed), but later.

kayceesrk: This PR lgtm to merge once CI passes.

stedolan:
Travis failed, but only in the debug testsuite which is known to fail on master. Trying to work out if it's the same issue or a new bug.

bluddy:
Cool. This can probably improve the single-core GC upstream as well, right?

stedolan:
Seems to be the same debug testsuite bug as master, so merging this. @bluddy Possibly. Upstream already has (somewhat different) incremental marking.

Issue #175, opened 2018-04-25
Can't extract continuation out of domain


zigaLuksic:
The following code should extract a continuation from a domain and re-run it.

```
effect Poke : unit

type result = Done | Poking of (unit -> result)

(* Debug help. *)
let print s = print_string s; Format.pp_print_flush Format.std_formatter ()

(* Just poke the handler n times. *)
let rec poke = function
  | 0 -> ()
  | n -> perform Poke; poke (n-1)

(* The handler inside the domain, that captures the continuation whenever
    it gets poked. *)
let domain_handler f =
  match f () with
  | effect Poke k -> Poking (fun () -> print "..."; continue k (); print "success"; Done)
  | () -> Done

(* Re-runs the poker that happened inside a domain. *)
let rerunner = function
  | Poking f ->  f () (*re-runs the function*)
  | Done -> Done

(* Test. *)
let test n =
  (* Messy handler wrapping. *)
  let inner () = domain_handler (fun () -> poke n) in
  rerunner (Domain.join (Domain.spawn inner))

let _ = test 100
```

PR #203, opened 2018-07-23
Continuation refactor

stedolan:
This patch makes continuations, rather than stacks, the primitive supported by the runtime. Stacks are now `malloc`ed rather than kept on the heap, and don't move except when being extended. The GC now treats continuations as immutable objects (continuations in the major heap may not point to minor heaps), instead of treating stacks as mutable objects (requiring a remembered set).

This greatly simplifies the implementation of effects. `Stack_tag`, `fiber_ref`, `Stack_dirty_domain`, `caml_scan_dirty_stack_domain`, `FIBER_SCANNING` and `should_promote_stacks` all go away.

Tasks:

  - [x] new primitives for continuations
  - [x] explicitly free stacks
  - [x] minor GC and promotion
  - [x] major GC
  - [x] native code support (amd64)
  - [x] continuation cloning and backtraces
  - [x] non-reversing continuation structure during delegate
  - [x] remove now-redundant BVars
  - [x] testcase for #175 (fixed by this patch)
  - ~~ARM64 support~~
  *Turns out ARM64 support is broken on `master` at the moment (probably because we don't have a CI 
 to shout at us when we break it), so I'll leave this out until it's fixed. See #120*


bluddy:
Wow. Do you have a link to a discussion of the trade-offs?

lpw25:
Isn't this going to make effects much slower? Compared to minor heap allocation, malloc is pretty slow.

stedolan:
@bluddy no, afraid not. 

@lpw25 possibly. If so, it would be worth implementing a per-domain pool of spare stacks. Even with a pool, it's still much simpler than what's there at the moment. Incidentally, the malloc version is not necessarily slower: since stacks are now freed explicitly when the fiber terminates, programs that make heavy use of effects will tend to reuse the same memory over and over for stacks, instead of causing GC pressure with short-lived stacks.

lpw25:
What about reusing the allocator for the major heap to do a per-domain pool of size-specific free lists? Obviously you'd only include the possible sizes of stacks in the pool. The allocation pattern of stacks is also very specific so I think it would never fragment.

stedolan:
The major heap allocator doesn't have a explicit free operation (sweeping happens in batch), so it wouldn't quite fit. I'll play around with this later, but at the moment I'm thinking of just a linked list of free stacks of the smallest size (if a fiber does enough work to grow the stack, then we can probably amortize a `malloc`).

stedolan:
@lpw25 here's a simple benchmark that does lots of effects:

```
effect E : unit

let () =
  for i = 0 to 1000000 do
    match perform E with
    | () -> ()
    | effect E k -> continue k ()
  done
```

On my machine, master and this branch take similar amounts of time (~210 ms for this branch, ~260ms for master), but this branch does much less GC (7 minor collections vs. 137). A custom allocator for stacks would probably improve things, but it's good enough for now.

lpw25:
I'm not sure that benchmark is ideal -- I'd want to try with something that was also doing a lot of other allocations alongside the stack ones -- but it seems good enough to make the change. As you say, a better allocator can be added later.

stedolan:
ARM support turns out to have bitrotted a bit, so I'm leaving it out of this patch until it's fixed.

This should be ready to merge! Waiting for CI to finish.

PR #205, opened 2018-07-27
Fixes issue 204

dhil:
.. by making the continuation variable inaccessible in `when` expressions. The fix is rather simple, we keep two typing environments live `ext_env` and `ext_env'`, where the latter is an extension of the former with the continuation variable bound. We use `ext_env` when typing the `when` expression, and `ext_env'` when typing the clause body.


stedolan:
Nice, thanks!

(the build failure is a known issue on master)

PR #209, opened 2018-08-07
Stack pointer refactor

stedolan:
Now that stacks no longer move during GC, it's safe to store stack pointers directly rather than as offsets. This simplifies the computations necessary to restore stack pointers in amd64.S.

This makes `caml_extern_sp` redundant, so it's removed. That makes `caml_save/restore_stack_gc` no-ops, so they're gone too.

### DIFF
--- a/asmrun/amd64.S
+++ b/asmrun/amd64.S
@@ -168,19 +168,15 @@
 /* Stack switching operations */
 /******************************************************************************/
 
-/* Switch from OCaml to C stack. Clobbers REG & %r14. */
+/* Switch from OCaml to C stack. Clobbers %r14. */
 #define SWITCH_OCAML_TO_C_NO_CTXT(REG) \
     /* Save OCaml SP in the stack slot */ \
-        mov     %rsp, REG; \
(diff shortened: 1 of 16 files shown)


asmrun/amd64.S
```diff
@@ -168,19 +168,15 @@
...
-        mov     %rsp, REG; \
-        LOAD_TL_VAR(caml_stack_high, %r14); \
-        subq    %r14, REG; \
-        sarq    $3, REG; \
         LOAD_TL_VAR(caml_current_stack, %r14); \
-        movq    REG, (%r14); \
+        movq    %rsp, Stack_sp(%r14); \
     /* Switch to system stack */ \
         LOAD_TL_VAR(caml_system_sp, %r14); \
+        movq    %rsp, 16(%r14); /* For DWARF CFI. See OFFXXX. */ \
```
kayceesrk: Do these still work?

stedolan:
I don't think so. The DWARF stuff is broken at the moment anyway. I'll get around to fixing it after these refactorings are done.

PR #213, opened 2018-08-23
Improve performance by doing fewer read barriers.

stedolan:
- Read barriers were done on references to global constants
    in native code due to a missed case in translmod.ml
  - Read barriers were done on loads of int refs due to a
    missed optimisation in translcore.ml


kayceesrk:
This looks good to me. Merging. 

I've been very conservative in marking things as mutable. It would be worth going over every introduction form of `Pfield` to see if it is too strong.

PR #215, opened 2018-08-30
R14 globals

stedolan:
This patch uses %r14 on amd64 to access globals, simplifying the assembly stubs and shortening the code needed for making allocations and C calls.

For reviewing: the first commit makes a nontrivial change, moving the current exception handler from %r14 to a field in Caml_state, and changing how the offsets are computed. The other commits are mechanical.


kayceesrk:
IIUC We don't need `Caml_state->stack_high` anymore do we? This can be a separate PR.

PR #216, opened 2018-08-31
Remove stack_high and cleanup

stedolan:
stack_high is now redundant (after #215), so it's removed. That makes `load_stack`, `save_stack`, `caml_restore_stack` and `caml_init_main_stack` trivial, so they're inlined & removed.

### DIFF
--- a/byterun/fiber.c
+++ b/byterun/fiber.c
@@ -20,19 +20,8 @@
 #include "frame_descriptors.h"
 #endif
 
+
 #ifdef NATIVE_CODE
-
-static struct stack_info* save_stack ()
-{
-  return Caml_state->current_stack;
-}
-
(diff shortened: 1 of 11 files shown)


byterun/fiber.c
```diff
@@ -280,9 +254,11 @@ void caml_scan_stack(scanning_action f, void* fdata, struct stack_info* stack)
 
 struct stack_info* caml_switch_stack(struct stack_info* stk)
```
kayceesrk: I guess this can be inlined in `amd64.S` now. No need to hop through C for switching OCaml stacks.

stedolan: Yep, that sounds good. Still need to jump out to C for freeing the old stack, and native code has a DWARF-related hack in the return value that need to be replaced before this gets inlined.

PR #219, opened 2018-09-06
Fix a missed optimisation in root scanning.

stedolan:
Trunk OCaml scans newly-initialised globals during minor GC, as
they may contain newly-allocated minor objects. Multicore deleted
the tracking of which globals were newly initialised, as the
synchronisation of this variable was tricky.

So, multicore scans all globals during minor GC. This was a
temporary hack, with the better fix being to make initialisation
of globals promote before initialisation, so that newly-initialised
globals never contain minor heap pointers.

The better fix got implemented, but I forgot to remove the hack,
so minor GC has been uselessly scanning globals for quite a while.
This commit removes the unnecessary scan.

### DIFF
--- a/byterun/roots.c
+++ b/byterun/roots.c
@@ -60,21 +60,6 @@ CAMLexport void caml_do_local_roots (scanning_action f, void* fdata,
   int i, j;
   value* sp;
 
-#ifdef NATIVE_CODE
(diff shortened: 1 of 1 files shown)


kayceesrk:
makes sense to me. can merge once the ci turns green.

Issue #196, opened 2018-06-28
Missing atomic operations


bordoley:
This is mostly a hypothetical request, but in general it would be useful to provide additional atomic operations including:
```
(* return the previous value when setting a new value *)
val exchange : 'a t -> 'a -> 'a

(* increment an int atomically *)
val incr: int t

(* decrement an int atomically *)
val decr: int t
```

stedolan:
I agree, these are very useful.

bordoley:
Somewhat related, are you considering any type of API for performing atomic operations upon mutable record fields? C# enables this via the Interlocked class and allows passing a direct pointer to a class field via the `ref` keyword. This might a useful feature for performance sensitive code that seeks to avoid allocations by packing multiple atomic values into the same record.

stedolan:
Not at the moment, no. There's no analogue of C#'s `ref` in OCaml, and we're not currently planning to add one.

stedolan:
Closed by #225

PR #225, opened 2018-10-04
Use C11 atomics to enforce the OCaml memory model.

### LINKED ISSUE #186: Enforce the memory model
[Multicore OCaml's memory model](http://kcsrk.info/papers/pldi18-memory.pdf) is free on `amd64` but requires fence (`dmb ld`) before stores on arm64. This requires fence insertion in the following places:

- [ ] In aarch64/emit.mlp, before `IStore` instructions if it is an assignment (as opposed to initialising store).
- [ ] In the write barrier `caml_modify` before the store. 

We also need a full fence:

- [ ] At the end of a minor GC
- [ ] After promotion
- [ ] After initialising globals
- [ ] After `caml_alloc_shr`

Our current focus in `amd64` and `aarch64` backends only. 

Currently, we only focus on `amd64` and `arm64` backends.

stedolan:
- Add Hp_atomic_val and Op_atomic_val to get atomic block accesses
  - Hp_val(v) is no longer an lvalue
  - Remove all uses of old GCC __sync_foo primitives
  - Remove architecture-specific atomic code in platform.h
    (except cpu_relax(), which annoyingly isn't in C11)
  - Add `fetch_and_add`, `exchange`, `incr`, `decr` to Atomic. (#196)

See the long comment in memory.c to understand how the OCaml memory
model is encoded in C11.

This PR closes #186.


byterun/caml/mlvalues.h
```diff
@@ -134,9 +129,8 @@ bits  63        (64-P) (63-P)        10 9     8 7   0
 #define Wosize_hd(hd) ((mlsize_t) ((hd) >> 10))
 #endif /* WITH_PROFINFO */
 
-#define Hd_val(val) (((header_t *) (val)) [-1])        /* Also an l-value. */
-#define Hd_op(op) (Hd_val (op))                        /* Also an l-value. */
-#define Hd_bp(bp) (Hd_val (bp))                        /* Also an l-value. */
+#define Hd_val(val) (((header_t *) (val)) [-1] + 0)
```
kayceesrk: Nice trick to kill l-valueness 👍

stedolan: "Nice" might be a bit strong. It works :)

byterun/memory.c
```diff
@@ -133,34 +227,41 @@ CAMLexport int caml_atomic_cas_field (value obj, int field, value oldval, value
...
+    ret = atomic_exchange(Op_atomic_val(ref), v);
+    write_barrier(ref, 0, ret, v);
+  }
+  return ret;
 }
 
 CAMLprim value caml_atomic_cas (value ref, value oldv, value newv)
 {
-  value* p = Op_val(ref);
-  if (Is_young(ref) || caml_domain_alone()) {
```
kayceesrk: Is there a reason why `caml_domain_alone` optimisation was removed here?

byterun/memory.c
```diff
@@ -169,24 +270,30 @@ CAMLprim value caml_atomic_cas (value ref, value oldv, value newv)
...
+    } else {
+      return Val_int(0);
+    }
   }
 }
 
-int caml_atomic_cas_raw (value* p, value oldv, value newv)
+CAMLprim value caml_atomic_fetch_add (value ref, value incr)
 {
-  if (caml_domain_alone()) {
```
kayceesrk: Is there a reason why `caml_domain_alone` optimisation is removed?

byterun/minor_gc.c
```diff
@@ -453,7 +453,7 @@ CAMLexport value caml_promote(struct domain* domain, value root)
         value new_p = old_p;
         forward_pointer (st.promote_domain, new_p, &new_p);
         if (old_p != new_p)
-          caml_atomic_cas_raw (*r,old_p,new_p);
+          atomic_compare_exchange_strong((atomic_value*)*r, &old_p, new_p);
```
kayceesrk: Change it to `caml_atomic_cas` here to take advantage of `caml_domain_alone`?

byterun/minor_gc.c
```diff
@@ -619,8 +619,10 @@ void caml_empty_minor_heap_domain (struct domain* domain)
           CAMLassert(Tag_val(vnew) == Infix_tag);
           v += offset;
         }
-        if (caml_atomic_cas_raw(*r,v,vnew)) ++rewrite_successes;
-        else ++rewrite_failures;
+        if (atomic_compare_exchange_strong((atomic_value*)*r, &v, vnew))
```
kayceesrk: Change it to `caml_atomic_cas` here to take advantage of `caml_domain_alone` info?

stedolan:
I left out `caml_domain_alone` when trying to work out the mapping of OCaml -> C++ memory models, forgot to re-add it.

byterun/caml/mlvalues.h
yallop: How about something like the following, for making the de-lvaluing more explicit?

```C
#define RVALUE(x) ((void)0,x)
#define Hd_val(val) RVALUE(((header_t *) (val)) [-1])
```

stedolan: That does look like a general solution, but I'd wait until there's a second instance of this before adding another definition.

PR #226, opened 2018-10-16
Multicore-safe implementation of Lazy

kayceesrk:
This PR adds multicore safe lazy values. 

## Lazy in trunk

The current design of lazy values in trunk is documented here in [0] . An unforced lazy block holds the think, say `f`, in the first field. When a lazy value is forced [1], the thunk is first replaced with another closure that raises `Undefined` exception. This is used to detect recursive forcing of lazy. After this `f` is evaluated to either result in a value or an exception. 

If the evaluation results in a value, then the first field is now updated to the value and the tag of the lazy object is updated to `Obj.forward_tag` using `Obj.set_tag`. If the evaluation results in an exception, then the first field of the lazy block is updated to a closure that raises the same exception but no change is made to the tag. Hence, subsequent forcing of the lazy will result in the same exception being thrown.

## Requirements for multicore

The aim here is to
1. Make lazy operations safe with concurrent GC
2. Make concurrent operations on the same lazy safe
3. Try to minimise the space and time overhead for muticore-safe lazy compared to the current design. 

For 1. `set_tag` is incompatible with the current marking scheme which does not use synchronisation on the header (See https://github.com/ocaml/ocaml/pull/1725). For 2. trunk uses no synchronisation on updating the header and the fields of the lazy block, which may lead to multiple domains evaluating the lazy thunk concurrently. This is undesireable since the thunk evaluation may have side effects. 

## Design

We achieve 2. by changing the closure that is stored in the lazy block. For example, `Lazy.from_fun f` used to create a lazy block `l` whose first argument is `f`. Now, the first field of the lazy block `l` stores `wf` where:

```ocaml
let wrap_fun (f: unit -> 'a) l =
  let myid = domain_self () in
  let bomb () =
    if myid = domain_self () then
      raise Undefined
    else raise RacyLazy
  in
  let rec wf () =
    if Obj.compare_and_swap_field (Obj.repr l) 0 (Obj.repr wf) (Obj.repr bomb) then
      f ()
    else raise RacyLazy
  in
  wf

let wf = wrap_fun f l
```

For the implementation, see [2]. The idea here is that, when the closure in the lazy block is forced, we use a CAS to update the first field from `wf` to `bomb`. `bomb` distinguishes between recursive forcing (raises `Undefined`) and concurrent forcing (raises `RacyLazy`). Note that the implementation assumes that the fiber that is forcing lazy does not migrate between domains during the forcing. In the case of recursive forcing, we will spuriously raise `RacyLazy` instead of `Undefined`. 

For 2. We use `cas_tag` to atomically update the tag from `lazy_tag` to `forward_tag` in the case of successful evaluation. Correspondingly, the marking code in `major_gc.c` is updated such that in the case of marking a lazy object, we use a CAS to update the mark bits. The rest of the implementation remains the same. 

## Overheads

The additional overheads over trunk now are (1) allocation of a new thunk (`wf` above) for lazy blocks and (2) CAS operations on header and field of lazy blocks. The benchmarks indicate that the overheads aren't observable: http://ocamllabs.io/multicore/. It is possible that the benchmarks aren't lazy heavy, in which case, the benchmark suite needs to be updated to include such benchmarks (candidates?). As an optimisation for the common sequential case, CAS is implemented as a non-atomic operation (See the use of `caml_domain_alone()` in `cas_tag` and `caml_atomic_cas_field`). 

## Future work: handling racy access gracefully

On racy access to a lazy, we raise `Racylazy` exception. We can do better with effect handlers,. Glasgow Haskell Compiler uses blackholing mechanism [3] to handle instances where multiple threads may evaluate the same thunk. We can build such a mechanism by making `RacyLazy` an effect, which suspends the execution of the fiber that concurrently evaluates the lazy. The continuation of the corresponding fiber is enqueued in the blackhole queue to be resumed when the lazy evaluation completes. 

[0] https://github.com/ocamllabs/ocaml-multicore/blob/master/stdlib/lazy.ml#L25
[1] https://github.com/ocamllabs/ocaml-multicore/blob/master/stdlib/camlinternalLazy.ml#L23
[2] https://github.com/ocamllabs/ocaml-multicore/blob/multicore-lazy/stdlib/camlinternalLazy.ml#L23-L35
[3] https://ghc.haskell.org/trac/ghc/wiki/LightweightConcurrency#Black-holeHandling


stedolan:
(back from vacation now, catching up)

Looks good! I'm looking forward to playing with the effectful version of `RacyLazy`.

Two minor notes about the implementation:

  - I think that `Obj.cas_tag` can only safely be used when `old_tag = Lazy_tag`, so it might be better not to provide the parameter.

  - Why is an assignment used instead of an initialisation in translcore.ml?

kayceesrk:
> I think that Obj.cas_tag can only safely be used when old_tag = Lazy_tag, so it might be better not to provide the parameter.

Good point. Initially, I was going to introduce a new `Forcing` tag for which I would have needed CAS from `Lazy` to `Forcing` and `Forcing` to `Forwarded`. This is no longer needed. I'll specialise the `cas_tag` to `cas_tag_lazy_to_forward`. 

> Why is an assignment used instead of an initialisation in translcore.ml?

Good catch! This should be initialisation. I'll change these.

Issue #185, opened 2018-05-10
[Native backend] Unify stack scanning between GC and Dwarf


kayceesrk:
There are two distinct mechanisms in the runtime for unwinding stacks. One used by the GC and backtrace infrastructure (`caml_scan_stack`) and various bits used by DWARF (See https://github.com/ocamllabs/ocaml-multicore/blob/master/asmrun/amd64.S#L390-L404). Of course, there is valid reason for the mechanisms to be distinct; GC does not need to walk C frames, whereas DWARF does. However, there are bits between these two mechanisms that could be unified.

Issue #146, opened 2017-08-02
Effects and caml_callback


stedolan:
Currently, performing an effect doesn't work from within caml_callback (that is, when C code has called OCaml code). The tricky thing is managing the C state - we don't manage the C stack, so the stack-switching we do for effects isn't enough to pause and resume C functions.

Instead, we could use multiple C threads to have the operating system provide multiple C stacks, and switch between those.

kayceesrk:
The idea is to introduce a new attribute for external calls `@@performseffects` which means that the external call or its children (including any callbacks into OCaml) may perform effects that escape. Our goal is to ensure that C frames belonging to two different logical stacks don't end up on the same C stack.

Before calling such function, if there is a handler currently installed on the stack, then we instantiate a new C thread and make the function call in the new C thread. The old C thread is suspended on a condition variable waiting to be resumed. If this function performs an effect, whose handler is in the old C stack, we suspend the current C thread and switch to the old C thread. Any subsequent external calls performed by the handler computation (which are not marked with @@performseffects) will end up on the old C stack. Subsequently on resuming the continuation, we switch back to the new C thread through the condition variable mechanism. 

In this scheme, the penalty of switching C threads is paid only when performing effects whose continuations may capture C frames. Similar ideas were presented in the paper [Threads Yield Continuations](https://www.cs.indiana.edu/~dyb/pubs/LaSC-10-3-pp223-236.pdf).

rbjorklin:
I'm taking a peek at this project every now and then and just noticed that b6d624f was merged a couple of days back. Does that commit bring this any closer to being complete?

PR #227, opened 2018-11-13
Refactor stack manipulation primitives

stedolan:
The original point of this patch was to modify `caml_start_program` (used in C -> OCaml calls) to push one of  the following to the C stack, making a linked list of callback frames:
```c
struct c_stack_link {
  struct stack_info* stack;
  void* sp;
  struct c_stack_link* prev;
};
```
Having this linked list of callback frames has two advantages:

  - it's easy to tell if you've context-switched since the last callback (a prerequisite for #146, the original motivation)
  - it's easier to unwind the stack in DWARF

This should have been a straightforward enough change to `amd64.S` (where the assembly stubs for stack-switching are implemented), but making any changes to that file was surprisingly awkward, because:

  - it was really long
  - only very few registers could be safely used, because:
    * some assembly stubs called C, which destroys most integer registers. Values needed to be shuffled into callee-save registers around C calls.
    * when the registers were saved, they were saved to the C stack, so the switch to the C stack had to be done without clobbering any registers.

    This lack of registers made `amd64.S` very complicated: some macros were parameterised over which register could be used as a temporary, and there were several different implementations of register saving / restoring.
  - the DWARF CFI code was fragile, had long-range dependencies (a number here depends on an offset defined many lines later), and was barely tested by the testsuite.

So, what should have been a simple change turned into a week-long `amd64.S` hacking frenzy, with these results:

  - `amd64.S` went from 1233 to 960 lines.
  - added `c_stack_link` structure above.
  - c_stack_link is now used both for returning to OCaml after C calls and unwinding the stack in DWARF (closing #185), replacing `Stack_debugger_slot`.
  - removed `caml_switch_stack` and `caml_update_gc_regs_slot` C calls, so performing effects / resuming continuations no longer calls into C.
  - split `stack_info` in two, putting the handler information at the top of the stack so that it's easy to find when unwinding the stack / returning from the topmost function.
  - combined the fiber creation logic into a single `caml_runstack` primitive, replacing `caml_fiber_val_handler`, `caml_fiber_exn_handler`, `caml_handler`, `INIT_FIBER_USED` and a bunch of complicated DWARF code.
  - simplified the DWARF code when switching stacks - rather than using an extra stub frame on the target stack, `DW_CFA_def_cfa_expression` is used to express how to find the previous CFA from the target stack.
  - registers are now saved to a separate `malloc`'d area rather than the C stack, greatly simplifying the stack-switching code. In particular, there is now only one implementation of register save / restore.
  - unwind tests now actually tests unwinding on Linux as well as OSX. Also, the test now actually tests the backtrace produced by libunwind, rather than just verifying that it makes it to `main`.
  - added assertions to `amd64.S` (in debug mode, stack misalignment or corruption of c_stack_link trigger an int3, raising SIGTRAP).

This patch also deletes some code for WITH_FRAME_POINTERS mode. This mode was unused and untested, so I don't think it's worth keeping the code around for. If someone really wants it later on, I think it will be less work to resurrect it then than to maintain it until then.

Finally, performance of effects is improved, as measured by this microbenchmark:

```
effect Yield : unit
let () =
  for i = 1 to 1000000 do
    match
      for j = 1 to 10 do
        perform Yield
      done
    with
    | () -> ()
    | effect Yield k -> continue k ()
  done
```

whose runtime decreases from ~350ms to ~230ms.


asmcomp/amd64/proc.ml
```diff
@@ -273,15 +273,18 @@ let regs_are_volatile _rs = false
 (* Registers destroyed by operations *)
 
 let destroyed_at_c_call =
+  (* C calling conventions preserve rbp, but it is clobbered
+     by the code sequence used for C calls in emit.mlp, so it
+     is marked as destroyed. *)
   if win64 then
-    (* Win64: rbx, rbp, rsi, rdi, r12-r15, xmm6-xmm15 preserved *)
```
kayceesrk: Does clobbering rbp have an impact on the performance?

asmrun/amd64.S
```diff
@@ -122,95 +114,88 @@
...
-        movq	Caml_state(system_sp), REG; \
-        movq    %rsp, 16(REG); /* For DWARF CFI. See OFFXXX. */ \
-        movq    REG, %rsp; \
-        .cfi_def_cfa_offset 8 /* SYSCFAXXX */
+/* Switch from OCaml to C stack. Clobbers %r10, %r11.
+ * PUSHED is the number of bytes already pushed by this function.
+ * It is used to compute the backtrace. */
+#define SWITCH_OCAML_TO_C_NO_CTXT(PUSHED)                  \
+    /* Fill in Caml_state->current_stack->sp */            \
+        movq	Caml_state(current_stack), %r10;           \
```
kayceesrk: tab->space

```suggestion
        movq    Caml_state(current_stack), %r10;           \
```

asmrun/amd64.S
```diff
@@ -122,95 +114,88 @@
...
+        movq    %rsp, Stack_sp(%r10);                      \
+    /* Fill in Caml_state->c_stack */                      \
+        movq	Caml_state(c_stack), %r11;                 \
```
kayceesrk:
```suggestion
        movq    Caml_state(c_stack), %r11;                 \
```

asmrun/amd64.S
```diff
@@ -122,95 +114,88 @@
...
+        movq    %rsp, Cstack_sp(%r11);                     \
+        movq    %r10, Cstack_stack(%r11);                  \
+    /* Switch to C stack */                                \
+        movq    %r11, %rsp;                                \
+        .cfi_remember_state;                               \
+        .cfi_escape DW_CFA_def_cfa_expression, 5,          \
```
kayceesrk: Assuming lib-unwind test will ensure that the DWARF is correct. :-)

asmrun/amd64.S
```diff
@@ -122,95 +114,88 @@
...
         movq	Caml_state(exn_handler), %rsp; \
         .cfi_def_cfa_offset 16; \
-        POP_EXN_HANDLER(%r11)
+        POP_EXN_HANDLER
 
-/* Switch between OCaml stacks.
- * arguments : target stack (%rdi)
- * preserves : %rbx, %r8, %r9, %r12, %r13 */
+/* Switch between OCaml stacks. Clobbers %r11. Expects target stack in %r10. */
 #define SWITCH_OCAML_STACKS \
```
kayceesrk: :+1

asmrun/amd64.S
```diff
@@ -122,95 +114,88 @@
...
-        movq    $1, %rsi; \
-        call    GCALL(caml_switch_stack); \
-        CLEANUP_AFTER_C_CALL; \
-    /* Load ocaml stack and restore global variables */ \
-        LEAVE_FUNCTION; \
-        SWITCH_C_TO_OCAML(%r10)
-
+    /* Build caml_context at the bottom of the stack */ \
+        pushq   $0 ; CFI_ADJUST(8); \
+        PUSH_EXN_HANDLER; \
```
kayceesrk: Perhaps useful to add a comment here that we're pushing an exception handler just to save the exception pointer in the source stack.

asmrun/amd64.S
```diff
@@ -343,13 +336,20 @@ G(caml_system__code_begin):
 #define DW_CFA_val_expression     0x16
 #define DW_CFA_expression         0x10
 #define DW_CFA_def_cfa_expression 0x0f
+#define DW_CFA_restore            0x03
+#define DW_CFA_val_offset         0x14
+#define DW_REG_rbp                6
+#define DW_REG_rsp                7
+#define DW_REG_retaddr            16
+#define DW_OP_breg                0x70
```
kayceesrk: Which register is this on amd64?

kayceesrk:
The changes look good to me. I don't spot any obvious gotchas. We should merge this since the testsuite passes. Esp happy that the backtraces are now working. I can profile my programs again!

asmcomp/amd64/proc.ml
stedolan: I haven't measured, but if this makes performance worse it's a problem in the register allocator.

The current code spills the stack pointer to the C stack around `[@noalloc]` calls, while the new code saves it in `%rbp` instead. If all of the registers are in use, then this change will make the register allocator spill something. Letting the register allocator check whether it needs to spill something seems better than unconditionally spilling `%rsp`, which is definitely in use.

asmrun/amd64.S
stedolan: Yep, if this was wrong unwind_test wouldn't get past `caml_c_call`.

asmrun/amd64.S
stedolan: Done

asmrun/amd64.S
stedolan: The 'retaddr' reg is a fiction of DWARF. You specify the return address of the next frame as the contents of a pseudoregister (or, on e.g. POWER, an actual register). The standard definition of the 'retaddr' reg on x86_64 is that it's the fictional register 16 that was "spilled" just after the CFA.

I'll delete this, it's not actually needed at the moment. An earlier version of the DWARF code had DWARF expressions to find the previous stack pointer and return address on a different stack, but it's easier to do the traversal for the CFA, and leave the CFA -> retaddr and CFA -> stack computations as standard.

PR #233, opened 2019-03-20
create a separate caml_obj_force_promote_to

sadiqj:
Just a quick PR to check this was in line with our earlier conversation @stedolan.

### DIFF
--- a/byterun/obj.c
+++ b/byterun/obj.c
@@ -40,22 +40,27 @@ CAMLprim value caml_obj_is_block(value arg)
 
 CAMLprim value caml_obj_tag(value arg)
 {
-  if (Is_long (arg)){
-    return Val_int (1000);   /* int_tag */
-  }else if ((long) arg & (sizeof (value) - 1)){
-    return Val_int (1002);   /* unaligned_tag */
-  }else{
+  if (Is_long(arg))
+  {
+    return Val_int(1000); /* int_tag */
+  }
+  else if ((long)arg & (sizeof(value) - 1))
+  {
+    return Val_int(1002); /* unaligned_tag */
+  }
+  else
+  {
     return Val_int(Tag_val(arg));
   }
 }
 
-CAMLprim value caml_obj_set_tag (value arg, value new_tag)
+CAMLprim value caml_obj_set_tag(value arg, value new_tag)
 {
-  Tag_val (arg) = Int_val (new_tag);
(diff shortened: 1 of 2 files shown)


stedolan:
Looks good!

For the sake of anyone following along: @sadiqj noticed that when Gc.promote_to is used to optimise parallel data structures, a good optimisation is to skip the promotion if there's only one thread running. However, since there are a couple of places where promotion is necessary for correctness, skipping it all the time seems dangerous, hence the split into two primitives.

I'll hold off on merging this just yet, though, as another change is needed for this to work: we do need to do the actual promotion if a second domain is started before the next GC. Rather than keeping track of promotions that have been elided, I think the best approach is just to do a minor GC before starting the second domain.

byterun/obj.c
```diff
@@ -150,13 +150,34 @@ CAMLprim value caml_obj_compare_and_swap (value v, value f, value oldv, value ne
   return Val_int(res);
 }
 
-/* caml_promote_to(obj, upto) promotes obj to be as least as shared as upto */
+/* caml_promote_to(obj, upto) may promote obj to be as least as shared as upto.
+   Skips promotion if there is only a single domain. This should only be used in
+   situations where there is a performance need to 'front-load' the promotion
+   across domains, such as for queues. If you need to guarantee promotion to the
+   major heap for correctness, */
```
stedolan: I think half of this sentence got lost somewhere!

byterun/obj.c
```diff
@@ -150,13 +150,34 @@ CAMLprim value caml_obj_compare_and_swap (value v, value f, value oldv, value ne
...
 CAMLprim value caml_obj_promote_to (value obj, value upto)
 {
-  if (Is_block(upto) && Is_minor(upto)) {
+  if (Is_block(upto) && (Is_minor(upto) || caml_domain_alone()))
+  {
```
stedolan: Indentation in the runtime is [K&R style](https://en.wikipedia.org/wiki/Indentation_style#K&R_style): function opening braces go on their own line, but other opening braces (for `if`, `while`, etc.) go on the same line as their statement.

byterun/obj.c
sadiqj: Whoops. I'm really not sure where that went. Fixed.

byterun/obj.c
sadiqj: Thanks. This should be resolved now.

PR #240, opened 2019-04-21
Proposed implementation of Threads in terms of Domain and Atomic

jhwoodyatt:
This is a request for review and comment on a candidate for implementing the `Threads` library in terms of the new `Domain` and and `Atomic` modules in the Multicore OCaml standard library. This change, when combined with my other patch for `Unix.system` to not use `Unix.fork` makes the stock Dune 1.9.2 package compile and run properly on Multicore OCaml. (Or at least, it did until something in the `master` branch related to Lazy values introduced a regression.)

Some caveats are worth noting up front:

Firstly, I've made a half-hearted attempt at introducing this as a third library without overwriting the existing _vmthread_ and _systhread_ implementations. I noticed this decision is visible when trying to compile the **ounit** external package, so I'm anticipating the need to revisit some of these interoperability concerns.

Secondly, I've tried to improve the assurance of correctness by using the unit tests, but unfortunately, this pull request does not currently pass all of them reliably, and moreover, the only reason it can pass most of them at all is that I inserted a call to `Gc.minor` in the blocking section entrance hook. Obviously, that's a suboptimal workaround. Also, the `swapchan` unit test looks to my eye like it has a concurrency error of its own, and can only pass reliably if the Thread library is expected to have some guarantee about context switches that I don't think it actually promises. In short, I think it only accidentally passes reliably in the mainline monocore distribution.

(This introductory message is updated from its previous content.)


asmrun/startup.c
```diff
@@ -74,8 +74,10 @@ static void init_segments(void)
 }
 
 /* These are termination hooks used by the systhreads library */
```
jhwoodyatt: Probably not the most friendly way to remove this logic. Should properly delete this logic if it's really obsolete.

asmrun/startup.c
```diff
@@ -128,10 +130,12 @@ value caml_startup_common(char_os **argv, int pooling)
   else
     exe_name = caml_search_exe_in_path(exe_name);
   caml_init_argv(exe_name, argv);
```
jhwoodyatt: Also here.

byterun/domain.c
```diff
@@ -1055,7 +1054,9 @@ CAMLprim value caml_ml_domain_yield_until(value t)
     if (ts < caml_time_counter ()) {
       ret = Val_int(0); /* Domain.Sync.Timeout */
       break;
```
jhwoodyatt: My recollection is that this change was necessary to address an intermittently observed deadlock failure in the threads unit tests. Unfortunately, the test for _pr5325_ is still failing deterministically. Not sure why.  It does seem to be deadlocking on this same call to `caml_plat_timedwait`.

configure
```diff
@@ -1280,7 +1280,7 @@ config UNIX_OR_WIN32 "$unix_or_win32"
 config UNIXLIB "$unixlib"
 config GRAPHLIB "$graphlib"
```
jhwoodyatt: The changes here are provisional. Not sure how this should be addressed properly.

otherlibs/dthreads/event.ml
```diff
@@ -0,0 +1,276 @@
+(**************************************************************************)
```
jhwoodyatt: I just basically copied the `Event` library from **systhreads**. It's weird, actually, there are two `Event` library implementations already. This is a copy-paste that adds a third.

otherlibs/dthreads/thread.mli
```diff
@@ -0,0 +1,46 @@
+(** Lightweight threads. *)
```
jhwoodyatt: Needs Doxygen.

otherlibs/dthreads/mutex.mli
```diff
@@ -0,0 +1,6 @@
+  type t
```
jhwoodyatt: Needs Doxygen.

otherlibs/dthreads/condition.mli
```diff
@@ -0,0 +1,6 @@
+type t
```
jhwoodyatt: Needs Doxygen.

jhwoodyatt:
_— This branch is not ready for merge. —_

I have reimplemented this work with two noteworthy improvements:

1. It passes the test suite most of the time (and fails for reasons that seem to be either unrelated to the Thread API or because the test suite has what I consider errors).

2. It has a removable global interlock (enabled at startup) that prevents more than one thread from executing at a time. It's possible for a domain that does not use the thread API to run concurrently with others that do, but the interlock applies as soon as a domain enters the thread module.

Also, the logic is peppered with verbose logging of a very idiosyncratic form. I would like to remove that logic before this branch is considered eligible for merging.

jhwoodyatt:
After rebasing, many of the tests, which were previously failing for reasons I suspected were caused by other problems, are now passing. There are still some tests that need attending, and I'll get to those soonish. I don't feel like there is enormous pressure to get this working now now now, but if there is some urgency I'm not sensing, then please respond.

jhwoodyatt:
And with a little bit of additional effort to make this `threads` library replace the obsolete ones, I'm happy to see that this branch builds and runs stock `Dune 2.4.0`. It also builds `ounit2` and successfully uses it to run all the tests in my `orsetto` project. Also, with a little bit of modification, the `omake` build tool also works fine, using this threads library.

Still needs some polish, but I feel like this is a noteworthy intermediate achievement.

avsm:
Many thanks for the sustained efforts on this PR!  We're going to review this soon as the rest of the GC is now pretty stable.

jhwoodyatt:
I'm grateful for the opportunity to make what I hope will be a noteworthy contribution. I hope.

avsm:
Now that the rebasing to 4.10 is complete in the default branch, @Engil has been working on rebasing this PR from 4.06. His working tree is at https://github.com/Engil/ocaml-multicore/tree/parallel_minor_gc_4_10%2Bdthreads (to avoid any duplication of efforts).

Engil:
Hello,
Thank you for your work on this PR @jhwoodyatt.
As @avsm mentioned, I will be assisting with upstreaming your contribution to OCaml Multicore, starting with rebasing it to our latest development branch.

A small update on the rebase work to 4.10 parallel_minor:
The rebase is still unclean and slightly broken, but I managed to get libthread building and running the testsuite. 
Minor notes: the `signal.ml` test, which used to not pass, now seems to reliably pass.
However the testsuite now fails on the `delayintr.ml` test (which was added at some point after 4.06.0), which I shall investigate in the next few days.
This is still very early observations, I will investigate these changes more properly once the rebasing work reaches an acceptable state.

jhwoodyatt:
Thank you for doing this! I'll follow the new branch. Feel free to close this one when that once that one is made authoritative for this change.

abbysmal:
Closing this Pull Request, as #342 is now the authoritative PR for merging this contribution.

PR #241, opened 2019-04-21
Implement Unix.system without Unix.fork

jhwoodyatt:
This is a request for review and comment. I have reimplemented `Unix.system` and various cognates without using `Unix.fork`. Instead, I've written necessary variations on the `Unix.create_process` functions that call `fork(2)` and various `execxxx(2)` functions inside their own singular blocking sections. This is one of the two patches I've found necessary to compile stock Dune 1.9.2 on Multicore OCaml.

It's worth noting here that I needed to write a private `Unix.ck_execvpe` function to deal with the sad fact that `execvpe(2)` is a Linux system call that isn't available on FreeBSD and macOS. The `Unix.ck_execvpe` function is a refactoring of the logic that used to be `Unix.execvpe_ml` and has importantly different semantics.

If I can make this pull request acceptable in Multicore OCaml, then I'll commit to helping make additional necessary changes to land this in the upstream OCaml distribution if invited to do so.

### DIFF
--- a/otherlibs/unix/unix.ml
+++ b/otherlibs/unix/unix.ml
@@ -200,71 +200,6 @@ type wait_flag =
     WNOHANG
   | WUNTRACED
 
-external execv : string -> string array -> 'a = "unix_execv"
-external execve : string -> string array -> string array -> 'a = "unix_execve"
-external execvp : string -> string array -> 'a = "unix_execvp"
-external execvpe_c :
-            string -> string array -> string array -> 'a = "unix_execvpe"
-
-let execvpe_ml name args env =
-  (* Try to execute the given file *)
-  let exec file =
-    try
-      execve file args env
-    with Unix_error(ENOEXEC, _, _) ->
-      (* Assume this is a script and try to execute through the shell *)
-      let argc = Array.length args in
-      (* Drop the original args.(0) if it is there *)
-      let new_args = Array.append
-        [| "/bin/sh"; file |]
-        (if argc = 0 then args else Array.sub args 1 (argc - 1)) in
-      execve new_args.(0) new_args env in
-  (* Try each path element in turn *)
-  let rec scan_dir eacces = function
-  | [] ->
-      (* No matching file was found (if [eacces = false]) or
-         a matching file was found but we got a "permission denied"
-         error while trying to execute it (if [eacces = true]).
-         Raise the error appropriate to each case. *)
-      raise (Unix_error((if eacces then EACCES else ENOENT),
-                        "execvpe", name))
-  | dir :: rem ->
-      let dir =  (* an empty path element means the current directory *)
-        if dir = "" then Filename.current_dir_name else dir in
-      try
-        exec (Filename.concat dir name)
-      with Unix_error(err, _, _) as exn ->
-        match err with
-        (* The following errors are treated as nonfatal, meaning that
-           we will ignore them and continue searching in the path.
-           Among those errors, EACCES is recorded specially so as
-           to produce the correct exception in the end.
-           To determine which errors are nonfatal, we looked at the
-           execvpe() sources in Glibc and in OpenBSD. *)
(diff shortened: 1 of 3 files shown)


otherlibs/unix/unix.ml
```diff
@@ -380,6 +315,76 @@ external unlink : string -> unit = "unix_unlink"
 external rename : string -> string -> unit = "unix_rename"
 external link : string -> string -> unit = "unix_link"
 
+external execv : string -> string array -> 'a = "unix_execv"
+external execve : string -> string array -> string array -> 'a = "unix_execve"
+external execvp : string -> string array -> 'a = "unix_execvp"
+external execvpe_c :
+            string -> string array -> string array -> 'a = "unix_execvpe"
+
```
jhwoodyatt: This function is a refactoring of the logic in the old `Unix.execvpe_ml` for reusabilty in `Unix.create_process` and its cognates.

otherlibs/unix/unix.ml
```diff
@@ -938,65 +943,47 @@ let rec waitpid_non_intr pid =
...
+    "unix_fork_and_execve_bytecode"
+    "unix_fork_and_execve"
+
+let create_process = fork_and_execvp
+
+let create_process_env path args envs stdin stdout stderr =
+  try
+    fork_and_execvpe path args envs stdin stdout stderr
+  with
+  | Unix_error(ENOSYS, _, _) ->
```
jhwoodyatt: So this is where a difference in error case behavior may be observable in downstream users.

On platforms that don't have `execvpe(2)`, e.g. macOS and other BSD variants, the `ENOSYS` exception case is used, and accordingly the argument translation using `ck_execvpe` and dispatch to `fork_and_execve` has the effect that failure to locate the executable image in the path happens before the call to `fork(2)` rather than afterward and prior to the call to `execve(2)`. Specifically, the old logic would iterate in the child process looking for a successful call to `execve(2)` if the first call returned an `ENOEXEC` error. The new logic does not attempt to call `execve(2)` more than once.

Accordingly, the user visible behavior change is that `Unix.create_process_env` can exception where it would previously succeed, _in the case where searching the path finds a valid executable later in the search after an earlier non-executable one is found._

jhwoodyatt: Worth noting: I suspect the current behavior of `Unix.create_process_env` may be associated with security considerations, and I think an argument could be made that the change I am proposing could be regarded as a security improvement.

kayceesrk:
Hi @jhwoodyatt, I am trying to understand the PR. Why does calling `fork` and `execxxx` in a blocking section help? Also, this may be something we can upstream to `ocaml/ocaml` as a multicore prerequisite. Apologies for getting to this late.

kayceesrk:
Ok. I seem to have missed the point that `fork` and `execvp` are in the same blocking section. Again, could you explain why this helps?

xavierleroy:
This helps because, as far as I understand, `Unix.fork` cannot be used in multithreaded Multicore OCaml programs:
https://github.com/ocaml-multicore/ocaml-multicore/blob/172fe18bbc7fcdaa1fc286869c28a2c59ecbf27a/otherlibs/unix/fork.c#L27-L29

Besides, it has already been suggested to use a different implementation of `Unix.system` and high-level process creation functions, for performance reasons too.  I have some code that I'll post as a pull request on ocaml/ocaml shortly.

kayceesrk:
@Engil Would you be able to work on porting https://github.com/ocaml/ocaml/pull/9573 to multicore?

Engil:
@kayceesrk Sure, putting it on my roadmap after #240 is rebased.

jhwoodyatt:
Unless I'm mistaken, this pull request can be closed now. It was mooted by #9573.

kayceesrk:
We have #342 now, which is the multicore version of ocaml/ocaml#9573. Closing.

PR #244, opened 2019-04-27
Read fault return value is made a global root.

kayceesrk:
Read fault return value field used to be a local root. It certainly needs to be a root since a GC may occur after the return value field is set. But the return value field is written to by a different domain different from the domain issuing the read fault request. The domain issuing the read fault request has the return value field in its local root set. This introduces a data race between the remote domain writing the return value of a read fault to the field and a minor GC (specifically, `oldify_one`) in the read fault issuing domain writing the default value (`Val_unit`) to the field. The race may overwrite the result written by the remote domain.

The fix is to make the read fault return value field a global root, which is only scanned when all the domains are blocked at the beginning of the major cycle, which removes the race. The commit creates one global root per domain for the read fault return value field; there can be only one outstanding read fault request per domain.

### DIFF
--- a/byterun/caml/domain_state.tbl
+++ b/byterun/caml/domain_state.tbl
@@ -1,31 +1,88 @@
 DOMAIN_STATE(volatile uintnat, young_limit)
+/* Minor heap limit. Typically young_limit == young_end, but this field is set
+ * by other domains to signal this domain by causing a spurious allocation
+ * failure. */
+
 DOMAIN_STATE(char*, young_ptr)
+/* Minor heap pointer */
+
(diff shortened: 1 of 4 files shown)


stedolan:
Wow, this is scarily subtle. (I remember being in this hole before, I thought the root handling during faults was right this time!).

Why does the sending domain do a minor GC? Hmm, is it responding to another fault? Is it possible to trigger this bug with less than three threads?

kayceesrk:
The buggy runs had at least 3 threads. You shouldn't be able to trigger
this with less than three threads.

PR #253, opened 2019-06-24
Fix #252 by making atomic exchange always call the write barrier.

### LINKED ISSUE #252: Failed assertion with atomics containing block values
I ran into segmentation faults while overwriting atomic cells containing option values. The optional value was sometimes corrupted, leading in my case to random integers, or invalid pointers.

A minimal example:
```ocaml
type u = U of unit

let make_cell (x : unit) : u Atomic.t =
  let cell = Atomic.make (U x) in
  Atomic.set cell (U x) ;
  cell

(* the error shows up with an array of length 256 or larger *)
let a = Array.make 256 (make_cell ())
```

This code itself is not apparently failing, but compiling and running it in debug mode:

```sh
ocamlc test.ml -o test.byte && ocamlrund test.byte
```

reveals a failed assertion:

```
### OCaml runtime: debug mode ###
[00] file minor_gc.c; line 555 ### Assertion failed: caml_addrmap_contains(&young_young_ptrs, (value)f)
Abandon (core dumped)
```

In case it matters:

 *  architecture = x86_64
 *  operating system = Linux 5.1.5-arch1-2-ARCH
 *  compiler = Multicore OCaml 4.06.1 (from the opam switch)

stedolan:
`caml_atomic_exchange` didn't call the write barrier if the atomic being modified was young. But because of the young-young table, we need to do work in the write barrier even when mutating young objects.

I suspect this isn't the only codepath that has this bug: the assumption that the write barrier does no work when young values are modified is easy to make, because it's true in stock OCaml.

### DIFF
--- a/byterun/memory.c
+++ b/byterun/memory.c
@@ -253,8 +253,8 @@ CAMLprim value caml_atomic_exchange (value ref, value v)
     /* See Note [MM] above */
     atomic_thread_fence(memory_order_acquire);
     ret = atomic_exchange(Op_atomic_val(ref), v);
-    write_barrier(ref, 0, ret, v);
   }
+  write_barrier(ref, 0, ret, v);
   return ret;
 }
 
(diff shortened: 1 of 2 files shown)


kayceesrk:
Good catch. Best to merge this one now and then keep an eye out for other such patterns.

PR #259, opened 2019-07-29
changes to the PUSH_EXN_HANDLER/POP_EXN_HANDLER assembly

sadiqj:
changes to the PUSH_EXN_HANDLER/POP_EXN_HANDLER assembly to avoid some problematic depdendencies between instructions

### DIFF
--- a/asmrun/amd64.S
+++ b/asmrun/amd64.S
@@ -117,7 +117,7 @@
    (Stored as an offset, to survive stack reallocations). Clobbers %r11 */
 #define PUSH_EXN_HANDLER \
         movq    Caml_state(exn_handler), %r11; \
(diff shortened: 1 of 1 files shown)


stedolan:
I think this is fine, but some comments in amd64.S need updating. e.g. SWITCH_C_TO_CAML says that it clobbers %r11, but with this patch it clobbers %r10 also.

kayceesrk:
My only concern was whether clobbering `%r10` is ok. I can't convince myself otherwise. @sadiqj please update the comments and then we can merge.

sadiqj:
I _think_ this should now cover all the places in the documentation that need to be changed.

Thanks for the review.

PR #260, opened 2019-07-31
changes to caml_hash to avoid needing to initialize HASH_QUEUE_SIZE values on every call

sadiqj:


stedolan:
This looks great, and probably worth proposing upstream. I'd prefer something that extended the CAMLxparamN interface to support this sort of usage, though, rather than poking around with the undocumented caml__roots_queue variables.

sadiqj:
So this change specifically is only needed on multicore because the read barrier could cause a GC and thus the queue elements need to be registered as local roots.

You're right, a macro would work much better here. Would something like `CAMLxparamNresize(x, n)` which we could modify the `CAMLxparamN(x,n)` to use as well, work?

sadiqj:
I've moved this to a macro, which avoids needing to poke around with undocumented variables.

With this change the delta between vanilla and multicore caml_hash is much smaller but multicore is still slower because of the need to register the local roots because we might be hashing something with fields we need to read.

We could restructure the method to not register the local roots and call a another method that does, if we find ourselves needing to hash something with fields we need to access. That would mean roughly similar performance for multicore caml_hash to vanilla for things that go through the fast path.

stedolan:
New version looks correct and I like the new `CAMLxparamNextend` macro.

Does this version perform OK on small blocks like `(1,2)`? It seems like as soon as a block that needs to be scanned is hit (on line 280), it always initialises all 256 stack entries, even if it only needs two.

sadiqj:
You're right, I thought it was only resizing as needed but `sz` is the maximum size. I'll get that changed so we're only initializing based on the number of fields. Good catch.

byterun/hash.c
```diff
@@ -273,7 +277,20 @@ CAMLprim value caml_hash(value count, value limit, value seed, value obj)
...
+        /* initialize any elements of the queue that are uninitialized */
+
+        len = Wosize_val(v);
+
+        /* we may have more fields than the maximum to use for the hash */
+        j = wr+len > sz ? sz : wr+len;
+
+        for( i = wr ; i < j; i++ ) {
+          queue[i] = Val_unit;
+        }
```
stedolan: This looks completely right, but the fact that the indexing logic is different from the loop below (wr..j instead of 0..len) meant it took me a while to read. I think it would be simpler if the two loops had the same indexing logic, e.g. by replacing the first with something like
```c
        for (i = 0; i < len; i++) {
          if (wr + i >= sz) break;
          queue[wr + i] = Val_unit;
        }
```
or equally, by replacing the second with something like
```c
        for( ; wr < j; wr++ ) {
          caml_read_field(v, i, &queue[wr]);
        }
```

stedolan:
Thanks!

stedolan:
Github seems to think Travis hasn't completed yet, but Travis says it has. Not sure what's up there.

PR #261, opened 2019-08-01
#129 Added a longjmp context to the newly created domain, to avoid bringing down the runtime 

rmdouglas:
I thought I'd take a look at issue #129.

I've created a commit causes any exceptions raised during domain initialization, to get back to the parent domain, get raised from the call to Spawn.
I suspect there's more that needs to be done when the longjmp occurs on the fresh domain to clean up, and I haven't really looked at what needs to be done for native code, but I'm hoping that I'm on the right track.
There are also (currently) no tests, as I'm unsure how to trigger an exception/OOM without making changes to the code solely for testing (I wrote the change with a hard coded OOM in the create_domain() function)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@
 *.out.dSYM
 *.swp
 _ocamltest
+.vscode
 
 # local to root directory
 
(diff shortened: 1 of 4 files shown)


rmdouglas:
I've written up a quick module+c_stub which wraps many of the allocation methods in the runtime, so I'll be able to write some tests which throw OOM's without making modifications to the production code.

PR #262, opened 2019-08-01
Inline loads specialized

anmolsahoo25:
Adding an entry in `translcore.ml` to specialize loads based on whether the load is to an `immediate` or `pointer`

### DIFF
--- a/bytecomp/translcore.ml
+++ b/bytecomp/translcore.ml
@@ -464,7 +464,7 @@ let specialize_primitive p env ty ~has_constant_constructor =
     | (Pmakeblock(tag, mut, None), fields) ->
         let shape = List.map (Typeopt.value_kind env) fields in
         Pmakeblock(tag, mut, Some shape)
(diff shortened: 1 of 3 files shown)


bytecomp/translcore.ml
```diff
@@ -464,6 +464,11 @@ let specialize_primitive p env ty ~has_constant_constructor =
     | (Pmakeblock(tag, mut, None), fields) ->
         let shape = List.map (Typeopt.value_kind env) fields in
         Pmakeblock(tag, mut, Some shape)
+    | (Patomic_load _, _) ->
```
stedolan: Looks good below, but this pattern should match `Patomic_load { immediate_or_pointer = Pointer }` rather than `Patomic_load _`: if the atomic load is already immediate, then we don't want to redetect it and risk turning it back into Pointer.

anmolsahoo25: All right. That makes sense. I will do that. Thanks you!

PR #263, opened 2019-08-09
alloc_stack_noexc should use the noexc version of caml_stat_alloc

rmdouglas:
I think the "alloc_stack_noexc" should be calling "caml_stat_alloc_noexc" instead of "caml_stat_alloc", as it checks the returned value for NULL, and going by the name, shouldn't raise an (ocaml) exception.


kayceesrk:
lgtm. Thanks!

PR #264, opened 2019-08-09
Fixes for error handling during domain creation 

rmdouglas:
Fixes for issue #129
Modified the calls in create_domain to use non-raising variants (caml_stat_alloc_noexc instead of caml_stat_alloc), and added some cleanup error handling in the case of failure.


rmdouglas:
I've cleaned up the changes, and squashed them into a single commit.

PR #265, opened 2019-09-21
cache local roots pointer to avoid needing to look it up each time

sadiqj:
As it stands Caml_state points to a thread local global. Examining the hot assembly for some of the microbenchmarks it looks like this is looked up multiple times in the compiled code.

Caching a pointer to Caml_state at the start of a method eliminates these and shortens the sequence of generated instructions for the CAML* macros.

This has the most impact on the test_many_args_noalloc and test_many_args_alloc microbenchmarks which see a 10% reduction in runtime.

### DIFF
--- a/byterun/caml/memory.h
+++ b/byterun/caml/memory.h
@@ -286,7 +286,8 @@ struct caml__roots_block {
 */
 
 #define CAMLparam0() \
-  struct caml__roots_block *caml__frame = CAML_LOCAL_ROOTS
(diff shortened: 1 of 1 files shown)


stedolan:
This looks good!

I was thinking about this and it occurred to me that we might be able to abuse GCC's `__attribute__((const))` to help with other repeated uses of `Caml_state`. If `Caml_state` is defined as a call to a function with `__attribute__((const))`, then GCC will know it's OK to cache the result.  (I'm not sure whether Clang supports this attribute).

kayceesrk:
Lgtm. I'll merge this one.

PR #268, opened 2019-11-27
Fix for a few space leaks

sadiqj:
This PR fixes a couple of space leaks that were exposed by some domain spawning/terminating stress tests.

The first set are per-domain structures that failed to be cleaned up on termination - these are all in `domain.c`.

The second issue is in `pool_find` in `shared_heap.c` which does not currently check the global pools when looking for a new pool. In discussions with @stedolan this seems to have been a bug rather than a deliberate design choice. This PR fixes that by searching for pools in the global list that have some space or could be swept to clear some up.

The third issue that gets fixed is a GC pacing one. Right now if domains are very short lived (as they are in this stress test), we don't factor in allocations they did during their brief existence and this means we don't sweep as much as we should so we end up allocating new pools rather than using full ones that might be freed up.

### DIFF
--- a/byterun/domain.c
+++ b/byterun/domain.c
@@ -212,6 +212,7 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
     domain_state->id = d->id;
     domain_state->unique_id = d->interruptor.unique_id;
     d->state.state = domain_state;
+    domain_state->critical_section_nesting = 0;
 
     if (caml_init_signal_stack() < 0) {
       goto init_signal_stack_failure;
@@ -880,6 +881,7 @@ static void domain_terminate()
   caml_ev_pause(EV_PAUSE_YIELD);
   s->terminating = 1;
   while (!finished) {
+    caml_orphan_allocated_words();
     caml_finish_sweeping();
     caml_empty_minor_heap();
     caml_finish_marking();
@@ -914,12 +916,17 @@ static void domain_terminate()
 
   caml_delete_root(domain_state->read_fault_ret_val);
   caml_stat_free(domain_state->final_info);
+  caml_stat_free(domain_state->ephe_info);
   caml_teardown_major_gc();
   caml_teardown_shared_heap(domain_state->shared_heap);
   domain_state->shared_heap = 0;
   caml_free_minor_tables(domain_state->minor_tables);
   domain_state->minor_tables = 0;
   caml_free_signal_stack();
+  
+  if(domain_state->current_stack != NULL) {
+    caml_free_stack(domain_state->current_stack);
+  }
 
   if (Caml_state->critical_section_nesting) {
     Caml_state->critical_section_nesting = 0;
--- a/byterun/major_gc.c
+++ b/byterun/major_gc.c
@@ -66,6 +66,8 @@ static atomic_uintnat num_domains_to_ephe_sweep;
 static atomic_uintnat num_domains_to_final_update_first;
 static atomic_uintnat num_domains_to_final_update_last;
 
+static atomic_uintnat terminated_domains_allocated_words;
+
 gc_phase_t caml_gc_phase;
 
 uintnat caml_get_num_domains_to_mark () {
@@ -175,6 +177,10 @@ static int no_orphaned_work ()
     orph_structs.ephe_list_todo == 0 &&
     orph_structs.ephe_list_live == 0 &&
     orph_structs.final_info == NULL;
+}
+
+void caml_orphan_allocated_words() {
+    atomic_fetch_add(&terminated_domains_allocated_words, Caml_state->allocated_words);
 }
 
 void caml_add_orphaned_ephe(value todo_head, value todo_tail,
@@ -333,7 +339,13 @@ static uintnat default_slice_budget() {
   uintnat heap_size = caml_heap_size(Caml_state->shared_heap);
   heap_words = (double)Wsize_bsize(heap_size);
   uintnat heap_blocks = caml_heap_blocks(Caml_state->shared_heap);
-  p = (double) Caml_state->allocated_words * 3.0 * (100 + caml_percent_free)
+
+  uintnat saved_terminated_words = terminated_domains_allocated_words;
+  if( saved_terminated_words > 0 ) {
+    while(!atomic_compare_exchange_strong(&terminated_domains_allocated_words, &saved_terminated_words, 0));
+  }
+
+  p = (double) (saved_terminated_words + Caml_state->allocated_words) * 3.0 * (100 + caml_percent_free)
       / heap_words / caml_percent_free / 2.0;
 
   computed_work = (intnat) (p * (heap_blocks + (heap_words * 100 / (100 + caml_percent_free))));
--- a/byterun/major_gc.c
+++ b/byterun/major_gc.c
@@ -748,6 +748,17 @@ void caml_accum_heap_stats(struct heap_stats* acc, const struct heap_stats* h)
   acc->large_blocks += h->large_blocks;
 }
 
+void caml_remove_heap_stats(struct heap_stats* acc, const struct heap_stats* h)
+{
+  acc->pool_words -= h->pool_words;
+  acc->pool_live_words -= h->pool_live_words;
+  acc->pool_live_blocks -= h->pool_live_blocks;
+  acc->pool_frag_words -= h->pool_frag_words;
+  acc->large_words -= h->large_words;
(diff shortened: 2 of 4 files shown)


sadiqj:
The testsuite failures for the debug build on travis have highlight an important issue which we need to think about:

`  Assert(local->stats.pool_words == pool_stats.alloced);
  Assert(local->stats.pool_live_words == pool_stats.live);
  Assert(local->stats.pool_live_blocks == pool_stats.live_blocks);
  Assert(local->stats.pool_frag_words == pool_stats.overhead);
  Assert(local->stats.pool_words -
         (local->stats.pool_live_words + local->stats.pool_frag_words)
         == pool_stats.free);
  Assert(local->stats.large_words == large_stats.alloced);
  Assert(local->stats.large_blocks == large_stats.live_blocks);`

We don't currently update the local stats for global pools that we adopt. Some of these are easy to add, like words but others are going to require scanning the pools as we adopt them - which seems reasonably excessive in `pool_find`.

byterun/shared_heap.c
```diff
@@ -220,6 +232,62 @@ static pool* pool_find(struct caml_heap_state* local, sizeclass sz) {
...
+  }
+
+  /* There were no global avail pools, so let's adopt the full ones and try 
+     our luck sweeping them later on */
+  if( !r ) {
+    struct heap_stats tmp_stats = { 0 };
+
+    for(v = pool_freelist.global_full_pools[sz]; v; v = v->next ) {
+      calc_pool_stats(v, sz, &tmp_stats);
+    }
```
sadiqj: I could move these pools somewhere else temporarily, unlock the platform lock, calculate their stats and _then_ move them to the local full pools.

byterun/shared_heap.c
```diff
@@ -188,25 +186,39 @@ static void pool_release(struct caml_heap_state* local, pool* pool, sizeclass sz
...
-    local->num_free_pools++;
-    pool->next = local->free_pools;
-    local->free_pools = pool;
-  } else {
-    /* FIXME: give free pools back to the OS */
-    caml_plat_lock(&pool_freelist.lock);
-    pool->next = pool_freelist.free;
-    pool_freelist.free = pool;
-    caml_plat_unlock(&pool_freelist.lock);
+  /* FIXME: give free pools back to the OS */
```
ctk21: Given that local->free_pools and local->num_free_pools are being factored out in this change. To remove future confusion, can we also remove them from the code in pool_acquire?

byterun/shared_heap.c
```diff
@@ -220,6 +232,62 @@ static pool* pool_find(struct caml_heap_state* local, sizeclass sz) {
...
+  caml_plat_lock(&pool_freelist.lock);
+  if( pool_freelist.global_avail_pools[sz] ) {
+    r = pool_freelist.global_avail_pools[sz];
+
+    if( r ) {
+      pool_freelist.global_avail_pools[sz] = r->next;
+      r->next = 0;
+      local->avail_pools[sz] = r;
+
+      int free_objs = 0;
```
ctk21: Apart from the assert in debug mode, I can't see next_obj and free_objs actually being used. Should we but this in DEBUG guards (or even remove)?

byterun/shared_heap.c
```diff
@@ -188,25 +186,39 @@ static void pool_release(struct caml_heap_state* local, pool* pool, sizeclass sz
...
+    header_t hd = (header_t)*p;
+    if (hd) {
+      s->pool_live_words += Whsize_hd(hd);
+      s->pool_frag_words += wh - Whsize_hd(hd);
+      s->pool_live_blocks++;
+    }
+
+    p += wh;
   }
+  CAMLassert(end - p == wastage_sizeclass[sz]);
```
ctk21: This PR is introducing CAMLassert, but elsewhere in the file it is Assert, is this intentional?

kayceesrk:
@sadiqj Can you provide a high-level overview of the problem and the fix in the description please?

byterun/shared_heap.c
kayceesrk: I think this is reasonable. Unless we find contention in practice, let's not complicate this code.

byterun/shared_heap.c
```diff
@@ -220,6 +232,62 @@ static pool* pool_find(struct caml_heap_state* local, sizeclass sz) {
   r = local->avail_pools[sz];
   if (r) return r;
 
+/* Haven't managed to find a pool locally, try the global ones */
```
kayceesrk: Can you briefly explain what this addition is doing and why the earlier code did not have this path?

byterun/domain.c
```diff
@@ -212,6 +212,7 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
     domain_state->id = d->id;
     domain_state->unique_id = d->interruptor.unique_id;
     d->state.state = domain_state;
+    domain_state->critical_section_nesting = 0;
```
kayceesrk: lgtm.

byterun/domain.c
```diff
@@ -880,6 +881,7 @@ static void domain_terminate()
   caml_ev_pause(EV_PAUSE_YIELD);
   s->terminating = 1;
   while (!finished) {
+    caml_orphan_allocated_words();
```
kayceesrk: lgtm

byterun/domain.c
```diff
@@ -914,12 +916,17 @@ static void domain_terminate()
 
   caml_delete_root(domain_state->read_fault_ret_val);
   caml_stat_free(domain_state->final_info);
+  caml_stat_free(domain_state->ephe_info);
```
kayceesrk: lgtm

byterun/domain.c
```diff
@@ -914,12 +916,17 @@ static void domain_terminate()
...
   caml_teardown_major_gc();
   caml_teardown_shared_heap(domain_state->shared_heap);
   domain_state->shared_heap = 0;
   caml_free_minor_tables(domain_state->minor_tables);
   domain_state->minor_tables = 0;
   caml_free_signal_stack();
+  
+  if(domain_state->current_stack != NULL) {
```
kayceesrk: lgtm

byterun/major_gc.c
```diff
@@ -177,6 +179,10 @@ static int no_orphaned_work ()
     orph_structs.final_info == NULL;
 }
 
+void caml_orphan_allocated_words() {
+    atomic_fetch_add(&terminated_domains_allocated_words, Caml_state->allocated_words);
```
kayceesrk: lgtm

byterun/shared_heap.c
```diff
@@ -593,7 +661,7 @@ struct mem_stats {
 static void verify_pool(pool* a, sizeclass sz, struct mem_stats* s) {
   value* v;
   for (v = a->next_obj; v; v = (value*)v[1]) {
-    Assert(*v == 0);
+    CAMLassert(*v == 0);
```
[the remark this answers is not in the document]
sadiqj: Heh, it looks like Assert is just #defined to be CAMLassert, so I'll change these back.

byterun/major_gc.c
```diff
@@ -333,7 +339,13 @@ static uintnat default_slice_budget() {
   uintnat heap_size = caml_heap_size(Caml_state->shared_heap);
   heap_words = (double)Wsize_bsize(heap_size);
   uintnat heap_blocks = caml_heap_blocks(Caml_state->shared_heap);
-  p = (double) Caml_state->allocated_words * 3.0 * (100 + caml_percent_free)
+
```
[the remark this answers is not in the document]
sadiqj: Hopefully this is now covered in the opening PR comment.

byterun/shared_heap.c
sadiqj: You're right, this is left over from some existing free space calculation code and we should only do it in DEBUG to verify the pool.

byterun/shared_heap.c
sadiqj: This should be addressed in the opening comment on the PR now.

byterun/shared_heap.c
sadiqj: Yes, I have now removed `free_pools` and `num_free_pools` entirely.

byterun/shared_heap.c
sadiqj: Looks like they're the same thing, will change back.

kayceesrk:
Lgtm. Thanks for the description.

PR #269, opened 2019-12-07
move from a global pools_to_rescan to a domain-local one

sadiqj:
The motivation behind this PR comes from experiencing a segfault during stress testing that was traced back to a pool being rescanned while another domain was allocating in to it. To address this the pools to rescan are now moved to a domain local list which ensures this case is no longer possible.

### DIFF
--- a/byterun/caml/domain_state.tbl
+++ b/byterun/caml/domain_state.tbl
@@ -80,6 +80,10 @@ DOMAIN_STATE(int, id)
 
 DOMAIN_STATE(int, unique_id)
 
+DOMAIN_STATE(struct pool**, pools_to_rescan)
+DOMAIN_STATE(int, pools_to_rescan_len)
+DOMAIN_STATE(int, pools_to_rescan_count)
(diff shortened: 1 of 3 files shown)


kayceesrk:
It would be useful to add a description of what the PR is intended to fix. I know that @sadiqj and I were looking at the issue together, but I am finding it hard to remind myself, given the time that has since passed, what the original issue this PR is fixing.

kayceesrk:
What happens to the pools_to_rescan when a domain terminates?

kayceesrk:
Thanks for adding the description. What about

> What happens to the pools_to_rescan when a domain terminates?

sadiqj:
It seems that we don't need to actually do anything about `pools_to_rescan` since `caml_empty_mark_stack` (via `mark`) actually ensures that those pools have been rescanned before the major gc is torn down on domain termination.

kayceesrk:
Ok. That sounds good to me.

PR #270, opened 2019-12-27
Fixes for DWARF CFI when switching between stacks

sadiqj:


asmcomp/amd64/emit.mlp
```diff
@@ -615,15 +615,21 @@ let emit_instr fallthrough i =
         end;
         cfi_remember_state ();
         I.mov (domain_field Domainstate.Domain_c_stack) rsp;
-        (* FIXME: CFI *)
-        (* cfi_def_cfa_offset 8; *)
+        cfi_def_cfa_offset 8;
```
kayceesrk: Why is the offset 8 here? Do you need this? 8 doesn't look right.

kayceesrk:
Can you explain what was the original problem and how the PR fixes it (for posterity)?

And, since we are pushing the OCaml stack pointer information into the C stack, is it necessary to still use `rbp` at all? Otherwise, can we use rbp, but use other DWARF info to avoid pushing into the stacl? I am not comfortable adding 4 memory operations for every fast external call in order to support debugging information. Perhaps this doesn't matter in practice?

mshinwell:
One general point that it's worth bearing in mind, apologies if this is known already: CFI information isn't just for debugging.  It is required to be correct by some language runtimes (Objective-C on macOS is one such) in order to unwind the stack for exception handling.

kayceesrk:
Agreed. I was a bit sloppy in my use of the term debugging.

asmcomp/amd64/emit.mlp
anmolsahoo25: Since here the address will be determined by the value in memory, I think we should load the value into a register use `cfi_def_cfa` to set stack pointer to the register value

anmolsahoo25:
The following is enough to generate the back-trace correctly - 

```
        I.mov rsp rbp;
        if Config.stats then begin
          I.inc (domain_field Domainstate.Domain_extcall_noalloc);
        end;
        I.mov (domain_field Domainstate.Domain_c_stack) rsp;
        cfi_def_cfa ();
        I.sub (int 8) rsp;
        cfi_def_cfa_offset 8;
        emit_call func;
        I.mov rbp rsp;
```

Tested it with an external call, seems to work. Also, can remove the space-time related stuff, because we are assuming this function does not alloc.

kayceesrk:
I.sub (int 8) rsp

Will make the stack misaligned. It will fail with functions that expect
16-byte alignment.

sadiqj:
This is superseded by the implementation in: https://github.com/ocaml-multicore/ocaml-multicore/pull/272

PR #272, opened 2020-01-13
Fix DWARF CFI for non-allocating external calls

kayceesrk:
Fixes the DWARF CFI for non-allocating external calls. Consider this program:

```ocaml
let foo () =
  classify_float 1.0;
  print_string "Hello, world!"

let _ = foo ()
```

In master, backtrace at the entry to `caml_classify_float_unboxed` used to produce a corrupted backtrace:

image

Now it produces the correct backtrace:
 
image

The position of frames `#1` and `#2` are swapped; Ideally, `<signal handler called>` should come before `camlTest__foo_1002`, to indicate the boundary between OCaml and C. But this is not important for correctness and will not be fixed. 

I have fixed the implementation in the `emit_profile` function as well, although `emit_profile` is not safe to be called in multicore since we may stack overflow in OCaml calling `mcount` directly. But that's a separate problem.

### DIFF
--- a/asmcomp/amd64/emit.mlp
+++ b/asmcomp/amd64/emit.mlp
@@ -615,14 +615,16 @@ let emit_instr fallthrough i =
         end;
         cfi_remember_state ();
         I.mov (domain_field Domainstate.Domain_c_stack) rsp;
-        (* FIXME: CFI *)
-        (* cfi_def_cfa_offset 8; *)
+        cfi_def_cfa_offset 0;
+
+        (* Keep GDB happy. If the cfa_offset was left at 0, then GDB complains
+         * that the frame did not save the PC and truncates the backtrace.
+         * Subtract rsp by 16 (instead of 8) to keep the stack aligned to
+         * 16-byte boundary. *)
+        I.sub (int 16) rsp;
+        cfi_adjust_cfa_offset 16;
+
         emit_call func;
-        (* FIXME KC:
-        if Config.spacetime then begin
-          record_frame Reg.Set.empty false i.dbg ~label:label_after
-        end
-        *)
         I.mov rbp rsp;
         cfi_restore_state ();
       end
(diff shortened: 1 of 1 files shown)


asmcomp/amd64/emit.mlp
```diff
@@ -615,14 +615,16 @@ let emit_instr fallthrough i =
...
+        I.sub (int 16) rsp;
+        cfi_adjust_cfa_offset 16;
+
         emit_call func;
-        (* FIXME KC:
-        if Config.spacetime then begin
-          record_frame Reg.Set.empty false i.dbg ~label:label_after
-        end
-        *)
         I.mov rbp rsp;
```
anmolsahoo25: We should free the stack we reserved, 

```suggestion
I.add (int 16) rsp;
cfi_adjust_cfa_offset -16;
I.mov rbp rsp;
```

asmcomp/amd64/emit.mlp
```diff
@@ -615,14 +615,16 @@ let emit_instr fallthrough i =
...
         cfi_restore_state ();
```
anmolsahoo25: Can remove this, 
```suggestion

```

asmcomp/amd64/emit.mlp
```diff
@@ -615,14 +615,16 @@ let emit_instr fallthrough i =
         end;
         cfi_remember_state ();
```
anmolsahoo25: Can remove this,
```suggestion
```

asmcomp/amd64/emit.mlp
kayceesrk: We don't need to since the next time we make a c call, we will use the `Domainstate.Domain_c_stack` value, which has not been updated.

asmcomp/amd64/emit.mlp
kayceesrk: Same reason as `cfi_remember_state`

asmcomp/amd64/emit.mlp
kayceesrk: We need it to restore the offsets to the last know good value after the call. It appears that it works fine in this specific example, but that might just be a coincidence. I would keep it around since any future change might make the code break.

asmcomp/amd64/emit.mlp
anmolsahoo25: Oh yeah. I agree, we allocated that space on the new C stack frame, which we are anyways throwing away.

asmcomp/amd64/emit.mlp
anmolsahoo25: Agree.

PR #276, opened 2020-01-30
Let only one domain scan the globals

kayceesrk:


kayceesrk:
@sadiqj you might want to cherry pick this one for stw parallel minor

PR #277, opened 2020-01-31
Give more space for mark stack.

kayceesrk:
Allow mark stack to be small initially and then grow until it is 1/32 of
the size of the shared heap on a domain. At that point, prune the mark
stack to limit the size of the growth.


kayceesrk:
I've added this branch to bench2.ocamllabs.io:8083. We should wait until the results are available and then merge.

kayceesrk:
The results are fine; there are a few outliers in micro-benchmarks.

PR #293, opened 2020-02-24
Mark only a little when pushing to mark stack

kayceesrk:


stedolan:
This patch removes an optimisation that I thought was useful at some point. (We've got more benchmark data now, so it's easier to tell if that was correct).

The idea is that `Some 42` doesn't need to push anything to the mark stack. So, before pushing a new block, `mark_stack_push` skips any integer fields at the start. If this results in the entire object being skipped, nothing is pushed.

I realise now that this has terrible latency if marking a long integer array. If desired, we could improve latency and keep the optimisation by replacing the `while (1)` with a bounded loop like `for (i = 0; i < 50; i++)` instead.

kayceesrk:
Good point. Given that most objects are small, something smaller than 50 should do. I'll make the edits now and push.

PR #297, opened 2020-02-28
Stats print heap_words and top_heap_words of the current domain if no major GC were done

kayceesrk:


kayceesrk:
This is mainly to get a fair comparison with stock OCaml behaviour

PR #299, opened 2020-02-28
Locking optimisations for I/O

stedolan:
- Do a trylock instead of a blocking section in With_mutex
  - Inline some lock-handling functions

This roughly doubles the throughput of this terrible program:
```
let () =
  for i = 0 to 1000_000_000 do
    print_char 'a'
  done
```


kayceesrk:
This looks good to me!

PR #305, opened 2020-02-29
More safepoints again

stedolan:
Safepoint insertion was generating bad frametables, by using liveness information from after regalloc had clobbered it. This patch moves the polling insertion to just after liveness computation, and restores #302


kayceesrk:
wonderful!

PR #316, opened 2020-03-13
Fix bug with Weak.blit

kayceesrk:
Multicore assumes that newly allocated ephemerons need not be marked or flushed in the cycle that they were allocated. The ephemerons themselves are allocated Marked. But can the keys or data be unmarked? Yes they can. `blit` operations copy weak references from one ephemeron to another without marking them. If the keys, for example, remain unmarked at the end of the marking phase, then the newly allocated ephemeron should sweep those keys to get rid of them in the same phase.

The fix is to mark the keys that are being blited. This will keep the unreachable keys alive for one more major cycle. Alternatively, we could have swept the newly allocated ephemerons in the cycle in which they were allocated. But this introduces a race between ephemeron sweeping and mutator allocating new ephemerons. This is not a hard problem but a tedious one to solve -- this makes `num_domains_to_sweep_ephe` not monotonically reduce to 0 and we need other mechanisms to get the synchronization right. 

If keeping the unreachable keys around for one more cycle causes performance problems, then we can revisit this issue.

### DIFF
--- a/byterun/caml/weak.h
+++ b/byterun/caml/weak.h
@@ -23,8 +23,13 @@
 extern value caml_ephe_none;
 
 struct caml_ephe_info {
-  value todo;
-  value live;
+  value todo; /* These are ephemerons which need to be marked and swept in the
+                 current cycle. If the ephemeron is alive, after marking, they
+                 go into the live list. */
+  value live; /* These are ephemerons which are alive in the current cycle.
+                 They still need to be swept to get rid of dead keys and
+                 values. Invariant: No keys are unreachable for these
+                 ephemerons. */
   uintnat cycle;
   struct {
     value* todop;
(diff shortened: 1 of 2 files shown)


kayceesrk:
@ctk21 you should merge this one onto parallel minor collector once the CI passes.

lpw25:
I think I know of at least one program where using `blit` to avoid marking the keys makes a noticeable difference to performance. I think it's fine not to fix this for now, but it would be good to address it eventually.

kayceesrk:
Noted. The right fix would be to keep the newly allocated ephemerons in a separate domain local list and sweep them at the end of the cycle. This makes domain termination detection only a little bit tricky since we already handle a similar problem with `Weak.get`.

PR #321, opened 2020-03-25
Remove locks in caml_shared_try_alloc critical path for all but large allocations

sadiqj:
The minor GC when promoting blocks to the major heap calls `caml_shared_try_alloc` which will try to find a space in a pool of the appropriate size for the blocks. If no pool can be found then `pool_find` is called which tries to grab one of the orphaned available pools or full pools of that size, if those aren't available then a new pool is allocated.

At the moment `pool_find` has a mutex around adopting orphaned pools and allocating a new pool. This lock ends up being contended reasonably heavily during minor collections especially with the parallel collector due to all minor GCs happening at the same time. We also do some slightly expensive bookkeeping while holding the lock which makes matters worse.

This PR removes locking from calls to `caml_shared_try_alloc` for all but large allocations (which are very infrequent). It does so by replacing the existing singly linked lists protected by a mutex with a set of lockfree stacks. The lockfree stack implementation is a simple Treiber Stack and we may want to look at something more performant later on. The implementation is kept separate from the shared heap code to allow for reuse elsewhere in the codebase.

### DIFF
--- a/byterun/shared_heap.c
+++ b/byterun/shared_heap.c
@@ -80,12 +80,9 @@ struct caml_heap_state* caml_init_shared_heap() {
   // the queues
   if( caml_domain_alone() ) {
     for (i = 0; i<NUM_SIZECLASSES; i++) {
-      caml_gc_log("initializing pools for sz: %d", i);
       lockfree_queue_init(&pool_freelist.global_avail_pools[i]);
       lockfree_queue_init(&pool_freelist.global_full_pools[i]);
     }
-
-    caml_gc_log("initializing free pools\n");
     lockfree_queue_init(&pool_freelist.free);
   }
 
--- a/byterun/shared_heap.c
+++ b/byterun/shared_heap.c
@@ -9,6 +9,7 @@
 #include "caml/fiber.h" /* for verification */
 #include "caml/gc.h"
 #include "caml/globroots.h"
+#include "caml/lockfree.h"
 #include "caml/memory.h"
 #include "caml/mlvalues.h"
 #include "caml/platform.h"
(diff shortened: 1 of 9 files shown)


byterun/shared_heap.c
```diff
@@ -75,6 +76,16 @@ struct caml_heap_state* caml_init_shared_heap() {
   int i;
   struct caml_heap_state* heap;
 
+  // This is the first time we've done this, so set up
+  // the queues
+  if( caml_domain_alone() ) {
```
kayceesrk: this can be turned into an `assert(caml_domain_alone())`?

kayceesrk (commented):
What is the performance impact of this change on the worst-performing benchmark? `test_decompress`?

kayceesrk:
We also need the same PR for concurrent minor, don't we?

kayceesrk:
I've run the performance results for this PR (stw+lf) vs parallel minor (stw). 

Speedup:

image

Nice improvements in game_of_life and decompress.

Max latency:

image

slowdown in decompress :-( I suspect that the contention in decompress is quite high and mutex degrades more gracefully lock-free data structure under heavy contention. 

99.9 percentile:

image

is the same. 

I've shared the full results by DM on slack.

ctk21:
We have updated our default parallel_minor_gc branch to be rebased to 4.10.
Apologies, an unintended consequence was that this PR got closed, I'm reopening it.
The PR may need rebasing.

kayceesrk:
Other than rebasing, what needs to happen to merge this PR in?

ctk21:
> Other than rebasing, what needs to happen to merge this PR in?

So from memory, I believe that we didn't merge this at the time because when benchmarking it turned out to have some strange regressions we didn't understand. 

I do wonder if this is the right next step for the allocator. The design direction of many other allocators (tcmalloc, jemalloc, etc) is to have some kind of CPU/domain local cache of free pools. We could try implementing that and then using the garbage collection cycles to balance the local cache of free pools back to the global list.

sadiqj:
I think it's worth rebasing this and testing it again - especially at very high levels of parallelism.

While state of the art allocators do maintain CPU/thread local free lists, we already have some of that in the per-domain pools that we allocate from - we only need to go to the global ones when we exhaust those.

That said, it's not clear if this is a higher priority than investigating multicore best fit.

kayceesrk:
Agree with @sadiqj. I would defer this until the allocator question is resolved.

PR #338, opened 2020-05-13
Introduce Lazy.try_force and Lazy.try_force_val

gadmm:
Hi @kayceesrk,

I though about it more and here is the fix I would like to propose for the `RacyLazy` exception.

From the commit log:

---

The `Lazy.RacyLazy` exception denotes a concurrency bug. As such, the following pattern:
```ocaml
  try Lazy.force x with Lazy.RacyLazy -> ...
```
does not do what one wants: we do not know whether it means that the evaluation of x ended with a bug, or if x itself is being concurrently forced.

The two new functions `Lazy.try_force` and `Lazy.try_force_val`, which return `None` if it is being concurrently forced, aims to replace the above pattern to implement concurrent lazy abstractions.

---

Here is the rationale:

- With @kayceesrk we agree that `Lazy.RacyLazy` can denote a bug.
- Catching the `RacyLazy` exception can otherwise be useful to implement concurrent lazy abstractions (e.g. lazy statics à la Rust); however doing so runs the risk of degrading an exception denoting a bug into a loop.
- Furthermore, as @kayceesrk explained, there is no obvious default behaviour replacing the RacyLazy exception for concurrent forcing of lazys in `Lazy.force`: implementing synchronisation at the level of domains by default might make sense in a vacuum (e.g. pay-as-you-go model à la C++/Rust: the user is free to not use it), but given the amount of legacy code that uses `Lazy.force`, this would threaten the cooperative concurrency story.

With this patch, concurrent lazy abstractions would call the `try_…` variants, while `RacyLazy` would always denote a bug.

Further notes:
- I needed to factor the code, but since the code looked optimised I tried to remain first-order.
- In the future, one can imagine that the `RacyLazy` exception could be eliminated via static control of interference.
- I saw interesting races for asynchronous exceptions in there... that is another story to look at in trunk for the future.

I believe this can also interest @stedolan.

### DIFF
--- a/stdlib/lazy.mli
+++ b/stdlib/lazy.mli
@@ -43,7 +43,8 @@ type 'a t = 'a CamlinternalLazy.t
    [Lazy.t] instead.
 
    Note: [Lazy.force] is not thread-safe.  If you use this module in
-   a multi-threaded program, you will need to add some locks.
+   a multi-threaded program, you will need to add some locks.  See also
+   [Lazy.try_force].
 
    Note: if the program is compiled with the [-rectypes] option,
    ill-founded recursive definitions of the form [let rec x = lazy x]
@@ -57,8 +58,6 @@ type 'a t = 'a CamlinternalLazy.t
 
 exception Undefined
 
-exception RacyLazy
-
 (* val force : 'a t -> 'a  *)
 external force : 'a t -> 'a = "%lazy_force"
 (** [force x] forces the suspension [x] and returns its result.
@@ -66,20 +65,28 @@ external force : 'a t -> 'a = "%lazy_force"
     same value again without recomputing it.  If it raised an exception,
     the same exception is raised again.
     Raise {!Undefined} if the forcing of [x] tries to force [x] itself
-    recursively.
-    Raise {!RacyLazy} if [x] is concurrently forced by another domain.
+    recursively, or if [x] is concurrently forced by another domain.
 *)
+
+val try_force : 'a t -> 'a option
+(** [try_force x] behaves similarly to [Some (force x)], except that
+    it returns immediately with [None] if [x] is concurrently forced by
+    another domain. *)
 
 val force_val : 'a t -> 'a
 (** [force_val x] forces the suspension [x] and returns its
     result.  If [x] has already been forced, [force_val x]
     returns the same value again without recomputing it.
     Raise {!Undefined} if the forcing of [x] tries to force [x] itself
-    recursively.
-    Raise {!RacyLazy} if [x] is concurrently forced by another domain.
+    recursively, or if [x] is concurrently forced by another domain.
     If the computation of [x] raises an exception, it is unspecified
     whether [force_val x] raises the same exception or {!Undefined}.
 *)
+
+val try_force_val : 'a t -> 'a option
+(** [try_force_val x] behaves similarly to [Some (force_val x)],
+    except that it returns immediately with [None] if [x] is
+    concurrently forced by another domain. *)
 
 val from_fun : (unit -> 'a) -> 'a t
 (** [from_fun f] is the same as [lazy (f ())] but slightly more efficient.
--- a/stdlib/lazy.mli
+++ b/stdlib/lazy.mli
@@ -42,9 +42,11 @@ type 'a t = 'a CamlinternalLazy.t
    for the [lazy] keyword.  You should not use it directly.  Always use
    [Lazy.t] instead.
 
-   Note: [Lazy.force] is not thread-safe.  If you use this module in
-   a multi-threaded program, you will need to add some locks.  See also
-   [Lazy.try_force].
+   Note: [Lazy.force] (and therefore the [lazy] pattern-matching) are
+   thread-safe, but will raise the [Undefined] exception if forced
+   concurrently. If you need to share a lazy between threads, then you
+   need to use [Lazy.try_force] and implement your own synchronisation.
+   (@since XXX)
 
    Note: if the program is compiled with the [-rectypes] option,
    ill-founded recursive definitions of the form [let rec x = lazy x]
@@ -70,8 +72,8 @@ external force : 'a t -> 'a = "%lazy_force"
 
 val try_force : 'a t -> 'a option
 (** [try_force x] behaves similarly to [Some (force x)], except that
-    it returns immediately with [None] if [x] is concurrently forced by
-    another domain. *)
+    it returns immediately with [None] if [x] is already being forced
+    concurrently by another domain. *)
 
 val force_val : 'a t -> 'a
 (** [force_val x] forces the suspension [x] and returns its
@@ -80,13 +82,12 @@ val force_val : 'a t -> 'a
     Raise {!Undefined} if the forcing of [x] tries to force [x] itself
     recursively, or if [x] is concurrently forced by another domain.
     If the computation of [x] raises an exception, it is unspecified
-    whether [force_val x] raises the same exception or {!Undefined}.
-*)
+    whether [force_val x] raises the same exception or {!Undefined}. *)
 
 val try_force_val : 'a t -> 'a option
 (** [try_force_val x] behaves similarly to [Some (force_val x)],
-    except that it returns immediately with [None] if [x] is
-    concurrently forced by another domain. *)
+    except that it returns immediately with [None] if [x] is already
+    being forced concurrently by another domain. *)
 
 val from_fun : (unit -> 'a) -> 'a t
 (** [from_fun f] is the same as [lazy (f ())] but slightly more efficient.
--- a/stdlib/camlinternalLazy.ml
+++ b/stdlib/camlinternalLazy.ml
@@ -18,7 +18,6 @@
 type 'a t = 'a lazy_t
 
 exception Undefined
-exception RacyLazy
 
 (* [update_tag blk old new] updates the tag [blk] from [old] to [new] using a
  * CAS loop (in order to handle concurrent conflicts with the GC marking).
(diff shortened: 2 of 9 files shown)


kayceesrk (changes requested):
Hi @gadmm! Thanks for the PR. The proposed functions are a useful addition and is what the user will want. Happy to merge this. I'd like to see two small changes:

1. Fix the `backtrace2.ml`. The line number in the backtrace will be different since the code has been refactored. You just need to copy the resulting backtrace as the reference. This should fix the CI.
2. It would also be useful to add a test case in the same vein as the one testing `RacyLazy` exception.

gadmm:
I am wondering, is it possible, in a cooperative concurrency scenario, that the `Lazy.Undefined` exception for recursive forcing might denote concurrent forcing, and that the user might try to catch it to implement synchronisation in that case? If so, it is probably better to have the try_ functions return it as well instead of raising.

Depending on the above, a further simplification is to conflate the two exceptions (both of which denote bugs, then). What do you think about it?

kayceesrk:
> I am wondering, is it possible, in a cooperative concurrency scenario, that the Lazy.Undefined exception for recursive forcing might denote concurrent forcing, and that the user might try to catch it to implement synchronization in that case? If so, it is probably better to have the try_ functions return it as well instead of raising.

I am not sure what useful code would do this since one cannot distinguish recursive forcing from forcing by another concurrent fiber. IINM, the code needs to be sure that the lazy under question cannot be recursively forced. Am I right?

> Depending on the above, a further simplification is to conflate the two exceptions (both of which denote bugs, then). What do you think about it?

What would this look like? `Undefined` raised in the case of both recursive and concurrent forcing, and any code that is intending to use lazy between two domains that may race to force the lazy **must** use `try_force*`? If so, this sounds reasonable to me. As long as we have the ability to identify lazy being forced in parallel, I'm happy with the simplifications. Do I understand you correctly?

ctk21:
We have updated our default `parallel_minor_gc` branch to be rebased to 4.10.
Apologies, an unintended consequence was that this PR got closed, I'm reopening it. 
The PR may need rebasing.

gadmm:
> > I am wondering, is it possible, in a cooperative concurrency scenario, that the Lazy.Undefined exception for recursive forcing might denote concurrent forcing, and that the user might try to catch it to implement synchronization in that case? If so, it is probably better to have the try_ functions return it as well instead of raising.
> 
> I am not sure what useful code would do this since one cannot distinguish recursive forcing from forcing by another concurrent fiber. IINM, the code needs to be sure that the lazy under question cannot be recursively forced. Am I right?

I thought about it more. In the end the interface where try_force returned either condition as an error was heavier and I was not sure it would be useful. I believe the user needs to implement their own synchronisation mechanism anyway, making a check for Lazy.Undefined redundant.

> 
> > Depending on the above, a further simplification is to conflate the two exceptions (both of which denote bugs, then). What do you think about it?
> 
> What would this look like? `Undefined` raised in the case of both recursive and concurrent forcing, and any code that is intending to use lazy between two domains that may race to force the lazy **must** use `try_force*`? If so, this sounds reasonable to me. As long as we have the ability to identify lazy being forced in parallel, I'm happy with the simplifications. Do I understand you correctly?

Yes, and I thought of another argument: with systhreads, a racy lazy would raise Undefined, so this preserves the meaning of Undefined. So I removed the RacyLazy exception and updated the documentation.

stdlib/lazy.mli
```diff
@@ -57,30 +58,36 @@ type 'a t = 'a CamlinternalLazy.t
...
     the same exception is raised again.
     Raise {!Undefined} if the forcing of [x] tries to force [x] itself
-    recursively.
-    Raise {!RacyLazy} if [x] is concurrently forced by another domain.
+    recursively, or if [x] is concurrently forced by another domain.
 *)
 
+val try_force : 'a t -> 'a option
+(** [try_force x] behaves similarly to [Some (force x)], except that
+    it returns immediately with [None] if [x] is concurrently forced by
```
kayceesrk, line 73:
```suggestion
    it returns immediately with [None] if [x] is already being forced concurrently and by
```

This distinction is important since the first `force` does not return `None`, only the subsequent concurrent forcing return `None`.

stdlib/lazy.mli
```diff
@@ -43,7 +43,8 @@ type 'a t = 'a CamlinternalLazy.t
    [Lazy.t] instead.
 
    Note: [Lazy.force] is not thread-safe.  If you use this module in
```
kayceesrk, line 45: I'm not sure if I agree with the use of "not thread-safe" here. The definition of thread-safety is vague, but I tend to view it as not having a defined semantics. Whereas here, we do have a defined semantics for concurrent forcing. If two domains concurrently force lazy, the program would not crash (quite important this one) and you'll get `Undefined` exception. Hence, I would not mark this API as not thread safe, but still point to `try_force`.

stdlib/lazy.mli
gadmm: ok

stdlib/lazy.mli
gadmm: It is true that your work has made the API thread-safe, so I will mark this in the documentation. IIUC the trunk implementation, and assuming the native behaviour for polling, there is a race where the lazy_t value could end up with a forward tag, but the first cell concurrently replaced with a closure raising an exception, breaking type safety. So it seems correct to me that previously, thread-unsafe could mean "could crash".

stdlib/lazy.mli
```diff
@@ -42,9 +42,11 @@ type 'a t = 'a CamlinternalLazy.t
    for the [lazy] keyword.  You should not use it directly.  Always use
    [Lazy.t] instead.
 
-   Note: [Lazy.force] is not thread-safe.  If you use this module in
-   a multi-threaded program, you will need to add some locks.  See also
-   [Lazy.try_force].
+   Note: [Lazy.force] (and therefore the [lazy] pattern-matching) are
+   thread-safe, but will raise the [Undefined] exception if forced
+   concurrently. If you need to share a lazy between threads, then you
+   need to use [Lazy.try_force] and implement your own synchronisation.
```
gadmm, line 48: I was not sure about the terminology: "implement your own synchronisation" -> "implement your own locking mechanism"? What do you think?

gadmm:
I do not understand the failure with clang
```
Error: Files primitives.cmo and ../otherlibs/unix/unix.cma(Unix)
       make inconsistent assumptions over interface Unix
```

ctk21:
> I do not understand the failure with clang
> 
> ```
> Error: Files primitives.cmo and ../otherlibs/unix/unix.cma(Unix)
>        make inconsistent assumptions over interface Unix
> ```

I've tried to reproduce on a local OS X box with clang but can not. 

There's a couple of things I'm going to try digging into, unless
others have seen this before?

kayceesrk:
@gadmm Have you tried the obvious? `make clean` and then `make world.opt`? I've seen this sort of error occur in dirty builds.

stdlib/lazy.mli
kayceesrk: Agree with your assessment reg trunk.

gadmm:
The bug is with the CI, not my computer

stdlib/lazy.mli
kayceesrk: `synchronization` the right word. Perhaps we could add the example of:

```ocaml
let rec conc_force l =
  match Lazy.try_force with
  | None -> Domain.Sync.cpu_relax(); conc_force l
  | Some v -> v
```

in the documentation as an example of synchronization where the concurrent access spin-waits until completion.

kayceesrk:
Apologies. I read that message fast and didn't notice that it was a CI failure. I've kicked a rebuild again to see if that turns green. I've seen issues with OSX builds previously.

gadmm:
No problem.
This happened after rebasing on top of 4.10 so this could also be a problem in the fresh multicore rebasing.
I tried to `make alldepend` (on a clean directory), I saw a huge change, so I pushed it in a separate commit. Let's see how it goes.

stdlib/lazy.mli
gadmm: I am wondering whether this could be offered in a library. It seems fine to offer it if the user has to make a conscious decision to use it.

kayceesrk:
The rebuild that I kicked off https://travis-ci.org/github/ocaml-multicore/ocaml-multicore/jobs/687442919 has gone past the point where the failure happened previously.

gadmm:
Ok but it is also good to find the root cause. Would you like that I keep the make alldepend?

kayceesrk:
Let's keep `make alldepend`.

ocamltest/.depend
```diff
@@ -1,25 +1,27 @@
...
-  ../runtime/caml/osdeps.h
+ ../runtime/caml/config.h ../runtime/caml/m.h ../runtime/caml/s.h \
+ ../runtime/caml/mlvalues.h ../runtime/caml/config.h \
+ ../runtime/caml/misc.h ../runtime/caml/camlatomic.h \
+ ../runtime/caml/domain_state.h ../runtime/caml/domain_state.tbl \
+ ../runtime/caml/byte_domain_state.tbl ../runtime/caml/memory.h \
+ ../runtime/caml/gc.h ../runtime/caml/mlvalues.h \
+ ../runtime/caml/major_gc.h ../runtime/caml/minor_gc.h \
+ ../runtime/caml/addrmap.h ../runtime/caml/domain.h \
+ ../runtime/caml/memory.h ../runtime/caml/alloc.h ../runtime/caml/alloc.h \
```
ctk21, line 23: there is something odd going on here, some of these dependencies are duplicated, e.g. `../runtime/caml/alloc.h` and `../runtime/caml/mlvalues.h`

gadmm: Ok. This is beyond my knowledge. I propose to remove this commit so that you can look at it separately.

stdlib/lazy.mli
kayceesrk: Ok. That's fine. Let's mark this one as resolved.

kayceesrk (approved):

ocamltest/.depend
kayceesrk: Odd to see duplications, but are sure we didn't have duplicates before?

kayceesrk:
I had another thought about `Undefined` exception for both the racy and recursive cases. Consider the example below:

```ocaml
let l1 = lazy ...
let l2 = lazy (Lazy.force l1)
let domain1 = Domain.spawn (fun _ -> ignore (Lazy.force l1))
let domain2 = Domain.spawn (fun _ -> ignore (Lazy.force l2))
```
It is possible that both the domains concurrently force `l1`. If `domain1` was the first one to force `l1`, then `domain2` will get an `Undefined` exception. The lazy value `l2` will be updated such that subsequent invocations of `l2` would cause undefined exception. The original deferred computation in `l2` is lost.

It might be better to actually distinguish the recursive and concurrent forcing with two different exceptions `Undefined` and `RacyLazy` with the added semantics that if the result of a deferred computation is `RacyLazy` then the lazy value is left with the original computation. The idea is that if `l2` was forced again, there is a chance that the `domain1` has completed forcing `l1`, and hence, forcing `l2` will successfully complete. 

Of course, this is dealing with the incorrect use of the API since `try_force` is available, and that is the one to be used. The current implementation is fine as it is. I'm recording this one so that we may revisit this when it is time for upstreaming.

gadmm:
> It might be better to actually distinguish the recursive and concurrent forcing with two different exceptions `Undefined` and `RacyLazy` with the added semantics that if the result of a deferred computation is `RacyLazy` then the lazy value is left with the original computation. The idea is that if `l2` was forced again, there is a chance that the `domain1` has completed forcing `l1`, and hence, forcing `l2` will successfully complete.

Sounds good.

> Of course, this is dealing with the incorrect use of the API since `try_force` is available, and that is the one to be used.

This should not be a problem.

kayceesrk:
@ctk21 how do we proceed with the duplicates in .depend? Clearly, it is not an issue with the changes here and the CI passes. I'm tempted to merge this one and debug the duplicate entries in a separate issue.

ctk21:
The .depend changes were reverted. I think we merge the PR as is, everything passes. 

In the future the C dependencies will not even be a part of the repo:
  https://github.com/ocaml/ocaml/pull/9332

kayceesrk:
Ack. Merging. Thanks @gadmm!

Issue #257, opened 2019-07-16
Assertion failed in interp.c line 317


gmevel:
* architecture = x86_64
  * operating system = Linux 5.1.5-arch1-2-ARCH
  * compiler = Multicore OCaml 4.06.1, dev version (commit 8fbacb11c)

The following (non-sensical, as this is a reduced example) code:

```ocaml
type node = Nil | Cons of stack
and stack = node Atomic.t

let make () : stack =
  Atomic.make Nil

let push (st : stack) : unit =
  let next = Atomic.make Nil in
  let new_hd = Cons next in
  let old_hd = Atomic.exchange st new_hd in
  Atomic.set next old_hd

let pop (st : stack) : unit =
  begin match Atomic.get st with
  | Nil       -> ()
  | Cons next -> Atomic.set st (Atomic.get next)
  end

let run n thread =
  Array.init n (fun _ -> Domain.spawn thread)
  |> Array.iter Domain.join

let () =
  let st = make () in
  run 2 begin fun () ->
    for _ = 1 to 10_000 do
      push st (* ← assertion failure at some point here *)
    done
  end ;
  run 1 begin fun () ->
    for _ = 1 to 10_000 do
      pop st (* ← segmentation fault at some point here *)
    done
  end ;
```

when compiled and run in debug mode:

```sh
ocamlc.opt unix.cma test.ml -o test.byte
ocamlrund ./test.byte
```

gives me:

```
### OCaml runtime: debug mode ###
[02] file interp.c; line 317 ### Assertion failed: !Is_foreign(accu)
Abandon (core dumped)
```

We observe an assertion failure in the `push` loop. In normal mode (`ocamlrun` instead of `ocamlrund`, it causes the program to segfault (or, sometimes, to hang) later, in the `pop` loop. Try playing with the constants to observe this behaviour more often.

ctk21:
Thanks for reporting. This is now fixed for branch `master` in #340 .

PR #340, opened 2020-05-15
Fix Atomic.exchange in concurrent_minor_gc 

ctk21:
Issue #257 gives an example using `Atomic.exchange` which causes crashes with the concurrent minor gc (`master` branch). 

The bug we have is that `caml_atomic_exchange` in `memory.c`:
```
CAMLprim value caml_atomic_exchange (value ref, value v)
{
  value ret;
  if (Is_young(ref) || caml_domain_alone()) {
    ret = Op_val(ref)[0];
    Op_val(ref)[0] = v;
  } else {
    /* See Note [MM] above */
    atomic_thread_fence(memory_order_acquire);
    ret = atomic_exchange(Op_atomic_val(ref), v);
  }
  write_barrier(ref, 0, ret, v);
  return ret;
}
```
is allowing a reference to a foreign minor heap to escape through `ret`. I couldn't figure a working way to insert a `caml_read_barrier` to fix the issue (you can't use `ref` as a location argument because it has been exchanged). 

I took an alternative approach that implements `Atomic.exchange` using `Atomic.get` (which induces the correct read barrier) and `Atomic.compare_and_set` to get the correct exchange semantics. It's possibly not efficient, but I think it is correct and the sample code in #257 now works without crashing. 

I'm open to better fixes or suggestions, but right now this is the best fix I could come up with.

### DIFF
--- a/stdlib/atomic.ml
+++ b/stdlib/atomic.ml
@@ -2,12 +2,22 @@ type 'a t
 
 external make : 'a -> 'a t = "%makemutable"
 external get : 'a t -> 'a = "%atomic_load"
-external exchange : 'a t -> 'a -> 'a = "%atomic_exchange"
+(* currently can not exchange if the contents are a block that could be in a foreign minor heap *)
+external exchange_unsafe : 'a t -> 'a -> 'a = "%atomic_exchange"
 external compare_and_set : 'a t -> 'a -> 'a -> bool = "%atomic_cas"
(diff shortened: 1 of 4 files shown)


stedolan:
Thinking about this a little, I can't see any better way of doing it. The read barrier needs the object being promoted to be reachable from the shared heap, but if you've just swapped it out it might not be. This version looks good.

kayceesrk:
Lgtm.

Issue #344, opened 2020-05-25
Backport Unix.system without Unix.fork from upstream to multicore


kayceesrk:
Initially, we had #241 which implements `Unix.system` without using `Unix.fork`. This was later implemented upstream by ocaml/ocaml#9573. We need to backport ocaml/ocaml#9573 to multicore.

PR #346, opened 2020-05-27
Backport Unix system without fork to parallel_minor_gc

abbysmal:
I cherry-picked the commits, compiled and ran the unix testsuite, it seems fine so far.
This is worth combining with #342 to test changes further in Multicore's context.
See #344

### DIFF
--- a/configure
+++ b/configure
@@ -15270,6 +15270,24 @@ if test "x$ac_cv_func_execvpe" = xyes; then :
 fi
 
 
+## posix_spawn
+
+ac_fn_c_check_header_mongrel "$LINENO" "spawn.h" "ac_cv_header_spawn_h" "$ac_includes_default"
+if test "x$ac_cv_header_spawn_h" = xyes; then :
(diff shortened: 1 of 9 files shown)


ctk21:
It looks good to me. 

One comment is that we will have some bits of `otherlibs/unix` where the concurrent minor gc macros are used (e.g. `Int_field(error, 0)` rather than `Int_val(Field(error, 0))`). Depending on how people feel about consistency, we could:
- take this as is and done. 
- take this as is then refactor out the concurrent minor gc macro usage in the other files as further work. 
- convert this to use the concurrent minor gc macros.

kayceesrk:
I suspect that concurrent gc macros are not going to land upstream anytime soon. There was an idea of taking the better, non-conflicting parts of the concurrent GC api and upstreaming it. But there is less motivation to do it now as parallel minor will land first. 

My aim is to keep the code as close as possible to upstream OCaml so that when the time comes to carve out parts of the multicore to make PRs, we don't end up having to more API refactoring than necessary. With that in my, my suggestion is to take this as is and call it done.

PR #347, opened 2020-06-01
Turn on -Werror by default

dra27:
First two commits are ocaml/ocaml#9625. This extends the logic already in `configure` which turns on `-Werror` if `VERSION` contains `+dev` to also recognise `+multicore`. The new `--disable-warn-error` option can then be used in multicore-opam.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -536,18 +536,18 @@ AS_CASE([$ocaml_cv_cc_vendor],
   [xlc-*],
     [outputobj='-o $(EMPTY)'
     warn_error_flag=''
-    gcc_warnings='-qflag=i:i'], # all warnings enabled
+    cc_warnings='-qflag=i:i'], # all warnings enabled
   [msvc-*],
     [outputobj='-Fo'
     warn_error_flag='-WX'
-    gcc_warnings=''],
+    cc_warnings=''],
   [outputobj='-o $(EMPTY)'
   warn_error_flag='-Werror'
-  gcc_warnings='-Wall -Wdeclaration-after-statement'])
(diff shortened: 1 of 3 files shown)


dra27:
cc @ctk21

configure.ac
```diff
@@ -530,15 +534,21 @@ AS_IF(
...
+    [outputobj='-Fo'
+    warn_error_flag='-WX'
+    cc_warnings=''],
+  [outputobj='-o $(EMPTY)'
+  # FIXME Should also be -Wdeclaration-after-statement here
+  warn_error_flag='-Werror'
+  cc_warnings='-Wall'])
+
+AS_CASE([$enable_warn_error,AC_PACKAGE_VERSION],
+  [yes,*|,*+dev*|*+multicore*],
```
ctk21: I might be getting the wrong end of the stick, but if someone does `--disable-warn-error` explicitly with a `+multicore` version. Will this still force a warn-error build?

That is should it be `,*+multicore*` or `*+multicore*` in the condition

dra27: You have completely the right end of the stick - there's a comma missing 😊

dra27:
The upstream PR in OCaml has been merged, so `--disable-warn-error` is now an upstream-compatible option. I've rebased this - it'd be handy for it to go in so I can remove it from the Windows PR.

ctk21:
I'm keen to get the C warnings enabled as errors and think we should merge.
It is a good idea, matches how trunk develops its C code and makes any future rebasing easier.

sadiqj (approved):

PR #351, opened 2020-06-02
mingw-w64 Port

dra27:
This is a rebased version of #173, extended with patches needed for the native code compiler.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Run the testsuite
         if: ${{ matrix.name != 'linux-O0' }}
         run: |
-          bash -xe tools/ci/actions/runner.sh test
+          bash -c 'SHOW_TIMINGS=1 tools/ci/actions/runner.sh test'
       - name: Run the testsuite (linux-O0, parallel)
         if: ${{ matrix.name == 'linux-O0' }}
         env:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,7 +31,7 @@ environment:
     FORCE_CYGWIN_UPGRADE: 0
     BUILD_MODE: world.opt
   matrix:
-    - PORT: mingw32
+    - PORT: mingw64
       BOOTSTRAP_FLEXDLL: true
 # OCaml 5.00 does not yet support MSVC
 #    - PORT: msvc64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,14 +33,15 @@ environment:
   matrix:
     - PORT: mingw32
       BOOTSTRAP_FLEXDLL: true
-    - PORT: msvc64
-      BOOTSTRAP_FLEXDLL: false
-      BUILD_MODE: steps
-    - PORT: msvc32
-      BOOTSTRAP_FLEXDLL: false
-      BUILD_MODE: C
-      SDK: |-
-        "C:\Program Files\Microsoft SDKs\Windows\v7.1\Bin\SetEnv.cmd" /x86
+# OCaml 5.00 does not yet support MSVC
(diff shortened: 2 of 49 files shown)


dra27:
~This should not be merged yet (just in case that wasn't already clear!) - forced to open a new PR after a distracted-by-noise incorrect git-push command!~

dra27:
This is ready to merge from my perspective. As it stands, `weak-ephe-final/ephetest_par.ml` and `weak-ephe-final/weaklifetime_par.ml` fail in native code, but we knew about that (see #352) and those tests ~magically~correctly become fixed when #573 is merged.

Some of the changes in `platform.c` related to the concurrent minor GC and I think may be no longer required - however, it'd be nice to merge them and have them in the history when the rest of it gets removed (they did work back in 2018 when I first did the port!)

The assembly object in the runtime is now built for each runtime individually (as all the other objects apart from `prims.o` are) which allows the use of `DEBUG` in the assembler functions (not strictly Windows-related).

The testsuite is much slower on Windows, so I really miss `make -C testsuite parallel`. Consequently I have added `make -C testsuite parallel-enabled` which does the same as `all-enabled`, but via GNU parallel (also not strictly Windows-related).

Some items which remain "TODO" (and should ultimately be tracked on a post-MVP roadmap somewhere):
- `print_trace` has moved from `misc.c` to the platform-specific modules (and consequently been renamed to `caml_print_trace`), but the mechanism around this is suspect on all platforms (especially its guaranteed inclusion in native code). For Windows, it's possible that `-lDbgHelp` can be omitted for non-debug builds, although this may end up being too confusing, as that means you'd need additional flags to use with `-runtime-variant` which is not at present supported. Implementing it has re-piqued my interest in getting proper .pdb files for both mingw-64 and MSVC!
- The uses of winpthreads in the runtime need converting to direct calls to the Windows API. This will be part of re-enabling the MSVC port, and shouldn't be too hard.
- `st_win32.h` needs to be fully restored in systhreads which will remove the dependency on winpthreads there too and, likewise, be part of re-enabling the MSVC port. This shouldn't be too hard either, as the code required is either already in trunk's `st_win32.h` or it's demonstrably working on mingw-w64 at the moment in `st_posix.h` and just needs translating. There are a couple of scandalous `#ifdef _WIN32`s in `st_posix.h` which would then be removed!

dra27:
This preserves the AppVeyor configuration but, for whatever reasons which we still need to look into properly, the run takes longer than an hour on AppVeyor's runners.

dra27:
This restores the build for mingw-w64: AppVeyor build log from my fork [here](https://ci.appveyor.com/project/dra27/ocaml-multicore/builds/41989961).

The testsuite takes too long to run in AppVeyor at the moment - there are a very small number of tests failing which need looking into, but these can be post-merge tasks... the useful thing is that it means ocaml/ocaml Windows CI will continue to report breakages to the build straightaway.

dra27:
`tests/lib-threads/prodcons2.ml` and `tests/backtrace/backtrace_systhreads.ml` seem to be failing sporadically when the machine is under load (e.g. with a parallel testsuite run) but not when the tests are run individually.

`tests/lib-dynlink-native/main.ml` has something fundamentally wrong here (and new - this may be worth bisecting). Consistently encountering `Fatal error: exception - Invalid_argument("output_value: abstract value (outside heap)")`

There is a general performance regression for Windows, given how much longer it's taking to run the testsuite, but this needs some dedicated time for pinning down - hopefully it will turn out there's something laughably daft in one of the patches (the build time hasn't changed much, for example).

kayceesrk:
Thanks @dra27. I agree that having a working mingw-w64 build is useful.

We should continue working on the regression and failing tests post-merge.

PR #353, opened 2020-06-05
Poll for interrupts in cpu_relax without locking

### LINKED ISSUE #348: Domain.Sync.cpu_relax should poll for interrupts without locking
`Domain.Sync.cpu_relax` will call the C function `caml_ml_domain_cpu_relax`. This function checks for incoming interrupts before calling `cpu_relax`. This is done by obtaining and releasing the domain lock. 

https://github.com/ocaml-multicore/ocaml-multicore/blob/172fe18bbc7fcdaa1fc286869c28a2c59ecbf27a/runtime/domain.c#L1298-L1300

This will interfere with the domain that is attempting to interrupt this domain. It would be better to poll for interrupt without the lock and if there happens to be an interrupt waiting, then we handle the interrupt with lock. 

In the polling version, we may need to load `s->qhead` using c11 atomics? 

https://github.com/ocaml-multicore/ocaml-multicore/blob/172fe18bbc7fcdaa1fc286869c28a2c59ecbf27a/runtime/domain.c#L944

What is the interaction between a store in a critical section protected by a lock and a load? Which version of `atomic_load` do I want? 

https://github.com/ocaml-multicore/ocaml-multicore/blob/172fe18bbc7fcdaa1fc286869c28a2c59ecbf27a/runtime/caml/camlatomic.h#L37-L38

kayceesrk:
Fixes #348.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1293,11 +1293,13 @@ CAMLprim value caml_ml_domain_yield_until(value t)
 
 CAMLprim value caml_ml_domain_cpu_relax(value t)
 {
-  int interrupts;
+  uintnat interrupts = 0;
   struct interruptor* s = &domain_self->interruptor;
-  caml_plat_lock(&s->lock);
-  interrupts = handle_incoming(s);
-  caml_plat_unlock(&s->lock);
+  if (Caml_check_gc_interrupt(Caml_state)) {
+    caml_plat_lock(&s->lock);
+    interrupts = handle_incoming(s);
+    caml_plat_unlock(&s->lock);
+  }
   if (!interrupts) cpu_relax();
   return Val_unit;
 }


ctk21 (commented):
LGTM

Might not matter in the big picture, but I wonder if the most lazy in CPU relaxation that touches only to check for interrupts is:
```C
  if (!Caml_check_gc_interrupt(Caml_state)) {
    cpu_relax();
  } else {
    int interrupts;
    struct interruptor* s = &domain_self->interruptor;
    caml_plat_lock(&s->lock);
    interrupts = handle_incoming(s);
    caml_plat_unlock(&s->lock);
    if (!interrupts) cpu_relax();
  }
  return Val_unit;
```

kayceesrk:
I suspect that the optimizer will already do something close to the code that you've presented. I'll leave the code as-is. Thanks for reviewing it!

PR #355, opened 2020-06-10
One more location to poll for interrupts without lock

kayceesrk:
Similar to #353.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1110,11 +1110,12 @@ static void caml_wait_interrupt_completed(struct interruptor* self, struct inter
   }
 
   while (!atomic_load_acq(&req->completed)) {
-    cpu_relax();
     if (Caml_check_gc_interrupt(Caml_state)) {
       caml_plat_lock(&self->lock);
       handle_incoming(self);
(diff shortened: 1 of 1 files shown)


stedolan:
If we're optimising this, the `cpu_relax` should probably be in an else block, rather than unconditional. We don't want to `PAUSE` if there's an interrupt pending.

kayceesrk:
Good point. I've moved `cpu_relax` to the else branch.

ctk21:
The functionality looks correct to me. 

However I wonder if we take the opportunity to clean things up further, although the code looks different in `caml_ml_domain_cpu_relax` both are doing the same thing (at least conceptually). Might it be better to have a single function that does the fast-path check, locking and interrupt handling which is called from both (e.g. an inline function `domain_cpu_relax`)? Or use `caml_ml_domain_cpu_relax` from here?

kayceesrk:
Good point. Let me cleanup the code.

kayceesrk:
@ctk21 I've factored out common code.

ctk21:
LGTM

Issue #90, opened 2017-01-19
Implement backup threads for domains


kayceesrk:
**tl;dr** Instantiate multiple system threads to service a single domain

In the current design, every domain is associated with a single system thread that services the domain. Whenever a domain blocks on a external call, the lock on the domain is released so that another domain may temporarily take over the domain to:

- perform major GC on behalf of the blocked thread
- service promotion requests

The fact that a single system thread might be responsible for managing multiple domains at the same time is tricky and leads to subtle bugs. 

The proposal is to create 2 system threads for each domain, one which is the primary mutator thread and runs OCaml code and another is a backup thread which only steps in when the primary thread is blocked on an external call. By default, the backup thread sleeps on a condition variable protected by the domain lock and is woken up to service GC or promotion requests.

kayceesrk:
Should be done after #146.

kayceesrk:
See #356

ctk21:
This issue is stale and backup threads got merged in #356.

PR #356, opened 2020-06-11
Backup threads for domains.

kayceesrk:
This PR adds backup threads for domains that service interrupts when the domain is blocked in the kernel. The PR is a little bit subtle, hence the commentary below. I would recommend reading the code and referring back to the commentary when things are not clear.

## Backup threads

Backup threads do not run OCaml code. They only perform GC on behalf of the main thread. A backup thread refers to the same thread-local state (`Caml_state` and `domain_self`) as the main thread.

Backup threads are only created if the program spawns a domain; single-threaded programs do not need backup threads. This is aimed at keeping the single-threaded overheads low. A backup thread is installed by a domain by calling `install_backup_thread` function. The backup thread logic itself is in `backup_threda_func`. 

## Backup thread waiting while the main thread is running

A `domain_lock`  (previously known as `roots_lock`) is a per-domain lock that represents the capability to perform GC on that domain. The `domain_lock` is held by the main thread executing OCaml code.  

When a backup thread is created, it first tries to obtain the `domain_lock`. When the main thread enters the blocking section, the backup thread obtains the `domain_lock`. 

When the main thread aims to leave the blocking section and enter OCaml code (represented by `backup_thread_msg == MSG_ENTERING_OCAML`), the backup thread goes on to the condition variables `domain_cond`, which releases the `domain_lock`, which is now locked by the main thread which continues. 

Before entering a blocking section, the domain releases the `domain_lock`, but also notifies the backup thread possibly waiting on the `domain_cond`, so that it may wake up and service the interrupts.

## Backup thread servicing interrupts

When the main thread is in the blocking section, the backup thread services outstanding interrupts. If there are none, it waits on the interruptor condition variable (`s->cond`), which will be signaled by domains that interrupt this domain. 

## Idling the backup thread when main thread leaves the blocking section

When the backup thread is servicing outstanding requests, it also checks to see if the main thread is attempting to leave the blocking section. If so, it releases the interruptor lock (`s->lock`) and goes on to wait on `domain_cond`. When the main thread returns from the blocking call, it may be the case that the backup thread is waiting on the interruptor condition variable `s->cond`. Hence, the main thread notifies the backup thread to wake up by signaling `s->cond` in `caml_leave_blocking_section`. 

## Domain termination

There is a little bit of work done to terminate the backup thread when the domain terminates. The backup thread is asked to terminate with `MSG_TERMINATE`. The main thread does not join the backup thread. This is safe since the only issue would be the new backup thread that will be created when the domain slot is reused by a new domain. But the new domain first obtains the `domain_lock` before doing anything interesting (such as spawning the backup thread), and the last thing the backup thread does before termination is to release the `domain_lock`.


kayceesrk:
Fixes #90.

runtime/domain.c
```diff
@@ -84,7 +88,12 @@ struct dom_internal {
   struct domain state;
   struct interruptor interruptor;
 
-  caml_plat_mutex roots_lock;
+  /* backup thread */
+  int backup_thread_running;
```
ctk21, line 93: I couldn't see where we initialize the `backup_thread_running` variable to a known state.
I see `backup_thread_msg` getting done in `create_domain`
and the lock/cond pair in `caml_init_domains`

runtime/domain.c
```diff
@@ -821,15 +885,32 @@ CAMLexport void (*caml_leave_blocking_section_hook)(void) =
    caml_leave_blocking_section_default;
 
 CAMLexport void caml_leave_blocking_section() {
-  caml_plat_lock(&domain_self->roots_lock);
+  dom_internal* self = domain_self;
+
+  atomic_store_rel(&self->backup_thread_msg, MSG_ENTERING_OCAML);
+
+  if (self->backup_thread_running) {
+    /* Lock is necessary here to avoid races with the backup thread
```
ctk21: I didn't really understand this comment and found it confusing. 

I think what you mean is: 
 When leaving a blocking section, the backup thread will be waiting on `interruptor.lock` to service interrupts. We need to signal the backup thread such that it stops servicing interrupts and drops `domain_lock`.

runtime/domain.c
```diff
@@ -1081,7 +1162,13 @@ static void domain_terminate()
     Caml_state->critical_section_nesting = 0;
     acknowledge_all_pending_interrupts();
   }
-  caml_plat_unlock(&domain_self->roots_lock);
+
```
abbysmal, line 1185: Worth encasing the following block in a check for `backup_thread_running` ?

abbysmal:
Thank you for this!
I took a look at the implementation and it looks good to me, however more and more astute eyes than mine may be for the better. :)
I ran it as well in the context of #342. The previously mentioned deadlock is effectively gone, the interrupt being now properly serviced and the aforementioned deadlock resolved.

runtime/domain.c
```diff
@@ -821,15 +885,32 @@ CAMLexport void (*caml_leave_blocking_section_hook)(void) =
...
 CAMLexport void caml_enter_blocking_section() {
+  dom_internal* self = domain_self;
+
   caml_process_pending_signals();
   caml_enter_blocking_section_hook();
-  caml_plat_unlock(&domain_self->roots_lock);
+  atomic_store_rel(&self->backup_thread_msg, MSG_IN_BLOCKING_SECTION);
+  /* Wakeup backup thread if it is sleeping */
+  caml_plat_signal(&self->domain_cond);
+  caml_plat_unlock(&self->domain_lock);
```
ctk21, line 930: Is it worth having an assert that if `!caml_domain_alone()` then the backup thread is running?

runtime/domain.c
```diff
@@ -821,15 +885,32 @@ CAMLexport void (*caml_leave_blocking_section_hook)(void) =
    caml_leave_blocking_section_default;
 
 CAMLexport void caml_leave_blocking_section() {
-  caml_plat_lock(&domain_self->roots_lock);
+  dom_internal* self = domain_self;
+
+  atomic_store_rel(&self->backup_thread_msg, MSG_ENTERING_OCAML);
+
```
ctk21, line 905: Is it worth having an assert that if `!caml_domain_alone()` then the backup thread is running?

ctk21 (commented):
I think the code of this PR logically made sense to me and we need it to fix things grinding to a halt when one domain enters a blocking call. 

The "race condition" comment had me perplexed until I thought through that it is about signalling who is responsible for servicing the interrupts. 

On the testing side, can we add a testcase which pushes the domain lifecycle a bit?
I'm just thinking that one day someone might touch this, but not get things right when a domain is reused. Maybe something like:
 ```ocaml
 (* start a dummy domain and shut it down to cause a domain reuse *)
 let d = Domain.spawn (fun _ -> ()) in 
 Domain.join d;
 let _d = Domain.spawn (fun _ ->
    Unix.sleep 10;
    print_endline "Should not reach here!") in
 Gc.full_major ();
 print_endline "OK"
```

File: runtime/domain.c (hunk 2 above)
Comment (kayceesrk) [reply runtime/domain.c]: Yes. But it is a bit more subtle. Without the lock, it is not necessary that the backup thread is waiting on the `interruptor.lock`. Singaling without may lose the signal. 

Let us consider for a moment that we signal without the lock. Consider that the main thread has returned from the blocking call, and is about to call `caml_leave_blocking_section`. Consider also that the backup thread is in the inner loop, has already read the `backup_thread_msg` (which would be `MSG_IN_BLOCKING_SECTION`) but has not performed the conditional check. Now, consider that the main thread executes `caml_leave_blocking_section`. The main thread sets the backup_thread_msg to `MSG_ENTERING_OCAML` signals the condition variable (on which the backup thread is not waiting right now), and goes on to wait to obtain the `domain_lock`. 

Now, the backup thread may continue, observe that there are no interrupts to be handled, and then wait on the `interruptor.cond` condition variable. Unless another domain interrupts this domain, this domain is stuck.

It may be best to just remove that comment.

File: runtime/domain.c (hunk 3 above)
Comment (kayceesrk) [reply runtime/domain.c:1185]: Since this is in `domain_terminate`, which is only called in the case a domain is spawned explicitly, the backup thread is definitely running. 

I'll add an assertion `Assert (domain_sef->backup_thread_running)`.

File: runtime/domain.c (hunk 2 above)
Comment (ctk21) [reply runtime/domain.c]: I'm not suggesting that you don't need the lock. As you say, you need to have it so that `backup_thread_func` is forced to be at the wait or has released the lock.

File: runtime/domain.c (hunk 4 above)
Comment (kayceesrk) [reply runtime/domain.c:930]: Done.

File: runtime/domain.c (hunk 3 above)
Comment (kayceesrk) [reply runtime/domain.c:1185]: This assertion triggered a bug in domain and backup thread termination. The latest commit makes the main domain thread wait for the backup thread to terminate before terminating itself.

File: runtime/domain.c (hunk 5 above)
Comment (kayceesrk) [reply runtime/domain.c:905]: done.

File: runtime/domain.c (hunk 1 above)
Comment (kayceesrk) [reply runtime/domain.c:93]: Done.

**Comment by kayceesrk** [comment]
There was a bug involving terminating domain and backup thread, and domain slot reuse. Now the main domain thread waits for the backup thread to terminate before it terminates. This should make the race condition go away.

File: runtime/domain.c (hunk 2 above)
Comment (kayceesrk) [reply runtime/domain.c]: I think removing the comment is the right thing to do. It is recommended that `pthread_cond_signal` is called while holding the lock. So the comment doesn't say anything special.

**Comment by kayceesrk** [comment]
The PR is ready now.

File: runtime/domain.c [hunk 6]
```diff
@@ -374,8 +384,74 @@ struct domain_startup_params {
...
+{
+  int err;
+
+  if (di->backup_thread_running == 0) {
+    Assert (di->backup_thread_msg == BT_INIT ||     /* Using fresh domain */
+            di->backup_thread_msg == BT_TERMINATE); /* Reusing domain */
+
+    while (atomic_load_acq(&di->backup_thread_msg) != BT_INIT) {
+      /* Give a chance for backup thread on this domain to terminate */
+      caml_plat_unlock (&di->domain_lock);
```
ctk21, line 437: I couldn't work out if we need to add this:
` caml_plat_signal(&di->domain_cond);`
or if the pthread cond signalling from `domain_terminate` will always wake the backup thread even if the `domain_lock` then ends up taken by this thread reusing the domain.

ctk21 (commented):
I think this works (aside the small query about if we need to signal when waiting for the terminating thread). 

Always tricky with parallel code using shared state to know we have everything covered.

runtime/domain.c
kayceesrk: The signal isn't necessary here. If the backup thread had been created, it is either waiting to obtain the lock, or if it has got the lock and it is waiting on the condition variable (since the `domain_terminate` code holds the `domain_lock`). In the former case, eventually, the backup thread will get the lock, read the message, and go on to terminate, setting the `backup_thread_msg` to `BT_INIT`. In the latter case, it will wake up certainly and will terminate.

ctk21: Ok, that makes sense. My concern was more around `pthread_cond_wait` and if that could somehow break. That's actually fine because `pthread_cond_signal` will push the backup thread to be waiting on the mutex.

kayceesrk:
thanks for the review!

PR #366, opened 2020-06-29
Add event to record idle domains

Sudha247:
This patch adds events `domain/idle_wait` and `domain/send_interrupt` to track domains that are not engaged in any work. This could be useful to debug imbalance in task distribution in a Multicore OCaml program. 

Following is an example eventlog for a program with unbalanced task distribution.

This is the eventlog generated by `parallel_minor_gc`:
image

This is the eventlog generated by this patch:
image

In the latter graph, we can see that one domain is active till the end and other domains have varying time periods of `domain/idle_wait` from which we can infer that the task distribution is not balanced.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1349,11 +1349,11 @@ CAMLprim value caml_ml_domain_interrupt(value domain)
   struct interruptor* target =
     &all_domains[unique_id % Max_domains].interruptor;
 
-  caml_ev_begin("domain/idle_wait");
+  caml_ev_begin("domain/send_interrupt");
   if (!caml_send_interrupt(&domain_self->interruptor, target, &handle_ml_interrupt, &unique_id)) {
     /* the domain might have terminated, but that's fine */
   }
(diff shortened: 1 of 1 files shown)


runtime/domain.c
```diff
@@ -1347,9 +1349,12 @@ CAMLprim value caml_ml_domain_interrupt(value domain)
   struct interruptor* target =
     &all_domains[unique_id % Max_domains].interruptor;
 
+  caml_ev_begin("domain/idle_wait");
```
kayceesrk: This event pair is not the same as the other two since sending interrupt polls for interrupts rather than waiting on a condition variable. Maybe use a different name for this event; perhaps `domain/send_interrupt`.

kayceesrk (changes requested):
I've left a small comment about naming the event associated with sending interrupts. Once that is changed, I shall merge.

runtime/domain.c
Sudha247: I have changed it to `domain/send_interrupt`. Thanks!

PR #367, opened 2020-07-09
remove some unused rpc consumers

sadiqj:
This PR removes two sets of code:

1. Sending and handling read faults
2. Stealing work during marking

Both make use of the domain rpc mechanism but are no longer in use.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -383,57 +383,6 @@ static uintnat default_slice_budget() {
   return computed_work;
 }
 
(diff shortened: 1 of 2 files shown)


ctk21:
I'm in favour of this PR. It makes it easier to understand how the domain API is used elsewhere in the codebase by removing entry points that never happen.

LGTM

kayceesrk:
Happy to remove it. 

I should mention that work-sharing in the major GC was not explored rigorously. This was before sandmark existed. That said, we can retrieve the design from git history if need be.

Issue #339, opened 2020-05-14
Proposal for domain-local storage


sadiqj:
This is a proposal for implementing domain-local storage in multicore. We've had a couple of cases where it would have been really useful recently (https://github.com/ocaml-multicore/ocaml-multicore/issues/333 and https://github.com/ocaml-multicore/ocaml-multicore/issues/336). I'm putting it here so we can get some feedback and see if there's potentially a better approach.

The current proposal would allow you to do something along the lines:

```
let dls = Domain.Local.create (fun () ->Random.State.make_self_init ())

... sometime later in a domain ...

let local_random = Domain.Local.get dls
```
To make this work we would need to:

1. Add a hash table to domain state (which would be a root)
2. Add a global domain_local_counter
3. When a new domain local is created, the domain_local_counter is incremented. This value is used as the newly created domain local's key into the domain state's hash table.
4. Retrieving the value from a domain local would involve it looking up the entry in the domain state's hash table for it's key and if not present then using the function provided to create, populate and return the value.
5. To avoid leaking entries in the domain state's hashtable we would probably need a finalizer for domain locals that removes their entries from all domain's hashtables.

(5) isn't ideal.

stedolan:
This sounds useful, and I think you can do it more easily than this. `(Domain.self () :> int)` is always a small integer, so an array makes a fine domain-local storage table.

Detecting domain termination for finalisation does sound awkward, though. Maybe make the user do it with explicit acquire and release functions?

sadiqj:
The necessity for the per-domain hash table is that you may have multiple domain locals (that may be in libraries you're using as well). 

What I think we need to avoid is users of domain local state needing to coordinate because that makes it almost impossible to use inside of libraries. I worry that having explicit acquire and release functions might do similar.

lpw25:
Maybe a stupid question, but why not one array per piece of domain local state? Why isn't `Domain.Local.t` just an `'a array`? Or maybe something like `'a ref array Atomic.t` so you can enlarge the array with compare and swap if more domains appear.

sadiqj:
That works but brings the mirror issue of (5) from earlier - unless we have a way of being notified about Domain termination we will hold on to domain local state after a domain's exited (and potentially unwittingly hand it out to a new domain).

There's an issue for potentially creating an at_exit hook for Domain (https://github.com/ocaml-multicore/ocaml-multicore/issues/326) which would solve that problem. We were hoping there might be an approach that didn't require that though.

[...]

PR #372, opened 2020-07-17
Domain-local Storage

Sudha247:
This PR aims to add Domain-local Storage proposed in #339.

### Interface

The `Domain.DLS` interface similar to what's described [here](https://github.com/ocaml-multicore/ocaml-multicore/issues/339#issuecomment-650117104), provides the functions `new_key`, `get` and `set`. Sample usage is illustrated below:

```
# let k1 : int Domain.DLS.key = Domain.DLS.new_key ();;
val k1 : int Domain.DLS.key = <abstr>
# Domain.DLS.set k1 100;;
- : unit = ()
# Domain.DLS.get k1;;
- : int option = Some 100
# Option.get @@ Domain.DLS.get k1;;
- : int = 100
```

### Implementation

The thread local variables are stored as a list of key and values. The list resides in `Caml_state` and is updated every time a new variable is added to the thread local storage.


runtime/caml/domain_state.tbl
```diff
@@ -104,6 +104,8 @@ DOMAIN_STATE(struct pool**, pools_to_rescan)
 DOMAIN_STATE(int, pools_to_rescan_len)
 DOMAIN_STATE(int, pools_to_rescan_count)
 
+DOMAIN_STATE(value, tls)
```
kayceesrk: Perhaps `tls_root` would a better name for this.

runtime/domain.c
```diff
@@ -1432,3 +1432,14 @@ CAMLprim value caml_ml_domain_cpu_relax(value t)
...
   return Val_unit;
 }
+
+CAMLprim value caml_domain_tls_set(value t)
+{
+  Caml_state->tls = t;
+  return Val_unit;
+}
+
+CAMLprim value caml_domain_tls_get()
```
kayceesrk:
```suggestion
CAMLprim value caml_domain_tls_get(value unused)
```

runtime/roots.c
```diff
@@ -46,6 +46,9 @@ CAMLexport void (*caml_scan_roots_hook)(scanning_action, void* fdata, struct dom
 
 void caml_do_roots (scanning_action f, void* fdata, struct domain* d, int do_final_val)
 {
+  if (d->state->tls)
```
kayceesrk: Why is this check needed? Since `tls` is a root, it should never be `NULL`. A freshly initialised `tls` should be `[]`, whose representation is `Val_unit` (== `0x1`). Hence, this check will never fail.

stdlib/domain.ml
```diff
@@ -112,3 +112,35 @@ let get_id { domain; _ } = domain
...
+
+  type 'a key = int ref
+
+  external get_primitive : unit -> (int ref * Obj.t) list
+    = "caml_domain_tls_get"
+
+  external set_primitive : (int ref * Obj.t) list  -> unit
+    = "caml_domain_tls_set"
+
+  let new_key () = let v = ref 0 in v
```
kayceesrk:
```suggestion
  let new_key () = ref 0
```

stdlib/domain.ml
```diff
@@ -112,3 +112,35 @@ let get_id { domain; _ } = domain
...
+
+  let set k x =
+    let cs = Obj.repr x in
+    let old = get_primitive () in
+    set_primitive @@ (k,cs)::old
```
kayceesrk: Multiple updates to the same key will create multiple entries in the list. The list will keep growing. Instead of 

```
(int ref * Obj.t) list
```

use 

```
(int ref * Obj.t ref) list
```

`set` should first check that the key is already present in the list. If so, update the second element of the pair with the new value. Otherwise, push to the front of the list. It may be nicer to do:

```
type entry = { key : int ref; slot : Obj.t ref }

external get_primitive : unit -> entry list
    = "caml_domain_tls_get"

  external set_primitive : entry list  -> unit
    = "caml_domain_tls_set"
```

A value of type `entry` has the same memory representation as `(int ref * Obj.t ref)`.

stdlib/domain.ml
```diff
@@ -112,3 +112,35 @@ let get_id { domain; _ } = domain
...
+  let get k =
+    let check k1 k2 =
+      let a1 : int = Obj.magic k1 in
+      let a2 : int = Obj.magic k2 in
+      (a1 = a2)
+    in
+    let rec search k l =
+      match l with
+      | [] -> None
+      | (k', v')::tl -> if (check k k') then (Some v') else search k tl
```
kayceesrk: `check` is not necessary here. It is also subtly wrong (due to integer encoding). What you want here is physical equality `==` 
```suggestion
      | (k', v')::tl -> if k == k' then (Some v') else search k tl
```

stdlib/domain.ml
```diff
@@ -112,3 +112,35 @@ let get_id { domain; _ } = domain
 
 let self () = Raw.self ()
 
+module TLS = struct
+
+  type 'a key = int ref
+
+  external get_primitive : unit -> (int ref * Obj.t) list
```
kayceesrk:
```suggestion
  external get_tls_list : unit -> (int ref * Obj.t) list
```

stdlib/domain.mli
```diff
@@ -110,3 +110,26 @@ module Sync : sig
   external poll : unit -> unit = "%poll"
   (** poll for interrupts *)
 end
+
+module TLS : sig
+(** Thread Local Storage *)
```
kayceesrk:
```suggestion
(** Thread-local Storage *)
```

stdlib/domain.mli
```diff
@@ -110,3 +110,26 @@ module Sync : sig
...
+
+    val new_key : unit -> 'a key
+    (** Returns a new key for accessing thread-local variable. The type of
+        the variable which will be stored with the key needs to be specified
+        during invocation. For example, to generate a key associated to an
+        integer:
+        let k1 : int Domain.TLS.key = Domain.TLS.new_key () *)
+
+    val set : 'a key -> 'a -> unit
+    (** [set k v] updates the calling domain's thread local state to
```
kayceesrk:
```suggestion
    (** [set k v] updates the calling domain's thread-local state to
```

stdlib/domain.mli
```diff
@@ -110,3 +110,26 @@ module Sync : sig
...
+        associate the key [k] with value [v] *)
+
+    val get : 'a key -> 'a option
+    (** [get k] returns [Some v] if a value was previously associated with
+        the key [k] on the calling domain. Otherwise returns [None].  *)
```
kayceesrk:
```suggestion
        the key [k] in the calling domain's thread-local state. Otherwise returns [None].  *)
```

kayceesrk (changes requested):

kayceesrk:
You are missing the initialisation and clearing up of `Caml_state->tls` field. 

Initialization is needed at https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/runtime/domain.c#L288 where you do `domain_state->tls = Val_unit`. 

Cleanup is needed at https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/runtime/domain.c#L1177, where you do `domain_state->tls = Val_unit`.

runtime/caml/domain_state.tbl
Sudha247: Yes, changed it.

runtime/roots.c
Sudha247: It was `NULL` at times possibly due to not initialising `tls`. Works now without the check after the initialisation is done. Thanks for pointing that out!

stdlib/domain.ml
Sudha247: Growing list is indeed going to be a problem. But updating a reference inside a list was getting quite tricky. I have now removed old associated values (if any present) and then updated the new entry, so the list now has only one value associated with a key.

Another possible solution to this could be using a hash table that looks something like this:
```ocaml
module RefHash =
    struct
      type t = int ref
      let equal i j = i == j
      let hash = (* need a suitable hash function here, default hash function will end 
      up with many collisions since we are using `ref 0` as all the keys*) 
    end
```
As long as there aren't too many variables in the thread-local state, I think the current implementation is okay.

stdlib/domain.ml
```diff
@@ -112,3 +112,25 @@ let get_id { domain; _ } = domain
...
+
+  external set_tls_list : (int ref * Obj.t) list  -> unit
+    = "caml_domain_tls_set"
+
+  let new_key () = ref 0
+
+  let set k x =
+    let cs = Obj.repr x in
+    let old = get_tls_list () in
+    set_tls_list @@ (k,cs)::(List.remove_assq k old)
```
kayceesrk: I'm not particularly fond of this. In the worst case, If you remove the last element in the list, then the whole spine of the list will need to be allocated again. I still recommend making the second element of the pair a ref.

stdlib/domain.ml
kayceesrk: I've left a comment. The current implementation using `remove_assq` is inefficient. I'd still recommend going to use the ref.

runtime/domain.c
```diff
@@ -1175,6 +1176,7 @@ static void domain_terminate()
     caml_plat_unlock(&s->lock);
   }
 
+  domain_state->tls_root = Val_unit;
```
ctk21: I'm not convinced this is going to be safe, shouldn't there be a `write_barrier`?

runtime/domain.c
```diff
@@ -1432,3 +1434,14 @@ CAMLprim value caml_ml_domain_cpu_relax(value t)
   handle_incoming_otherwise_relax (Caml_state, self);
   return Val_unit;
 }
+
+CAMLprim value caml_domain_tls_set(value t)
+{
+  Caml_state->tls_root = t;
```
ctk21: Do we need a `write_barrier` here?

ctk21 (commented):
This looks sensible in the whole to me. 

Couple of comments:
- I know we always gravitate to naming things as thread (myself included!) but I wonder if we are going to cause confusion down the line. This is really `Domain Local Storage`. If multiple systhreads share a domain, then they will use the same `dls`. If fibers/continuations share a domain, they will use the same `dls`; although notice that if a fiber/continuation moves domain, then they will get a different `dls`. Should we rename `Thread Local Storage` -> `Domain Local Storage` and `tls` -> `dls` in the code and documentation?
- We often have issues with domain lifecycle/reuse and garbage collection when we come to touch it in the future; is it possible to beef up the tests to include one that reuses a domain and another that forces a major GC with reuse of the `dls`?
- I don't think operating raw on mutable pointers is safe with the garbage collector, another part of the code could be using the contents of the old mutable pointer before it is changed. Does it simplify things to create, modify and delete the variable `tls_root` using the `globroot.c` functions `caml_create_root`, `caml_modify_root`, `caml_delete_root`? Then we don't need to manage the GC with this change.

runtime/domain.c
kayceesrk: The local roots are marked immediately after the end of the cycle, before the mutator gets to run. Hence, `tls_root` will always be `MARKED`. Write barrier will serve to only mark the object which we know is marked.

runtime/domain.c
kayceesrk: Same as above.

hhugo:
What about using an univ_map (e.g. https://github.com/janestreet/core_kernel/blob/master/src/univ_map_intf.ml) too avoid any kind of linear complexity ?

kayceesrk:
> Should we rename Thread Local Storage -> Domain Local Storage and tls -> dls in the code and documentation?

Good point. We most definitely should.

kayceesrk:
> Does it simplify things to create, modify and delete the variable tls_root using the globroot.c functions caml_create_root, caml_modify_root, caml_delete_root? Then we don't need to manage the GC with this change.

I think so. @Sudha247 an example of how to do this is at https://github.com/ocaml-multicore/ocaml-multicore/blob/711d589bc19b87836500e08ae489a4ca8372b57e/runtime/domain.c#L284 and https://github.com/ocaml-multicore/ocaml-multicore/blob/711d589bc19b87836500e08ae489a4ca8372b57e/runtime/domain.c#L1180. When you use this, you no longer need to modify `caml_do_roots`.

stdlib/domain.ml
```diff
@@ -112,3 +112,40 @@ let get_id { domain; _ } = domain
...
+
+  external set_dls_list : entry list  -> unit
+    = "caml_domain_dls_set"
+
+  let new_key () = ref 0
+
+  let set k x =
+    let cs = Obj.repr x in
+    let old = get_dls_list () in
+    let rec add_or_update_entry k v l l' =
```
kayceesrk: Isn't `l` always `old`. No need to pass this around explicitly.

stdlib/domain.ml
```diff
@@ -112,3 +112,40 @@ let get_id { domain; _ } = domain
...
+      match l' with
+      | [] -> {key = k; slot = ref v}::l
```
kayceesrk:
```suggestion
      | [] -> {key = k; slot = ref v}::old
```

stdlib/domain.ml
```diff
@@ -112,3 +112,40 @@ let get_id { domain; _ } = domain
...
+      | hd::tl -> if (hd.key == k) then begin
+        hd.slot := v;
+        l
```
kayceesrk:
```suggestion
        old
```

stdlib/domain.mli
```diff
@@ -110,3 +110,26 @@ module Sync : sig
...
+        integer:
+        let k1 : int Domain.DLS.key = Domain.DLS.new_key () *)
+
+    val set : 'a key -> 'a -> unit
+    (** [set k v] updates the calling domain's domain-local state to
+        associate the key [k] with value [v] *)
+
+    val get : 'a key -> 'a option
+    (** [get k] returns [Some v] if a value was previously associated with
+        the key [k] in the calling domain's domain-local state. Otherwise returns [None].  *)
```
kayceesrk: Runs over 80 chars?

stdlib/domain.ml
Sudha247: Yeah, that makes sense. Modified it.

stdlib/domain.mli
Sudha247: Oh yes, I didn't realise that, thanks! fixed now.

kayceesrk:
> What about using an univ_map (e.g. https://github.com/janestreet/core_kernel/blob/master/src/univ_map_intf.ml) too avoid any kind of linear complexity ?

I am leaning towards keeping the list-based implementation and linear complexity. My reasoning is this -- wherever one may need to store lots of keys (and suffer linear complexity), one may instead use a single slot in the domain-local storage which stores the map data structure that permits efficient lookup, insert and delete operations. @Sudha247 perhaps the documentation should mention that `get` and `set` are linear-time operations.

stdlib/domain.mli
```diff
@@ -118,19 +118,21 @@ module DLS : sig
...
+        For example, to generate a key which is associated to an integer:
+
+        let k : int Domain.DLS.key = Domain.DLS.new_key () *)
 
     val set : 'a key -> 'a -> unit
-    (** [set k v] updates the calling domain's domain-local state to
-        associate the key [k] with value [v] *)
+    (** [set k v] updates the calling domain's domain-local state to associate
+        the key [k] with value [v]. It overwrites any previous values associated
+        to [k], which cannot be restored later. [set] is a linear-time
```
kayceesrk: We can make this a little more precise. How about

```
[set] is an [O(n)] operation where [n] is the number of keys set on this domain. 
```

kayceesrk (approved):

stdlib/domain.ml
```diff
@@ -112,3 +112,40 @@ let get_id { domain; _ } = domain
...
+
+  external set_dls_list : entry list  -> unit
+    = "caml_domain_dls_set"
+
+  let new_key () = ref 0
+
+  let set k x =
+    let cs = Obj.repr x in
+    let old = get_dls_list () in
+    let rec add_or_update_entry k v l =
```
kayceesrk, line 132: Can we make this function return `entry option`? The idea is that if the key was already bound, then we'd have updated the value in-place. Hence, no need to invoke `set_dls_list`. Otherwise, a new entry `e` is added to the head. So the function can just return `Some e`. Then, you can change the last line of `set` function to:

```
match add_or_update_entry k cs old with
| None -> ()
| Some e -> set_dls_list (e::old)
```

PR #373, opened 2020-07-17
Fix the opam pin command in case the current directory name has spaces

kit-ty-kate:
One of your user came up with this issue. Their current directory had spaces and too many argument would be given to `opam pin`.


kayceesrk:
Thanks!

Issue #357, opened 2020-06-15
Implementation of systhreads with pthreads


ctk21:
Following further discussion around implementing systhread compatibility with [domains](https://github.com/ocaml-multicore/ocaml-multicore/pull/342#issuecomment-643119638), we should explore porting the existing systhread implementation using pthreads to multicore. 

The idea would be to implement systhread as it is done [upstream](https://github.com/ocaml/ocaml) by utilizing the `caml_enter_blocking_section`/`caml_leave_blocking_section` hooks to provide a mechanism to context switch as in `caml_thread_save_runtime_state`/`caml_thread_restore_runtime_state`. It would need to make sure that we interface with the multicore runtime to handle local roots registration, local root scanning and backup threads (#356) correctly. It should inherit the existing systhread scheduling mechanism. 

The aim is to provide an implementation which is compatible with the existing systhread. 

It is left as future work to support a more advanced threading implementation that integrates with effects, as discussed on #100.

jhwoodyatt:
Presumably this would still provide the global thread interlock behavior, i.e. only one thread is allowed to proceed at any particular time even where threads are spawned from multiple domains. Is that right?

gasche:
Naive question: I assumed that Threads would be implemented with fibers on a single domain. Is that not a sensible step forward anymore? My understanding of the state of fibers was that they are not efficient enough to be exposed to users right now (I remember a Discuss post mentioning that their implementation would need to change first to use assembly for some operations, which multicore-ocaml has decided to delay for later); but if there is an internal implementation (that is a bit slow), wouldn't that be a reasonable basis for Threads?

ctk21:
There are currently two implemented parallelism & concurrency abstractions in multicore:
- domains: these execute in parallel and are fairly heavyweight coming with their own minor heap and the synchronization required to manage a multicore minor and major GC.
- fibers exposed through effects: fibers are light weight execution contexts with dynamic stacks that allow for flexible concurrency. fibers are (currently) created and scheduled using effects. The continuations can be passed between domains allowing for complex schedulers to be constructed on top of the multicore runtime. However fibers currently do not have any mechanism for handling blocking sections that could occur in external system or C-library calls; if a fiber enters a blocking C call, then all fibers in that domain will be blocked until the call returns.

This is where implementing systhreads with pthreads sharing a domain comes in:
[...]

PR #381, opened 2020-07-30
Reimplementing Systhreads with pthreads (Domain execution contexts)

abbysmal:
This PR acts as a RFC for Domain Execution Contexts.

Following the decision made in #357, this branch is an attempt at adding a mechanism to Multicore that would allow multiple threads to run atop a domain and execute code on it in a similar fashion to systhreads.

https://github.com/ocaml-multicore/ocaml-multicore/wiki/Concurrency-and-parallelism-design-notes for more notes about the planned story for `Thread` in Multicore.

This branch started as a standalone implementation of the planned `domain execution contexts`, but ended up also borrowing a lot from the `systhreads` implementation found in OCaml currently. 
This branch is still a work in progress, as a few features are still missing to this day:

- No *forced preemption*: OCaml's systhreads runs a ticking thread in the background that will regularly send signals to running threads to force preemption, for what I assume is related to fairness. This is not yet implementation.
- Channel mutexes are not handled properly yet.
- The error handling story is still lacking, code skips a few verification.
- Registering threads from C is not yet implemented.
- The signal handling story needs improvement I believe.

The current state of this implementation however still allows to run the `lib-threads` testsuite, to completion with no error.
The `lib-systhreads` testsuite however cannot succeed because of the aforementioned absence of the ticking thread.
Also to be noted that this branch builds `dune 2.6.1` successfully and allowd compiling non-trivial projects when I gave it a test run a few days back.

A few notes about the implementation:

A few additions and tweaks were required to the runtime's API, and more are going to be needed:

- The `caml_scan_roots_hook` had to be rearranged to properly handle scanning local roots and stacks for a given `dec`.
- The addition of a `no_mutex` version of `caml_plat_cond`, which may be not desirable but was added to match 1 to 1 the API provided by the `Condition` module, where a mutex is separately affixed to a cond, which is not how `caml_plat_cond` usually works. I am willing to rework this if required, I attempted to stick to using the platform API as much as possible while porting some pieces from systhreads.
- For running decs across multiple domains, a hook may be required on domain creation/termination. (to handle initialization logic.)

While the idea for domain execution contexts started as something somewhat separate from `Thread` (merely just as a mechanism to enable it.), this implementation ended up borrowing a fair bit from `systhreads`.
This raises the question as to whether `decs` are to be a real concept of its own, with a separate API than systhreads. or if `decs` are indeed really just a port of systhreads to Multicore.
For instance, I am not sure `decs`, as an interim concurrency primitive to deal with blocking C calls, benefits from matching 1:1 with all the features of `systhreads`. (I'm thinking of the ticking thread and preemption through signals. in particular.).
In the current state, `decs` and `systhreads` are the same thing, maybe `decs` could provide just a simple API allowing to build `systhreads` on top of it, among other use cases.

I will resume work on implementing the last remaining features and from there we may rework the implementation to a desirable state in terms of API and consistency across planned features. 

(and thank you very much @ctk21 for your help figuring out many bugs we encountered while working on this. :)

Edit (19/08/2020): Removal of a few entries in the "missing items" list.


avsm:
Dune works great with this under load so far (on a monorepo with around ~30000 jobs). Lots of build failures due to omp being missing, but that's a separate issue.

otherlibs/systhreads/decs.c
```diff
@@ -0,0 +1,801 @@
...
+  if (waiters == 0) {
+    caml_plat_unlock(&m->lock);
+    return;
+  }
+
+  atomic_store_rel(&m->busy, 0);
+  caml_plat_signal(&m->free);
+  atomic_fetch_add(&m->waiters, +1);
+  do {
+    /* Note: the POSIX spec prevents the above signal from pairing with this
```
avsm, line 214: Just so I understand this correctly, what part of the POSIX spec prevents the above signal from hitting the wait below?

abbysmal: This part of the logic is imported from `systhreads`, especially this implementation comes from https://github.com/ocaml/ocaml/pull/2112
When I reworked this code for multicore I looked into the specifics, and I think it means that a `signal` will only signal to currently `wait`ing threads. (if no waiters are available, it will effectively do nothing, noting that there's also a specific handling for this at the start of the function for what I gathered was optimization purpose).
As such there's not really a way this `signal` can pair with the following `wait`.
But maybe I missed something more subtle.

Slightly related, but since the linked PR does mention this, I think some locking logic may benefit from a rework since the previous use of `volatile` is gone with the use of C11 atomic operations through `platform.h`.

runtime/roots.c
```diff
@@ -70,5 +71,5 @@ CAMLexport void caml_do_local_roots (scanning_action f, void* fdata,
     }
   }
   /* Hook */
```
ctk21: We should probably move the `/* Hook */` comment with the `caml_scan_roots_hook` (or remove the comment completely).

runtime/minor_gc.c
```diff
@@ -647,7 +647,8 @@ void caml_empty_minor_heap_promote (struct domain* domain, int participating_cou
 #endif
 
   caml_ev_begin("minor_gc/local_roots");
-  caml_do_local_roots(&oldify_one, &st, domain, 0);
+  caml_do_local_roots(&oldify_one, &st, domain->state->local_roots, domain->state->current_stack, 0);
+  if (caml_scan_roots_hook != NULL) (*caml_scan_roots_hook)(&oldify_one, &st, domain);
   caml_scan_stack(&oldify_one, &st, domain_state->current_stack);
```
ctk21: We should remove this call to `caml_scan_stack` as it is now done in `caml_do_local_roots`

runtime/roots.c
```diff
@@ -42,24 +42,25 @@
...
+  caml_do_local_roots(f, fdata, d->state->local_roots, d->state->current_stack, do_final_val);
+  if (caml_scan_roots_hook != NULL) (*caml_scan_roots_hook)(f, fdata, d);
   caml_scan_global_roots(f, fdata);
   caml_final_do_roots(f, fdata, d, do_final_val);
+
 }
 
 CAMLexport void caml_do_local_roots (scanning_action f, void* fdata,
-                                     struct domain* domain, int do_final_val)
+                                     struct caml__roots_block *local_roots, struct stack_info *current_stack,  int do_final_val)
```
ctk21, line 57: While we are touching this, shall we remove `do_final_val`, it's actually unused in this function.

runtime/domain.c
```diff
@@ -906,13 +906,46 @@ void caml_handle_gc_interrupt() {
   }
 }
 
+CAMLexport void caml_bt_leave_blocking_section_hook(void)
+{
+  dom_internal* self = domain_self;
+
```
ctk21, line 968: Can we put back the assert:
`Assert(caml_domain_alone() || self->backup_thread_running);`

runtime/domain.c
```diff
@@ -906,13 +906,46 @@ void caml_handle_gc_interrupt() {
...
+  caml_plat_lock(&self->domain_lock);
+
+  return;
+}
+
+CAMLexport void caml_bt_enter_blocking_section_hook(void)
+{
+  dom_internal* self = domain_self;
+
```
ctk21, line 991: Can we put back the assert:
`Assert(caml_domain_alone() || self->backup_thread_running);`

otherlibs/systhreads/decs.c
```diff
@@ -0,0 +1,801 @@
...
+
+static int dec_atfork(void (*fn)(void))
+{
+  return pthread_atfork(NULL, NULL, fn);
+}
+
+
+CAMLprim value caml_dec_initialize(value unit)
+{
+  // FIXME: init guard missing
```
ctk21: Is something left to do here?

ctk21 (commented):
So good to have systhreads working with dune. 

This review is focusing more on the touch points between the runtime and systhreads (see code comments). 

There's a couple of things we should add test coverage for:
 - Can we add a test to make sure that the backup thread is going to run when we set the `caml_enter_blocking_section_hook`? (see `tests/parallel/backup_thread.ml` probably should copy that and add in a couple of systhreads)
 - Are we able to add a domain lifecycle test where we create a couple of domains where each creates a couple of systhreads, they do something (including maybe a major gc), then they come down and do it all again? We've often ended up with domain lifecycle stuff not quite working out. 

Also this PR isn't cleanly passing the testsuite, do we need to disable some of them as there are things that will be further work?

abbysmal:
Hello,

I worked a bit more on this PR in the last few weeks.
I will update the initial post to reflect changes on the branch.
The last changes includes a few fixes enabling the branch to run a wider range of software in the ecosystem (utop in particular), as well as a less finicky handling of `fork`.
To note as well, the changes includes the ability to run `decs`/`systhreads` accross many domains at once.
The core bulk of this addition was to figure out how to interact with the backup thread.
There's quite a few addition to the domain API as a result, come more desirable than others, I think we can discuss these changes and figure out the proper way forward.

Excerpt of a few of these changes:
- The backup thread API is now split between two parts: the signaling part (through the caml_bt_enter/exit_ocaml functions), and the domain_lock acquiring part. Such a distinction is required in order to allow the dec scheduler to trigger the backup thread only when required. This API may be a bit heavy handed, and I think there's may be a way to reduce the amount of locks in the codebase. (as `domain_lock` and the `caml_dec_main_lock->lock` are doing strictly similar jobs, ie, preventing threads to execute OCaml code when the time is not right.).
@ctk21 has a few suggestions regarding how we could handle the backup thread state in a slightly more refined way (introducing a `backup_thread_lock`), I think addressing this branch's, overly complicated state, will be important in the future.
- The `caml_plat_cond` API receives a few rather unsafe additions, which are merely a result of the inherent incompatibility between the systhreads-style Conditions with no explicit mutexes bound to a condvar, and the platform variant which makes the style explicit. This should eventually be refactored later on.
- We uncovered a bug with @ctk21 where the testsuite would not pass, due to a corruption happening with the debug runtime, where the layout of `domain_state` is not the same as the one `systhreads` expect. The debug runtime indeed introduce a new field, and this has the potential to corrupt programs with different assumption on domain_state's layout, when fiddling around to much with `Caml_state`. We decided to make `inside_stw_handler` static for now as a result.
- We have one more issue in the testsuite: the Clang/MacOS testcase beat.ml does not pass currently, as one counter is off by one. I did not investigate it further yet, this seems to be a subtle timing issue, or a scheduling bug, happening only on MacOS, that will need proper investigation further. However the Linux testsuite passing, I think this can be checked after an initial review and assessment of this work.  

One more thing to note: the latest changes enables using Mutex and Conditions across Domains, meaning these can be used in the same fashion between domains, and between threads.
This property is tested in the `prodcons_domains.ml` testcase.

I think this PR is now ready for another batch of reviews, this is far from being perfect and a few pieces are missing, but I think an iterative approach on this feature would be beneficial. (and the current state seems to enable a wider range of testing on the ecosystem, as it enables building and running dune from a purely multicore switch.).

ctk21 (commented):
I don't think we want to keep `otherlibs/old_systhreads` hanging around.
Shouldn't we just trust our source control to hold the old version?
Or is there a functional reason for it to remain?

kayceesrk:
>  I am not sure decs, as an interim concurrency primitive to deal with blocking C calls, benefits from matching 1:1 with all the features of systhreads. 

The only reason to not name it systhreads is that I did not want the `thread` in the name. Having thread in the name leads to a lot of confusion since we have domains which run OS threads in parallel, fibers which are lightweight user-level threads. The intention of DEC is to implement systhreads and that will be the only consumer I believe.

otherlibs/systhreads/condition.ml
```diff
@@ -1,19 +1,5 @@
-(**************************************************************************)
```
kayceesrk, line 1: Is there a reason to remove the header?

otherlibs/systhreads/decs.c
```diff
@@ -0,0 +1,899 @@
...
+/* overall table for decs accross domains */
+struct caml_dec_table {
+  caml_dec_t all_decs;
+  caml_dec_t current_dec;
+  pthread_key_t dec_key;
+  struct caml_dec_main_lock dec_lock;
+};
+
+/* dec_table instance, up to Max_domains */
+static struct caml_dec_table dec_table[Max_domains];
```
kayceesrk, line 101: Is it better to make this global state part of the per-domain `Caml_state`? Is there a reason for it to be here?

runtime/domain.c
```diff
@@ -953,24 +975,61 @@ CAMLexport void caml_leave_blocking_section() {
...
+{
+  caml_bt_exit_ocaml();
+  caml_bt_release_domain_lock();
+  return;
+}
+
+static void caml_leave_blocking_section_default(void)
+{
+  caml_bt_enter_ocaml();
+  caml_bt_acquire_domain_lock();
```
kayceesrk, line 1014: This function should be named `caml_acquire_domain_lock` shouldn't it? `bt` stands for backup thread, and we don't involve backup threads here, for example, in the case of exactly 1 domain in the program?

kayceesrk:
Perhaps I don't understand the naming convention. Is `old_systhreads` really the DECs now?

kayceesrk (approved):

runtime/domain.c
```diff
@@ -925,23 +944,26 @@ void caml_handle_gc_interrupt() {
   }
 }
 
-static void caml_enter_blocking_section_default(void)
+CAMLexport inline int caml_bt_is_bt_working(void)
```
ctk21, line 947: Took me a couple of takes to parse the `caml_bt_is_bt_working`. Suggested alternative:
 `caml_is_bt_status_in_blocking_section`
Feel free to ignore if you have strong opinions.

runtime/domain.c
```diff
@@ -925,23 +944,26 @@ void caml_handle_gc_interrupt() {
...
+  uintnat status = atomic_load_acq(&self->backup_thread_msg);
+  if (status == BT_IN_BLOCKING_SECTION)
+    return 1;
+  else
+    return 0;
+
 }
 
-static void caml_leave_blocking_section_default(void)
+CAMLexport void caml_bt_acquire_domain_lock(void)
```
ctk21, line 958: Naming here is a bit confusing to me `caml_acquire_domain_lock` is better because we do it even when there is no backup thread (e.g. one domain only).

testsuite/disabled
```diff
@@ -42,15 +42,12 @@ tests/callback/'signals_alloc.ml' with 1.2 (native)
 # TODO: not clear compatibility is sensible for multicore
 compatibility
 
+# TODO: preemption is not properly handled yet (no tick threads implemented)
+lib-systhreads
```
ctk21, line 46: should we split the tests that are TODO into a new directory `lib-systhreads-todo` or `lib-systhreads-preemption` just so we get the test coverage from anything that works?

otherlibs/systhreads/decs.c
```diff
@@ -0,0 +1,899 @@
...
+  new_dec->next = new_dec;
+  new_dec->prev = new_dec;
+
+  #ifdef NATIVE_CODE
+  new_dec->exit_buf = &caml_termination_jmpbuf;
+  #endif
+
+  // Hooks setup, if caml_scan_roots_hook is set, it was done already
+  // by another domain.
+  if (caml_scan_roots_hook != caml_dec_scan_roots) {
```
ctk21, line 409: I don't think this is going to be safe if two domains try to initialize at the same time: 
- domain A starts initializing and gets `caml_scan_roots_hook` set to `caml_dec_scan_roots`, but gets suspended before the other hooks are initialized
- the prev_scan_roots_hook gets written with something unexpected

Could put in a mutex or there you could way with a CAS on `prev_scan_roots_hook` and be careful with the ordering of the operations.

ctk21 (commented):
I looked mainly at the interactions between the runtime and the domains. Not in depth at the systhread internal implementation. 

It looks reasonable to me. Most of my comments are minor tweaks.

otherlibs/systhreads/decs.c
abbysmal: But putting it in Caml_state would require us to have all the underlying definitions related to systhreads be a part of the runtime, if I'm not mistaking?
systhreads sits currently outside of the runtime

runtime/domain.c
abbysmal: I think this makes much more sense, I applied it in the engil/multicore_systhreads change, thank you!

PR #382, opened 2020-07-31
Optimise caml_continuation_use function.

kayceesrk:
The first optimisation is to use `caml_domain_alone` to avoid a CAS. The second
optimisation is to move the call to `caml_gc_log` under `DEBUG` mode.

These two optimisations lead to 25% performance improvement in [generator
example](https://github.com/ocaml-multicore/effects-examples/blob/master/generator.ml).

### DIFF
--- a/runtime/fiber.c
+++ b/runtime/fiber.c
@@ -394,15 +394,25 @@ CAMLprim value caml_clone_continuation (value cont)
 
 CAMLprim value caml_continuation_use (value cont)
 {
+#ifdef DEBUG
   caml_gc_log("cont: is_block(%d) tag_val(%ul) is_minor(%d)", Is_block(cont), Tag_val(cont), Is_minor(cont));
+#endif
   CAMLassert(Is_block(cont) && Tag_val(cont) == Cont_tag);
 
   value v;
   if (!Is_minor(cont) ) caml_darken_cont(cont);
 
   v = Op_val(cont)[0];
-  if (v != Val_ptr(NULL) &&
-      atomic_compare_exchange_strong(Op_atomic_val(cont), &v, Val_ptr(NULL))) {
+
+  if (v == Val_ptr(NULL))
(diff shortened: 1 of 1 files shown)


runtime/fiber.c
```diff
@@ -394,15 +394,25 @@ CAMLprim value caml_clone_continuation (value cont)
...
+
+  if (v == Val_ptr(NULL))
+    caml_invalid_argument("continuation already taken");
+
+  if (caml_domain_alone()) {
+    Field(cont, 0) = Val_ptr(NULL);
+    return v;
+  }
+
+  if (atomic_compare_exchange_strong(Op_atomic_val(cont), &v, Val_ptr(NULL))) {
```
ctk21, line 415: Do we need that `Op_atomic_val(cont)`? 
Couldn't we do `atomic_compare_exchange_strong(v, &v, Val_ptr(NULL))`

ctk21:
LGTM - merging

gasche:
I have a naive curiosity question (coming from a Discuss report): optimizing on caml_domain_alone to use non-atomic operations sounds like a useful abstraction that could be used in several places, could a helper function be reused here? For example `caml_atomic_cas_field` looks very similar to what is done here, could/should it be used?

`caml_atomic_cas_field` uses a write barrier, even in the domain_alone case, while the current PR does not. Is it not useful in the present case?

kayceesrk:
@gasche Just getting to this. Using `caml_atomic_cas_field` would be incorrect since a simple write barrier wouldn't suffice. In order to switch to a stack, we need to mark all the objects that are on the stack. This is analogous to marking the roots on the stack at the beginning of the cycle in stock OCaml. We achieve this using `caml_darken_cont`: https://github.com/ocaml-multicore/ocaml-multicore/pull/382/files#diff-b356594c14072449affc17148d53b1e8R403

PR #384, opened 2020-07-31
Add a primitive to insert nop instruction

kayceesrk:
When debugging low-level code, it is often difficult to identify the start and end of an instruction sequence of interest. This PR introduces `nop` primitive, that inserts the `nop` instruction on x86. This is only for debugging and is not meant to be upstreamed to stock OCaml (though it would be useful to have this upstream).  For example,

```ocaml
let foo () =
  print_string "Hello";
  nop ();
  print_string "World";
  nop ()
```

produces

```asm
camlTest__foo_82:                                                                   
  .cfi_startproc                                                                    
  leaq  -168(%rsp), %r10                                                            
  cmpq  32(%r14), %r10                                                              
  jb  .L104                                                                         
.L103:                                                                              
  subq  $8, %rsp                                                                    
  .cfi_adjust_cfa_offset 8                                                          
.L102:                                                                              
  movq  camlTest__1@GOTPCREL(%rip), %rbx                                            
  cmpq  (%r14), %r15                                                                
  jb  .L105                                                                         
.L106:                                                                              
  movq  camlStdlib@GOTPCREL(%rip), %rax                                             
  movq  328(%rax), %rax                                                             
  call  camlStdlib__output_string_265@PLT                                           
.L100:                                                                              
  movl  $1, %eax                                                                    
  nop                                                                               
  movq  camlTest__2@GOTPCREL(%rip), %rbx                                            
  movq  camlStdlib@GOTPCREL(%rip), %rax                                             
  movq  328(%rax), %rax                                                             
  call  camlStdlib__output_string_265@PLT                                           
.L101:                                                                              
  movl  $1, %eax                                                                    
  nop                                                                               
  addq  $8, %rsp                                                                    
  .cfi_adjust_cfa_offset -8                                                         
  ret                                                                               
  .cfi_adjust_cfa_offset 8        
```

The actual instructions inserted are more than `nop`s since `nop` primitive takes a unit argument. Hence, the `movl $1, %eax` before the `nop`.


asmcomp/selectgen.ml
```diff
@@ -355,8 +356,8 @@ method effects_of exp =
   | Cop (op, args, _) ->
     let from_op =
       match op with
-      | Capply _ | Cextcall _ -> EC.arbitrary
-      | Cpoll | Calloc -> EC.none
+      | Capply _ | Cextcall _ | Cpoll | Cnop -> EC.arbitrary
+      | Calloc -> EC.none
```
ctk21: Is the change for `Cpoll` from `EC.none` to `EC.arbitrary` intentional here?

ctk21 (approved):
Can see this might be useful for debugging and timing sections. 

One query about a change for `Cpoll` in `effects_of` which might be intentional.

asmcomp/selectgen.ml
kayceesrk: The compiler had an assertion failure for `Cnop` having no effect since it doesn't return anything useful; such a primitive is not useful as far as the compiler can tell :-). I don't remember seeing the error with `Cpoll`. So it should probably be put back to what it way it was.

kayceesrk: Fixed.

ctk21:
I think this is safe in the sense it will not change any code which doesn't use the `nop`. I also think it will insert a `nop` when the user asks for it in a way that (hopefully) the compiler doesn't optimize away or move about. 

Merging!

PR #390, opened 2020-08-20
Initial implementation of Mutexes and Condition Variables

kayceesrk:


runtime/domain.c
```diff
@@ -1481,3 +1481,160 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+  wrapper = caml_alloc_custom(&caml_mutex_ops, sizeof(caml_plat_mutex*), 0, 1);
+  Mutex_val(wrapper) = m;
+  return wrapper;
+}
+
+CAMLprim value caml_mutex_lock (value wrapper)
+{
+  CAMLparam1(wrapper); /* prevent deallocation of wrapper and destruction of
+                          mutex */
+  caml_plat_mutex* m = Mutex_val(wrapper);
```
ctk21: We should move the roots and wrapper to the pattern used here:
 https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/otherlibs/systhreads/st_stubs.c#L738-L744
which utilizes try_lock as a fast path for an uncontended lock.

runtime/domain.c
```diff
@@ -1481,3 +1481,160 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+  caml_plat_lock(m);
+  caml_leave_blocking_section();
+  CAMLreturn (Val_unit);
+}
+
+CAMLprim value caml_mutex_unlock (value wrapper)
+{
+  CAMLparam1(wrapper);
+  caml_plat_mutex* m = Mutex_val(wrapper);
+  caml_enter_blocking_section();
```
ctk21: why do we need the blocking section for unlocking the mutex?
do we need to do the local roots for wrapper?

runtime/domain.c
```diff
@@ -1481,3 +1481,160 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+  caml_plat_unlock(m);
+  caml_leave_blocking_section();
+  CAMLreturn (Val_unit);
+}
+
+CAMLprim value caml_mutex_try_lock (value wrapper)
+{
+  CAMLparam1(wrapper);
+  caml_plat_mutex* m = Mutex_val(wrapper);
+  caml_enter_blocking_section();
```
ctk21: I don't think we need the blocking section or the local roots for try_lock

runtime/domain.c
```diff
@@ -1481,3 +1481,160 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+  caml_plat_unlock(m);
+  caml_leave_blocking_section();
+  CAMLreturn (Val_unit);
+}
+
+CAMLprim value caml_mutex_try_lock (value wrapper)
```
ctk21, line 1554: This function has a bug. The return value of `caml_plat_try_lock` is not being passed back.

runtime/domain.c
```diff
@@ -1481,3 +1481,160 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+    caml_alloc_custom(&caml_condition_ops, sizeof(caml_plat_cond*), 0, 1);
+  Condition_val(wrapper) = c;
+  CAMLreturn (wrapper);
+}
+
+CAMLprim value caml_condition_signal (value wrapper)
+{
+  CAMLparam1(wrapper);
+  caml_plat_cond* c = Condition_val(wrapper);
+  caml_enter_blocking_section();
```
ctk21: Do we need the blocking section and the local root protection to call the signal?

runtime/domain.c
```diff
@@ -1481,3 +1481,160 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+  caml_plat_signal(c);
+  caml_leave_blocking_section();
+  CAMLreturn (Val_unit);
+}
+
+CAMLprim value caml_condition_broadcast (value wrapper)
+{
+  CAMLparam1(wrapper);
+  caml_plat_cond* c = Condition_val(wrapper);
+  caml_enter_blocking_section();
```
ctk21: Do we need the blocking section and the local root protection to call broadcast?

stdlib/domain.ml
```diff
@@ -42,6 +42,22 @@ module Sync = struct
...
+  external unlock : t -> unit = "caml_mutex_unlock"
+  external try_lock : t -> unit = "caml_mutex_try_lock"
+end
+
+module Condition = struct
+  type t
+  external create : Mutex.t -> t = "caml_condition_new"
+  external wait : t -> unit = "caml_condition_wait"
+  external broadcast : t -> unit = "caml_condition_broadcast"
+  external signal : t -> unit = "caml_condition_signal"
```
ctk21: can `signal` and `broadcast` be `noalloc`?

stdlib/domain.ml
```diff
@@ -42,6 +42,22 @@ module Sync = struct
   external poll : unit -> unit = "%poll"
 end
 
+module Mutex = struct
+  type t
+  external create : unit -> t = "caml_mutex_new"
+  external lock : t -> unit = "caml_mutex_lock"
+  external unlock : t -> unit = "caml_mutex_unlock"
+  external try_lock : t -> unit = "caml_mutex_try_lock"
```
ctk21: can `try_lock` and `unlock` be `noalloc`

stdlib/domain.mli
```diff
@@ -111,6 +111,22 @@ module Sync : sig
...
 
+module Mutex : sig
+  type t
+  val create : unit -> t
+  val lock : t -> unit
+  val unlock : t -> unit
+  val try_lock : t -> unit
+end
+
+module Condition : sig
```
ctk21, line 122: This interface is different for `create` and `wait` vs the systhread `Condition` module. I guess we need to make a call at some point which will be the final one.

ctk21 (commented):
I know this PR is a draft, but I will be on vacation for a bit, so figured some feedback now is better than nothing for a while. 

I don't think it matters that this is not unified with the `Mutex` and `Condition` modules as it is in its own namespace. We can experimentally put this in and understand the systhread overhead. 

One thing I'm not sure about is if the symbols (e.g. `caml_mutex_lock`, etc) being the same as those expected by the existing modules (`Mutex` and `Condition`) will cause problems. Particularly for #381 and co-existing with systhreads.

runtime/domain.c
kayceesrk: done.

runtime/domain.c
kayceesrk: fixed.

runtime/domain.c
kayceesrk: fixed.

runtime/domain.c
kayceesrk: fixed.

stdlib/domain.ml
kayceesrk: yes.

runtime/domain.c
kayceesrk: fixed.

stdlib/domain.ml
kayceesrk: yes,

ctk21:
Closing this PR as we have implemented and merged it in #548

PR #394, opened 2020-08-30
Changes to polling placement

sadiqj:
This is a draft PR which changes the polling placement. Instead of Feely's algorithm we now place polls in at most the start of functions (after the prologue) and on the back-edge of loops.

I've created a PR so we can do some more testing on this.

### DIFF
--- a/.depend
+++ b/.depend
@@ -2534,14 +2534,10 @@ asmcomp/mach.cmi : \
 asmcomp/polling.cmo : \
     asmcomp/reg.cmi \
     asmcomp/mach.cmi \
-    lambda/debuginfo.cmi \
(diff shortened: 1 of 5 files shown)


kayceesrk:
The CI passes. Next thing we should do is run sandmark on this.

kayceesrk:
One suggestion I would have is to keep this as a draft PR until the changes are merged upstream. I would think that when this PR is submitted upstream there will be several changes requested. It would be tedious to keep both in sync. So let's wait until the upstream PR is merged and then port that onto multicore.

sadiqj:
This is now stale. I'll follow with an updated one once it's had some more reviewing.

PR #400, opened 2020-09-21
Demux eventlog for backup thread

kayceesrk:
With this change, in the eventlog, every even-numbered process is the domain's main thread and the subsequent odd-numbered process is the backup thread. Earlier the events from the backup thread were emitted with the same process id as the main thread and this lead to incomprehensible logs. 

Here is a sample run with this change:

image

In the above, the backup threads are active when the main thread is waiting on a condition variable.


ctk21:
Good idea. Implementation looks good to me.

PR #401, opened 2020-09-21
Do not handle interrupts recursively

kayceesrk:
Introduces a domain local variable to prevent handling interrupts recursively.


kayceesrk:
This is not quite right as `caml_do_final_calls` may execute arbitrary OCaml code and trigger a GC.

ctk21:
This PR is being closed in favour of #402

PR #404, opened 2020-09-29
Domain.DLS.new_key takes an initialiser

### LINKED ISSUE #388: Domain.TLS.new_key should take an initializer
`Domain.TLS.new_key` is currently `unit -> 'a key`, which is implemented as 

```ocaml
type 'a key = int ref
```

`Domain.TLS.get : 'a key -> 'a option` returns `None` if the key hasn't been previously set on this domain. Almost all the time, if the key is not set on a particular domain, then the thing to do is to initialise the key on that domain. For example, here is the code snippet from [`LU_decomposition` benchmark in sandmark](https://github.com/ocaml-bench/sandmark/blob/master/benchmarks/multicore-numerical/LU_decomposition_multicore.ml#L7-L11):

```ocaml
let get_state () = try Option.get @@ Domain.DLS.get k with _ ->
  begin
    Domain.DLS.set k (Random.State.make_self_init ());
    Option.get @@ Domain.DLS.get k
  end
```

Instead of this API, it would be better to take the intializer as an argument during key creation as in Java: https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/ThreadLocal.html. Now the new API would be

```ocaml
type 'a key = int ref * (unit -> 'a) (* initialiser *)
val new_key : (unit -> 'a) -> 'a key
val get : 'a key -> 'a
val set : 'a key -> 'a -> unit
```

`get` can always return a value since either it is the case that the key has been already set on this domain, or the initialiser is used to set the key on this domain.

Sudha247:
This PR adds an initialiser argument to `Domain.DLS.new_key`. The initialiser used to set an associated value to that key in case no other value is explicitly set to it. `Domains.DLS.get` no longer returns an option value. For example:

```ocaml
# let k = Domain.DLS.new_key (fun () -> 42);;
val k : int Domain.DLS.key = <abstr>
# Domain.DLS.get k;;
- : int = 42
```
Due to changes in type signatures of `DLS.new_key` and `DLS.get` in this patch, any existing code using `DLS` module needs to be altered to use the modified API if this PR is merged.

Fixes #388

### DIFF
--- a/stdlib/domain.ml
+++ b/stdlib/domain.ml
@@ -144,7 +144,7 @@ module DLS = struct
     | None -> ()
     | Some e -> set_dls_list (e::vals)
 
-    let get k =
+  let get k =
     let vals = get_dls_list () in
     let (key, init) = k in
     let rec search k l =
--- a/stdlib/domain.ml
+++ b/stdlib/domain.ml
@@ -116,7 +116,7 @@ module DLS = struct
 
   type 'a key = int ref * (unit -> 'a)
 
-  type entry = {key: int ref; slot: Obj.t ref}
+  type entry = {key_id: int ref; slot: Obj.t ref}
 
   external get_dls_list : unit -> entry list
     = "caml_domain_dls_get" [@@noalloc]
@@ -124,44 +124,38 @@ module DLS = struct
   external set_dls_list : entry list  -> unit
     = "caml_domain_dls_set" [@@noalloc]
 
-  let new_key f =
-    let k = ref 0 in
-    let vals = get_dls_list () in
-    let e = {key = k; slot = ref (Obj.repr (f ())) } in
-    set_dls_list (e::vals);
-    (k, f)
+  let new_key f = (ref 0, f)
 
   let set k x =
     let cs = Obj.repr x in
     let vals = get_dls_list () in
     let rec add_or_update_entry k v l =
       match l with
-      | [] -> Some {key = k; slot = ref v}
+      | [] -> Some {key_id = k; slot = ref v}
       | hd::tl ->
-        if (hd.key == k) then begin
+        if (hd.key_id == k) then begin
           hd.slot := v;
           None
         end
         else add_or_update_entry k v tl
     in
-    let k = fst k in
-    match add_or_update_entry k cs vals with
+    let (key, _) = k in
+    match add_or_update_entry key cs vals with
     | None -> ()
     | Some e -> set_dls_list (e::vals)
 
     let get k =
     let vals = get_dls_list () in
-    let key = fst k in
-    let init = snd k in
+    let (key, init) = k in
     let rec search k l =
       match l with
       | [] ->
         begin
-          let e = {key = k; slot = ref (Obj.repr (init ())) } in
+          let e = {key_id = k; slot = ref (Obj.repr (init ())) } in
           set_dls_list (e::vals);
           !(e.slot)
         end
-      | hd::tl -> if hd.key == k then !(hd.slot) else search k tl
+      | hd::tl -> if hd.key_id == k then !(hd.slot) else search k tl
     in
     Obj.obj @@ search key vals
 
--- a/stdlib/domain.ml
+++ b/stdlib/domain.ml
@@ -114,7 +114,7 @@ let self () = Raw.self ()
 
 module DLS = struct
 
-  type 'a key = int ref
+  type 'a key = int ref * (unit -> 'a)
 
   type entry = {key: int ref; slot: Obj.t ref}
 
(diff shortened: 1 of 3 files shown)


stdlib/domain.ml
```diff
@@ -124,30 +124,45 @@ module DLS = struct
...
+    let k = fst k in
+    match add_or_update_entry k cs vals with
     | None -> ()
-    | Some e -> set_dls_list (e::old)
+    | Some e -> set_dls_list (e::vals)
 
-  let get k =
+    let get k =
     let vals = get_dls_list () in
+    let key = fst k in
```
ctk21: Might this be better:
```ocaml
let (key, init) = k in 
```

stdlib/domain.ml
```diff
@@ -124,30 +124,45 @@ module DLS = struct
...
-      end else add_or_update_entry k v tl
+      | hd::tl ->
+        if (hd.key == k) then begin
+          hd.slot := v;
+          None
+        end
+        else add_or_update_entry k v tl
     in
-    match add_or_update_entry k cs old with
+    let k = fst k in
```
ctk21: I found this function confusing to follow, because the variable `k` gets rebound from being an `int ref * (unit -> 'a)` to a `int ref`. If you read the function from top to bottom, the `k` coming into the function is not the `k` that is used in `add_or_update_entry`. Can we rename to match what is happening in `get`?

stdlib/domain.ml
```diff
@@ -114,7 +114,7 @@ let self () = Raw.self ()
 
 module DLS = struct
 
-  type 'a key = int ref
+  type 'a key = int ref * (unit -> 'a)
 
   type entry = {key: int ref; slot: Obj.t ref}
```
ctk21: This is a little confusing to me now that we've made `'a key` an `int ref * (unit -> 'a)`
Might it be better to call the `int ref` that is part of the `entry` record the `key_id` or `id`?

ctk21 (commented):
There are two things here that I'm not sure about:
- `new_key` is calling the intitializer and putting the value into the DLS list. I would expect it to just bind the initializer to the key value. Subsequent `get` calls would actually do the intitialization. Consider the following use case:
 ```ocaml
   let k = DLS.new_key very_expensive_fn in
   let domain_work () = DLS.get k in
   let domains = Array.init 3 (fun _ -> Domain.spawn(domain_work)) in
   Array.iter Domain.join domains
```
  This snippet should compute the `very_expensive_fn` in each spawned domain as part of the `DLS.get k`. The current PR is that it will compute it in `new_key` *and* in each spawned domain.

- if we land this PR we are going to break existing code using `DLS.get` as the signature will change. We might consciously want to do that; Sandmark may be the only user and we can update that, but we should state we are doing that.

**Comment by Sudha247** [comment]
>This snippet should compute the very_expensive_fn in each spawned domain as part of the DLS.get k. The current PR is that it will compute it in new_key and in each spawned domain.

I did consider not doing the actual initialisation in the `new_key` method, but decided to go with doing it since it might reduce number of operations in `get` for the calling domain. Having thought about it again, I think you are right about deferring the initialisation until `get` is explicitly called. Fixed now.

>if we land this PR we are going to break existing code using DLS.get as the signature will change. We might consciously want to do that; Sandmark may be the only user and we can update that, but we should state we are doing that.

This PR will indeed break existing usage of `DLS.new_key` and `DLS.get` in Sandmark and possibly domainslib, I have edited the original message to add this information. Thanks for pointing this out. I can submit patches to fix them if this PR lands.

File: stdlib/domain.ml (hunk 2 above)
Comment (Sudha247) [reply stdlib/domain.ml]: done

File: stdlib/domain.ml [hunk 4]
```diff
@@ -114,40 +114,49 @@ let self () = Raw.self ()
 
 module DLS = struct
 
-  type 'a key = int ref
+  type 'a key = int ref * (unit -> 'a)
 
-  type entry = {key: int ref; slot: Obj.t ref}
+  type entry = {key_id: int ref; slot: Obj.t ref}
```
kayceesrk: Perhaps I've asked this earlier, is there a reason why the type isn't

```ocaml
type entry = {key_id: int ref; mutable slot: Obj.t}
```

stdlib/domain.ml
```diff
@@ -114,40 +114,49 @@ let self () = Raw.self ()
...
   let get k =
     let vals = get_dls_list () in
+    let (key, init) = k in
     let rec search k l =
       match l with
-      | [] -> None
-      | hd::tl -> if hd.key == k then Some !(hd.slot) else search k tl
+      | [] ->
+        begin
+          let e = {key_id = k; slot = ref (Obj.repr (init ())) } in
```
kayceesrk: Better to do

```ocaml
let r = init () in
let e = {key_id = k; slot = ref (Obj.repr r) } in
set_dls_list (e::vals);
r
```

as it avoids 2 memory reads. OCaml doesn't forward ref reads to optimise the code.

stdlib/domain.ml
Sudha247: `mutable` seems cleaner than `ref`. Changed this.

stdlib/domain.ml
Sudha247: done

kayceesrk:
I'm happy with the changes. @ctk21 feel free to merge if you are satisfied.

kayceesrk (approved):

ctk21:
This looks good to me. 

Given that it is a breaking change. Can we have the PRs for sandmark and domainslib lined up? 
That way we can merge this and then the dependent pieces together.

kayceesrk:
Good idea. @Sudha247 would you be able to make the PRs for sandmark and domainslib?

Sudha247:
Sure, done. They are https://github.com/ocaml-bench/sandmark/pull/178 and https://github.com/ocaml-multicore/domainslib/pull/20.

PR #410, opened 2020-10-08
systhreads: caml_c_thread_register and caml_c_thread_unregister

abbysmal:
This PR reimport the `caml_c_thread_register` and `caml_c_thread_unregister` functions to systhreads.

When registering a C thread:
- The thread needs to setup Caml_state, to be tied to a domain.
  This is also how the regular systhreads behave, modulo the fact that the domain information is transmitted to them through the thread descriptor, passed to the thread's startup function. We do not have that luxury here and decide to register the thread to the domain 0.
- The thread can then properly access the right systhreads table (domain 0's) and check for its existence
- if it doesn't exist, then the thread can setup its own entry, and wait for the domain lock to become available to allocate the heap-allocated thread descriptor.

I think the locking logic is sane, and I did manage to successfully run luv's examples.
There's may be some weird interactions if domain 0 is actually busy doing something else, but I think requesting the domain lock and accessing the threads main locks prevent such problems.


kayceesrk:
LGTM. Are there no tests in the testsuite that tests these features?

abbysmal:
I just added a few tests in the testsuite to cover some basic usage of these functions, in both single and multi domains usecases.
I addressed a bug that triggered very simply on the many domain testcase, the locking logic did not cover the allocation of a root that could trigger a bug if the backup thread were to execute a GC before the thread could finish its registration.

abbysmal:
Ha, the CI went red because the include for `caml/thread.h` didn't go through.
I removed the `caml/` bit since the systhreads directory is included in the compiler call anyway.
It is not perfectly clean but I don't think any caml headers not originating from the runtime are ever used in the testsuite, so there's no specific support for this in ocamltest as far as I can see.
thanks @ctk21 for finding the build error. :)

PR #413, opened 2020-10-13
Replace the non-standard PTHREAD_MUTEX_ERRORCHECK_NP and fix a leak

### LINKED ISSUE #266: 'PTHREAD_MUTEX_ERRORCHECK_NP' undeclared
Hi, i'm trying to build ocaml-multicore on alpine linux, which relies 
on musl libc rather than the glibc. 

```

# gcc -c -O2 -fno-strict-aliasing -fwrapv -Wall -fno-tree-vrp  -g -D_FILE$
# gcc -c -O2 -fno-strict-aliasing -fwrapv -Wall -fno-tree-vrp  -g -D_FILE$
# gcc -c -O2 -fno-strict-aliasing -fwrapv -Wall -fno-tree-vrp  -g -D_FILE$
# platform.c: In function 'caml_plat_mutex_init':
# platform.c:25:34: error: 'PTHREAD_MUTEX_ERRORCHECK_NP' undeclared (firs$
#    pthread_mutexattr_settype(&ma, PTHREAD_MUTEX_ERRORCHECK_NP);
#                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
#                                   PTHREAD_MUTEX_ERRORCHECK

```
Understandably gcc complains it cannot find the non-portable version
`PTHREAD_MUTEX_ERRORCHECK_NP`

I understand what `PTHREAD_MUTEX_ERRORCHECK` does ([Posix manual](http://man7.org/linux/man-pages/man3/pthread_mutex_lock.3p.html))

But I am not sure about `PTHREAD_MUTEX_ERRORCHECK_NP`.
I Couldn't find much information... 
I tried to have a look into glibc source, 
found something about not doing lock elision, ...

Could someone help me understand what `PTHREAD_MUTEX_ERRORCHECK_NP` does?

Also, in `platform.c`, should I simply discard 
`#define _GNU_SOURCE /* for PTHREAD_MUTEX_ERRORCHECK_NP */`
and use the standard `PTHREAD_MUTEX_ERRORCHECK` ?

I was able to build successfully doing so but I suspect the `_NP` was here for a reason...

gadmm:
Adapting code from ocaml trunk, this does two things:
1) destroy the mutexattr
2) replace the older and non-standard `PTHREAD_MUTEX_ERRORCHECK_NP` with the standard `PTHREAD_MUTEX_ERRORCHECK`. This is reported as necessary for compiling multicore with musl (fixes #266). According to @stedolan [“This function was part of SUSv2 in 1997, and was included in POSIX.1-2001 as part of the XSI extension. The recent change is that it was made mandatory rather than part of an extension, but I'm not aware of any pthread implementations that lack it.”](https://github.com/ocaml/ocaml/pull/9757#issuecomment-658820515)

### DIFF
--- a/runtime/caml/platform.h
+++ b/runtime/caml/platform.h
@@ -2,9 +2,6 @@
 #define CAML_PLAT_THREADS_H
 /* Platform-specific concurrency and memory primitives */
 
-#ifdef __linux__
-#define _GNU_SOURCE /* for PTHREAD_ERRORCHECK_MUTEX_INITIALIZER_NP */
-#endif
 #include <pthread.h>
 #include <errno.h>
 #include <string.h>
--- a/runtime/platform.c
+++ b/runtime/platform.c
@@ -1,6 +1,5 @@
 #define CAML_INTERNALS
 
-#define _GNU_SOURCE /* for PTHREAD_MUTEX_ERRORCHECK_NP */
 #include <sys/mman.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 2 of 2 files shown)


kayceesrk:
Thanks @gadmm. Looks good to me.

PR #420, opened 2020-10-22
Implement fiber stack cache

kayceesrk:
This PR implements a stack cache for fibers. The cache initially starts out empty. When fibers are freed, they're returned to the pool rather than calling `free` on them. Such pooled fibers are reused for later allocations. We use a small number of allocation size classes; larger ones are returned to the OS eagerly. 

Here are some initial performance observations. The test program is:

```ocaml
effect E : unit

let n = try int_of_string Sys.argv.(1) with _ -> 10

let rec fib n =
  match
    if n < 2 then 1
    else fib (n-1) + fib (n-2)
  with
  | v -> v
  | effect E k -> failwith "impossible"

let main () = ignore @@ fib n

let _ = main ()
```

I have also benchmarked with `mimalloc` just to see if a difference `malloc` implementation makes a difference. Here are the results from 10 runs obtained using `hyperfine`. 

| Variant | Time | SD | Min | Max |
|-------|-----|------|------|-----|
| 4.10+mc | 1.053s | 0.033s | 1.017s | 1.112s |
| 4.10+mc+mimalloc | 0.914s | 0.041s | 0.850s | 0.969s |
| 4.10+mc+cache | 0.986s | 0.037s | 0.921s | 1.044s |
| 4.10+mc+cache+mimalloc | 0.963s | 0.035s | 0.912s | 1.031s |

Plain `4.10+mc+mimalloc` is the clear winner here. I'm surprised it does better than the stack cache.

### DIFF
--- a/runtime/caml/config.h
+++ b/runtime/caml/config.h
@@ -190,7 +190,7 @@ typedef uint64_t uintnat;
 
 /* Number of words used in the control structure at the start of a stack
    (see fiber.h) */
-#define Stack_ctx_words 3
+#define Stack_ctx_words 4
 
 /* Default maximum size of the stack (words). */
 #define Max_stack_def (1024 * 1024)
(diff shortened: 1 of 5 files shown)


kayceesrk:
Closing this one in favour of #442.

Issue #398, opened 2020-09-20
Domain-local allocation buffers - step 1: unified minor heap


sadiqj:
Multicore currently has a separate minor heap per domain, a vestige of the concurrent minor collector. This issue is to get the details down for a unified minor heap, which is a prerequisite for domain-local allocation buffers.

The intention is that on startup the specified minor heap size will be pre-allocated for `Max_domains`.

When we create a new domain, if it doesn't already have an assigned range in the unified minor heap (which we can identify by looking at something in `Caml_state` left over from a previous domain), then we allocate it one.

To avoid having minor heap sitting around from Domains that are no longer required, we may want to reallocate all currently running Domains minor heaps at the end of a major cycle and then `MADV_DONTNEED` the rest of the reserved space.

@Engil @ctk21

kayceesrk:
Closing this one as there is a negative performance impact due to this change. See PR https://github.com/ocaml-multicore/ocaml-multicore/pull/508 and the summary of the work done https://github.com/ocaml-multicore/ocaml-multicore/wiki/Domain-Local-Allocation-Buffers-Addendum.

PR #422, opened 2020-10-22
Simplify minor heaps configuration logic and masking

abbysmal:
This PR is a first step toward Domain local allocation buffers. (see also #398).
This first step aimed to rewrite the previous bit-twiddling logic of `Minor_heap_sel_bits` and `Minor_heap_align_bits` with a scheme closer to trunk.
The commits are split by duty, starting with the first one is likely the best.

The first commit introduce `Minor_heap_max`, and used this potential maximum size to reserve the minor heaps area.
The minor heaps area is reserved for Max_domains (so `Max_domains * Minor_heap_max`).
Further refactoring is done by removing `Is_minor` and only relying on `Is_young`, which is now a simple bound check (`minor_heaps_base < v < minor_heaps_end`). In parallel minor I believe such distinction is not required anymore.
This specific transition is done in the second commit however.

The third commit introduces a way to override the `Minor_heap_max` setting using `OCAMLRUNPARAM=s`.
This may not be the prettiest solution, and does not align with trunk (`Minor_heap_max` should not be exceeded by the user-configured minor heap.)

Sandmark run: 

minor_heap_re2

minor_heap_re1

### DIFF
--- a/runtime/caml/config.h
+++ b/runtime/caml/config.h
@@ -212,18 +212,10 @@ typedef uint64_t uintnat;
    This must be at least [Max_young_wosize + 1]. */
 #define Minor_heap_min (Max_young_wosize + 1)
 
-/* There may be at most 1<<Minor_heap_sel_bits minor
-   heaps allocated */
-#define Minor_heap_sel_bits 7
-
-/* An entire minor heap must fit inside one region
-   of size 1 << Minor_heap_align_bits, which determines
-   the maximum size of the heap */
-#if SIZEOF_PTR  <= 4
-#define Minor_heap_align_bits 20
-#else
-#define Minor_heap_align_bits 24
-#endif
+/* Maximum size of the minor zone (words).
+   Must be greater than or equal to [Minor_heap_min].
+*/
+#define Minor_heap_max (1 << 28)
 
 /* Default size of the minor zone. (words)  */
 #define Minor_heap_def 262144
@@ -264,7 +256,7 @@ typedef uint64_t uintnat;
 #define Percent_to_promote_with_GC 10
 
 /* Maximum number of domains */
-#define Max_domains (1 << Minor_heap_sel_bits)
+#define Max_domains 128
 
 /* Default setting for the major GC slice smoothing window: 1
    (i.e. no smoothing)
--- a/runtime/caml/mlvalues.h
+++ b/runtime/caml/mlvalues.h
@@ -204,31 +204,14 @@ bits  63        (64-P) (63-P)        10 9     8 7   0
 /* Fields are numbered from 0. */
 #define Field(x, i) (((value *)(x)) [i])           /* Also an l-value. */
 
-/* All values which are not blocks in the current domain's minor heap
-   differ from caml_domain_state in at least one of the bits set in
-   Young_val_bitmask */
-#define Young_val_bitmask \
-  ((uintnat)1 | ~(((uintnat)1 << Minor_heap_align_bits) - (uintnat)1))
+/* Is_young(val) is true iff val is in the reserved area for minor heaps */
 
-/* All values which are not blocks in any domain's minor heap differ
-   from caml_domain_state in at least one of the bits set in
-   Minor_val_bitmask */
-#define Minor_val_bitmask \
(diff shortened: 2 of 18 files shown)


runtime/caml/config.h
```diff
@@ -206,17 +206,10 @@ typedef uint64_t uintnat;
...
 #define Minor_heap_min (Max_young_wosize + 1)
 
-/* There may be at most 1<<Minor_heap_sel_bits minor
-   heaps allocated */
-#define Minor_heap_sel_bits 7
-
-/* An entire minor heap must fit inside one region
-   of size 1 << Minor_heap_align_bits, which determines
-   the maximum size of the heap */
 #if SIZEOF_PTR  <= 4
```
ctk21: Trunk has
```C
/* Maximum size of the minor zone (words).
   Must be greater than or equal to [Minor_heap_min].
*/
#define Minor_heap_max (1 << 28)
```
`(1<<28)` words is ~2GB per minor heap on a 64bit machine, I think we should just take that.

runtime/caml/mlvalues.h
```diff
@@ -204,31 +204,13 @@ bits  63        (64-P) (63-P)        10 9     8 7   0
...
-   via bitmasking. */
-#define Is_young(val) \
-  ((((uintnat)(val) ^ (uintnat)Caml_state) & Young_val_bitmask) == 0)
+/* Is_young(val) is true iff val is in the reserved area for minor heaps */
 
-/* Is_minor(val) is true iff val is a block in any domain's minor heap. */
-#define Is_minor(val) \
-  ((((uintnat)(val) ^ (uintnat)Caml_state) & Minor_val_bitmask) == 0)
+#define Is_young(val) \
+  ((char *)(val) < (char *)caml_minor_heaps_end && \
```
ctk21: Can we add `CAMLassert (Is_block (val))` to the assert?
This will mirror ocaml/ocaml:
```C
#define Is_young(val) \
  (CAMLassert (Is_block (val)), \
   (char *)(val) < (char *)Caml_state_field(young_end) && \
   (char *)(val) > (char *)Caml_state_field(young_start))
```

runtime/caml/mlvalues.h
```diff
@@ -204,31 +204,13 @@ bits  63        (64-P) (63-P)        10 9     8 7   0
...
+   (char *)(val) > (char *)caml_minor_heaps_base)
 
 #define Is_block_and_young(val) Is_young(val)
```
ctk21: This macro needs updating to:
`Is_block(val) && Is_young(val)`
previously with the bit twiddling `Is_young(val)` would also tell you if val was a block, but with the lo/hi we need to explicitly check the block as well.

ctk21 (changes requested):
There is a bug with `Is_block_and_is_young` that will need fixing. 

Having looked at the trunk setting for `Minor_heap_max` which is `(1 << 28)`; that is 268M words or 2GB per domain minor heap on a 64bit platform. I'm wondering if we need the commit that allows overriding of the minor heap limit with `OCAMLRUNPARAM`: 
https://github.com/ocaml-multicore/ocaml-multicore/pull/422/commits/949479687a34f8bcad5a887d7ffc64027788e4f9

Wouldn't it be easier to drop that override commit and settle with `(1 << 28)`? I'm thinking if people want to experiment with larger than `(1<<28)` they can change that setting in the header fairly easily and recompile.

Dropping the override commit, it makes us a bit closer to trunk.  I'm in favour of dropping the override commit.

ctk21 (approved):

abbysmal:
I think this PR is ready to be merged if everyone agrees with it.
I will need to squash it before though, this history is needlessly messy.

Here are the current results, comparing a few different version (stock, stock+256M, multicore, multicore+default (this patch + default minor heap size), and multicore+256M.)
There is now obvious slowdown, and I will gather notes and notebooks to discuss further the entry cost of bigger minor heaps in multicore in another issue.

## Normalized (stock)
sequential_normalized_stock

## Non normalized
sequential_defaut_stock

ctk21:
I think this one is good to go. 
You mention wanting to squash the history. There is also a merge conflict that needs patching up.

PR #426, opened 2020-10-25
Replace global roots implementation

sadiqj:
This PR replaces the existing global roots implementation with one that incorporates most of trunk OCaml's `globroots`.

There's further work to be done on global roots.

1) This implementation places locks around the skip lists. We should have a per-domain skip lists and a global one for orphans when a domain is terminated. This would eliminate the need to lock when registering/removing a global root.

2) We should remove the existing `Caml_root` usage and them remove these functions from `globroots`. This would leave us in a position where the global roots implementation very closely matches trunk.

### DIFF
--- a/runtime/globroots.c
+++ b/runtime/globroots.c
@@ -17,260 +17,166 @@
 
 /* Registration of global memory roots */
 
-#include "caml/memory.h"
-#include "caml/misc.h"
 #include "caml/mlvalues.h"
 #include "caml/roots.h"
 #include "caml/globroots.h"
-#include "caml/callback.h"
-#include "caml/platform.h"
-#include "caml/alloc.h"
-#include "caml/shared_heap.h"
-#ifdef NATIVE_CODE
-#include "caml/stack.h"
-#endif
+#include "caml/skiplist.h"
 
-/* A caml_root is in fact a value. We don't expose that fact outside
-   of this file so that C code doesn't attempt to directly modify it.
-   The value points to a block on the shared heap with the following
-   fields:
+/* The three global root lists.
+   Each is represented by a skip list with the key being the address
+   of the root.  (The associated data field is unused.) */
 
-    0: the actual root, as set by caml_modify_root
-    1: an integer flag stating whether this root has been deleted
-    2: the next root in roots_all
+struct skiplist caml_global_roots = SKIPLIST_STATIC_INITIALIZER;
+                  /* mutable roots, don't know whether old or young */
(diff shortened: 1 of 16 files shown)


runtime/globroots.c
```diff
@@ -17,134 +17,196 @@
...
+/* legacy multicore API that we need to fix */
 CAMLexport caml_root caml_create_root(value init)
 {
   CAMLparam1(init);
-  CAMLlocal1(v);
-  v = caml_alloc_shr(3, 0);
-  caml_initialize_field(v, 0, init);
-  caml_initialize_field(v, 1, Val_int(1));
+  
+  value* v = (value*)malloc(sizeof(value));
```
ctk21: Do we want any error handling here?

runtime/globroots.c
```diff
@@ -17,134 +17,196 @@
...
-  roots_all = v;
-  caml_plat_unlock(&roots_mutex);
+  *v = init;
+
+  caml_register_global_root(v);
 
   CAMLreturnT(caml_root, (caml_root)v);
 }
+
 CAMLexport caml_root caml_create_root_noexc(value init)
```
ctk21, line 42: Do we want to comment that the `caml_create_root_noexc` and `caml_create_root` versions are the same because we intend to remove?

ctk21 (commented):
It's great that we make the global roots code line up with ocaml/ocaml. What we are doing looks like it should work. 

Couple of things we should do:
 - can we also get the ocaml/ocaml testsuite code for gc-roots included in this PR? 
https://github.com/ocaml/ocaml/tree/4.10/testsuite/tests/gc-roots
 - given that there are a bunch of C include changes here, can you also run `make alldepend` and update the C `.depend` files?

runtime/globroots.c
sadiqj: This is a good point. I've switched from `malloc` to `caml_stat_alloc` and I've changed the `_noexc` version to return NULL correctly.

runtime/globroots.c
sadiqj: I've actually split them out now because they have different behaviour on allocation failure.

sadiqj:
Debugging the test failure on travis

sadiqj:
Right, this test was failing because I forgot to include the accompanying c file for the test from trunk. Included that and modified it to work with multicore's colours. Should pass here now.

ctk21:
Looks good, will merge!

Issue #399, opened 2020-09-20
Backport Garbage Collector Colours change


sadiqj:
Now that the GC colours change (https://github.com/ocaml/ocaml/pull/9756) has been merged to trunk, we should port the changes over to Multicore.

I think these are:
* the optimisations in `mark_stack_push`
* the representation of `mark_entry` (no longer storing end)
* marking behaviour

This is also a good opportunity to simplify by removing the existing `addrmap` implementation and something similar to https://github.com/ocaml/ocaml/pull/9756. We don't need skip lists in Multicore as the Pools in the allocator are all of fixed size.

The trunk PR also includes shrinking the mark_stack which may be something we want to consider.

@Sudha247 @kayceesrk

Sudha247:
Done in #427 and #435.

PR #427, opened 2020-10-26
Garbage Collector colours change backport

Sudha247:
This patch addresses some items from #399 back-porting changes from https://github.com/ocaml/ocaml/pull/9756. They are:

* `mark_entry` no longer stores `end`.
* `mark_stack_push` is closer to trunk.
* Imports marking behaviour from trunk, every field of a mark stack entry is iterated without creating a child entry.
* `caml_shrink_mark_stack` is adapted from trunk. On trunk, this is invoked during compaction. Since multicore doesn't yet have compaction, this is being invoked after mark stack is emptied. If there's a more appropriate place to do that, it could be moved there.

Of the items listed in #399 mark stack overflow needs to be imported from trunk, I plan to do that in a separate PR.

Sandmark results for this patch are below:

Time:
image

Normalised Time:
image

Major Collections:
image

Normalised Major Collections:
image

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -512,8 +512,8 @@ static void mark_stack_push(struct mark_stack* stk, value block,
 
   CAMLassert(Is_block(block) && !Is_minor(block));
   CAMLassert(Tag_val(block) != Infix_tag);
-  CAMLassert(Tag_val(block) != Cont_tag);
   CAMLassert(Tag_val(block) < No_scan_tag);
+  CAMLassert(Tag_val(block) != Cont_tag);
   /* Optimisation to avoid pushing small, unmarkable objects such as [Some 42]
    * into the mark stack. */
   end =  (block_wsz < 8 ? block_wsz : 8);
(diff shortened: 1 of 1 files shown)


runtime/major_gc.c
```diff
@@ -504,104 +503,149 @@ static void realloc_mark_stack (struct mark_stack* stk)
...
-            }
-          }
-        }
+      CAMLassert(Is_markable(me.block) &&
+                 Has_status_hd(Hd_val(me.block), global.MARKED) &&
+                 Tag_val(me.block) < No_scan_tag &&
+                 Tag_val(me.block) != Cont_tag);
+      int is_cont = 0;
+      mark_slice_darken(stk, me.block, me.offset++, &budget, &is_cont);
+      if(is_cont){
```
ctk21: I think there is something not right about this. If we have a block with a continuation then we stop marking the remaining fields in the block. 

The testcase in ocaml-multicore#430 segfaults with this branch and I think the issue might be here.

Sudha247: Fixed now, the testcase in #430 passes now on my machine. Perhaps we can kick off another CI run after #430 is merged.

kayceesrk:
What is the status of this PR? What needs to be done before merging this one?

Sudha247:
I think this PR is good to go, after a sanity check. There was some issue with how continuation tag was handled earlier, that has been fixed now. It was validated with the help of the test in #430.

ctk21:
I took a look at this, I didn't see anything jump out at me as wrong.

Issue #425, opened 2020-10-25
Issues with `Mutex.try_lock`


toots:
Hi!

I've followed the steps outlined by Anil [here](https://discuss.ocaml.org/t/multicore-ocaml-september-2020/6565) to install the multicore compiler and am running into an issue with this code:
```ocaml
let () =
  let m = Mutex.create () in
  Mutex.lock m;
  assert(not (Mutex.try_lock m))
```
After compiling and running it:
```shell
% ./test.exe
Fatal error: exception File "./test.ml", line 4, characters 2-8: Assertion failed
Raised at file "string.ml", line 115, characters 19-34
Called from file "string.ml", line 179, characters 15-34
zsh: exit 2     ./test.exe
```

Details:
* OCaml: `4.10.0 [ocaml-multicore-ocaml-multicore-no-effect-syntax]`

abbysmal:
Thank you very much for your report!

As a side note, I am unsure what was your usecase with it, but Mutex (as the module stands right now in multicore, when linked with systhreads), are expected to be used for cross-systhreads interactions.
If your intended usecase is cross-domains interactions, the following document will provide you with the current abstractions:
https://github.com/ocaml-multicore/parallel-programming-in-multicore-ocaml
For now the Mutex module is a part of systhreads, and while it *may* work accross domains (there is a few testcases in the testsuite), this is maybe not the better path. :)
We do have a work in progress toward enabling Mutex and Condition for cross-domains interactions, but it is not merged yet:
https://github.com/ocaml-multicore/ocaml-multicore/pull/390

PR #428, opened 2020-10-26
Fix Mutex.try_lock in systhreads.

abbysmal:
See #425
In what appears to be a terrible mistake, it seems I inverted the result of `Mutex.try_lock` during the systhreads port to multicore.
Indeed it would returns `false` on successfully locking, and `true` on being unable to lock.
Sorry for this mistake, as @ctk21 mentioned I think it is for the better if we actually add `try_lock` to the testsuite, in order to avoid such small mistakes.
I extracted the test program provided by @toots to use as our new testcase for `try_lock`.


otherlibs/systhreads/st_stubs.c
```diff
@@ -554,7 +554,7 @@ CAMLprim value caml_mutex_try_lock(value wrapper)           /* ML */
 {
   st_mutex mut = Mutex_val(wrapper);
   int retcode = st_mutex_trylock(mut);
-  if (retcode == 1) return Val_false;
+  if (retcode == 0) return Val_false;
```
ctk21: Is it better do something a bit clearer and use a define for the return code like:
https://github.com/ocaml/ocaml/blob/4.10/otherlibs/systhreads/st_stubs.c#L846

ctk21 (commented):
Can we also add the positive testcase for trylock?

That is, something along the lines of:
```ocaml
let () =
  let m = Mutex.create () in
  assert (Mutex.try_lock m);
  Mutex.unlock m
```

abbysmal:
Thank you for your comments, I just pushed a positive testcase for trylock.
I also updated and reused the defines for trylock return values to match more how trunk handles trylock's returned values.

toots:
Great, can't wait to test the fixed compiler!

ctk21:
LGTM - merging!

PR #433, opened 2020-11-05
Collect gc stats on domain termination

abbysmal:
Gc stats are collected at the end of a major cycle (when all participants go through `cycle_all_domains_callbacks`).
This means that if a domain terminates long before the next major cycle, the domain's latest allocations/promotion/else stats won't be accounted for.
While this may be a minor annoyance in a regular context, when the program does very little major gc (or none at all), this may cause various kind of inconsistencies in the results provided by the sampled stats.
The fix is simply to collect the statistics on domain termination.

I am willing to rename `caml_sample_gc_collect` in my patch, I think `collect` may be a bit of a confusing term. (at least in the overall GC terminology book.)
I also think the placement should be fine in the domain's termination lifecycle, but a second look wouldn't hurt.

### DIFF
--- a/runtime/caml/major_gc.h
+++ b/runtime/caml/major_gc.h
@@ -85,6 +85,7 @@ struct gc_stats {
   struct heap_stats major_heap;
 };
 void caml_sample_gc_stats(struct gc_stats* buf);
+void caml_sample_gc_collect(caml_domain_state *domain);
 
 /* Forces finalisation of all heap-allocated values,
    disregarding both local and global roots.
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1270,6 +1270,7 @@ static void domain_terminate()
     }
     caml_plat_unlock(&s->lock);
   }
+  caml_sample_gc_collect(domain_state);
 
   caml_stat_free(domain_state->final_info);
   caml_stat_free(domain_state->ephe_info);
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -854,6 +854,19 @@ void caml_sample_gc_stats(struct gc_stats* buf)
   buf->major_heap.large_max_words = large_max;
 }
 
+/* update GC stats for this given domain */
+inline void caml_sample_gc_collect(caml_domain_state* domain)
+{
+  int stats_phase = caml_major_cycles_completed & 1;
+  struct gc_stats* stats = &sampled_gc_stats[stats_phase][domain->id];
+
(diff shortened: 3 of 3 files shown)


ctk21:
I'm not convinced this PR works with domain reuse within the same major GC cycle. The case I'm worried about is:
- domain terminates and now writes to `sampled_gc_stats[stats_phase][domain->id]`
- domain starts, reusing the same domain id
- domain terminates and now writes to `sampled_gc_stats[stats_phase][domain->id]`, but this write clobbers the old values from the first write on termination above

I might have something wrong in my understanding, or is that sequence possible?

abbysmal:
Hello, thanks for your review!
I just double checked, `sampled_gc_stats` is technically as cumulative as the related `Caml_state->stat_` field is.
What is means is that it cannot be clobbered by a slot reuse, as the reused slot will carry on the stat data from the previous instance. (all allocated words and so on, these carry over when slots are reused.).
I just checked that and I think this scenario is fine, however if this is a bug or a feature is another question, it does sound fine to me!
Example of the `stat_minor_words` field on slot reuse (the first integer is the domain id, second is the word count for the minor heap):
```
●●●● OCAMLRUNPARAM='v=0x400' ./a.out
### OCaml runtime: debug mode ###
hello buddy: 1 0
hello buddy: 2 0
hello buddy: 3 3
hello buddy: 5 3
hello buddy: 4 3
hello buddy: 6 0
hello buddy: 7 0
hello buddy: 3 294919
hello buddy: 4 294919
hello buddy: 5 294919
hello buddy: 6 294922
hello buddy: 7 294919
hello buddy: 3 589838
hello buddy: 8 0
hello buddy: 4 589838
hello buddy: 5 589838
hello buddy: 3 884757
hello buddy: 6 589838
hello buddy: 3 1179676
hello buddy: 4 884760
hello buddy: 5 884757
hello buddy: 6 884757
hello buddy: 3 1474595
hello buddy: 4 1179676
hello buddy: 6 1179676
hello buddy: 5 1179676
hello buddy: 3 1769514
hello buddy: 3 2064433
hello buddy: 4 1474598
hello buddy: 5 1474595
hello buddy: 6 1474598
hello buddy: 3 2359355
hello buddy: 7 589838
hello buddy: 4 1769514
hello buddy: 5 1769514
hello buddy: 6 1769514
hello buddy: 3 2654271
*snip*
```

abbysmal:
Superseded by #446, closing

PR #435, opened 2020-11-09
Mark stack overflow backport

Sudha247:
This patch removes the existing mark stack overflow and adds an implementation that is closer to trunk OCaml. Three fourth of the mark stack is cleared, all the pools storing those removed entries are added to `pools_to_rescan`. Pools in `pools_to_rescan` are marked later during a major slice. The pools are added to a skiplist first to eliminate any duplicate values.

Sandmark results for macro-benchmarks are below:
image
image

Results of the micro-benchamrk `finalise` measuring performance difference in mark stack overflow:

image
image

Normalised major collections against `parallel_minor_gc` on this branch:
image

### DIFF
--- a/.depend
+++ b/.depend
@@ -1852,10 +1852,8 @@ bytecomp/printinstr.cmx : \
     bytecomp/printinstr.cmi
 bytecomp/printinstr.cmi : \
     bytecomp/instruct.cmi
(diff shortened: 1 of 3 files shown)


kayceesrk:
Thanks @Sudha247 and @sadiqj! Excellent debugging work.

Issue #409, opened 2020-10-05
caml_thread_scan_roots may be subject to race condition when two domains are initializing


abbysmal:
Quoting @ctk21 on #381 
```
I don't think this is going to be safe if two domains try to initialize at the same time:

   -  domain A starts initializing and gets caml_scan_roots_hook set to caml_dec_scan_roots, but gets suspended before the other hooks are initialized
   -  the prev_scan_roots_hook gets written with something unexpected

Could put in a mutex or there you could way with a CAS on prev_scan_roots_hook and be careful with the ordering of the operations.
```

ctk21:
Fixed by #439

PR #439, opened 2020-11-15
Systhread lifecycle rework

abbysmal:
After investigating something with #433, I triggered a series of bugs with systhreads when forcibly linked with the spawn_burn testcase.
This led me to a goose chase where I ended up having to rethink a bit more clearly the initialization situation for systhreads.
This PR is relatively cleanly split up on a per duty basis, and address the following issues:

- General resource handling and freeing up descriptors and stacks
- Addressing general concerns and bugs in how systhreads is initialized.

This PR also addresses #409, that shouldn't be an issue after this.
I recommend reading the associated commit messages, they contain all the information on what is being changed, and why.
There is one obvious change to the runtime, adding a new hook on domain termination (systhreads already adding a domain startup hook as well.). Reading the commit messages will clear up why, and there may be a better approach but this is my solution so far.
The testsuite now runs fine, and my initial testcase at the root of these issues also passes fine now.
General ecosystems tests did not show any regression. (with dune and softwares known to work with our systhreads implementation.)

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -477,6 +477,11 @@ static void install_backup_thread (dom_internal* di)
   }
 }
 
+static void caml_domain_stop_default(void)
+{
+  return;
+}
+
 static void caml_domain_start_default(void)
 {
   return;
@@ -484,6 +489,9 @@ static void caml_domain_start_default(void)
 
 CAMLexport void (*caml_domain_start_hook)(void) =
    caml_domain_start_default;
+
+CAMLexport void (*caml_domain_stop_hook)(void) =
+   caml_domain_stop_default;
 
 static void domain_terminate();
 
@@ -1225,6 +1233,8 @@ static void domain_terminate()
 
   caml_gc_log("Domain terminating");
   caml_ev_pause(EV_PAUSE_YIELD);
+  // Before wrapping up further, we run the termination hook
+  caml_domain_stop_hook();
   caml_delete_root(domain_state->dls_root);
   s->terminating = 1;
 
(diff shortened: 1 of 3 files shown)


runtime/domain.c
```diff
@@ -1225,6 +1233,8 @@ static void domain_terminate()
 
   caml_gc_log("Domain terminating");
   caml_ev_pause(EV_PAUSE_YIELD);
+  // Before wrapping up further, we run the termination hook
+  caml_domain_stop_hook();
```
ctk21: I'm a bit worried about the location of this being here, because the domain still has to fully participate in any major and minor GC cycles. 

Here's a collection of events that worry me:
 - a domain stops and this hook is called which will result in `caml_thread_domain_stop_hook` running which will set `Current_thread` to null.
 - a minor GC is triggered but we don't scan all the stacks in `caml_thread_scan_roots` because `Current_thread` is null.

I guess the bit I'm not sure about is if the domain can terminate while systhreads are still active and what that means.

ctk21 (commented):
I think this PR makes sense.

One bit I'm unsure about is the location in `domain_terminate` for the stop hook; with its current placement the GC can run after the hook. 

Another thing that might be useful, given the history of the bugs this fixes, is to add a systhread specific test which is similar to `domain_spawn_burn` but where the domains getting spawned create and schedules several systhreads that do something.

runtime/domain.c
abbysmal: Thank you for your review.
I think one my assumptions do not hold regarding potential _rogue_ systhreads still running on domain termination, and there may be larger implication regarding the behavior on stock OCaml vs how Multicore should handle it in a multiple domain setting.
The semantic is not quite clear in my head right now, I need to investigate a bit and I will come up with a proposal.
This PR is not super urgent as it only impact many-domains style systhreads, which should not prevent ecosystem testing.

ctk21:
We're going to merge. It fixes #409. 

There is going to be a follow up issue for how to deal with live systhreads if a domain should somehow exit (not an issue with a single domain)

PR #438, opened 2020-11-13
Allow C++ to use caml/camlatomic.h

kit-ty-kate:
Failure in `ubpf.0.1`:
```
- (cd _build/default/src && /usr/bin/gcc -O2 -fno-strict-aliasing -fwrapv -fPIC -std=c++11 -Iubpf_c/ubpf/vm/inc -g -I /home/opam/.opam/4.10.0+multicore+no-effect-syntax/lib/ocaml -I ubpf_c -o ubpf_stubs.o -c ubpf_stubs.cpp)
- In file included from /usr/include/c++/10/memory:85,
-                  from cxx_wrapped.h:37,
-                  from ubpf_stubs.cpp:10:
- /usr/include/c++/10/bits/shared_ptr_atomic.h:76:58: error: macro "atomic_is_lock_free" passed 2 arguments, but takes just 1
-    76 |     atomic_is_lock_free(const __shared_ptr<_Tp, _Lp>* __p)
-       |                                                          ^
- In file included from /home/opam/.opam/4.10.0+multicore+no-effect-syntax/lib/ocaml/caml/camlatomic.h:17,
-                  from /home/opam/.opam/4.10.0+multicore+no-effect-syntax/lib/ocaml/caml/misc.h:33,
-                  from /home/opam/.opam/4.10.0+multicore+no-effect-syntax/lib/ocaml/caml/mlvalues.h:23,
-                  from ubpf_stubs.cpp:4:
- /usr/lib/gcc/x86_64-linux-gnu/10/include/stdatomic.h:96: note: macro "atomic_is_lock_free" defined here
-    96 | #define atomic_is_lock_free(OBJ) __atomic_is_lock_free (sizeof (*(OBJ)), (OBJ))
-       | 
[...]
```

PR #444, opened 2020-11-24
Merge branch 'parallel_minor_gc' into 'no-effect-syntax'

kit-ty-kate:
I hope this is ok to request. It would be great to have #438 in the `no-effect-syntax` branch so that opam-health-check can test more packages.

I'm not seeing any change in the `parsing` directory so I'm guessing this is fine to merge.


kayceesrk:
Thanks Kate. The failing tests all have to do with `effect` syntax not being available. I've merged this one in.

PR #446, opened 2020-11-26
Collect GC stats at the end of minor collection and remove double buffering of gc sampled stats.

abbysmal:
This PR is an extension of #433.
The target is to remove the use of double buffering in GC stats collection, leveraging on the barrier present during minor collection in the parallel_minor_gc schema.
Double buffering could cause inconsistencies in stats, depending the cycle on which a program ends. (when using `OCAMLRUNPARAM=v=0x400`)
I tried went through reading the stat collection process and remap how it would behave with this patch and it seems like a much consistent setup. (@ctk21 maybe could spot more potential inconsistencies.)

I ran the sequential bench suite on it, here's the time result, normalize against stock OCaml.

sequential_minor_gc_sample_normalize_stock_time

I don't see any significant slowdown (noise, at best), I have some parallel results as well that do not show any further slowdown. 

Edit: the graph I showed was the wrong one, I updated it.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -780,12 +780,7 @@ intnat ephe_sweep (struct domain* d, intnat budget)
   return budget;
 }
 
-/* double-buffered sampled GC stats.
-   At the end of GC cycle N, domains update sampled_gc_stats[N&1],
-   but requests to Gc.stats() read from sampled_gc_stats[!(N&1)].
-   That way, Gc.stats() returns the statistics atomically sampled
-   at the end of the most recently completed GC cycle */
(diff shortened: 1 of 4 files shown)


ctk21:
LGTM.

runtime/major_gc.c
```diff
@@ -818,14 +813,13 @@ void caml_sample_gc_stats(struct gc_stats* buf)
   /* we read from the buffers that are not currently being
```
kayceesrk: This comment should be deleted.

kayceesrk (approved):
I've suggested a minor change to delete a stale comment. Otherwise, looks good.

kayceesrk:
LGTM.

PR #449, opened 2020-11-26
Fix stdatomic.h when used inside C++ for good

kit-ty-kate:
stdatomic.h is not compatible when included in C++ code. To fix this issue the atomic header has to be used instead, the two interfaces should be moderatly compatible. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60932 for more details. extern "C++" is here to prevent extern "C" used in user code to take over and invalidates the definitions inside the atomic header.

cc @dra27 do you think there would be any issue on Windows?

This fixes build failures for at least `upbf.0.1` and `libsvm.0.10.0` but I'm guessing this unlocks a substantial number of other projects too.

### DIFF
--- a/runtime/caml/camlatomic.h
+++ b/runtime/caml/camlatomic.h
@@ -14,14 +14,22 @@
  */
 
 #ifdef __cplusplus
-extern "C" {
-#endif
 
-#if defined(HAS_STDATOMIC_H)
+extern "C++" {
+#include <atomic>
+using namespace std;
+#define ATOMIC_UINTNAT_INIT(x) (x)
+typedef atomic<uintnat> atomic_uintnat;
+typedef atomic<intnat> atomic_intnat;
+}
+
+#elif defined(HAS_STDATOMIC_H)
+
 #include <stdatomic.h>
 #define ATOMIC_UINTNAT_INIT(x) (x)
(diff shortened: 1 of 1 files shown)


dra27:
No need to worry about Windows at the moment - there'll be a story, but there's much more to do there than worrying about C++ support 🙂 MSVC's has C++11 <atomic> but _not_ C11's `stdatomic.h`

I'm not sure how much we worry now, but the code seems to assume that g++ 4.8 will have `<atomic>`. Might it be better to structure this `#ifdef HAS_STDATOMIC_H` first and then `#ifdef __cplusplus` including `<atomic>` and keep the compatibility stuff surrounded by `extern "C"`?

kit-ty-kate:
> I'm not sure how much we worry now, but the code seems to assume that g++ 4.8 will have `<atomic>`. Might it be better to structure this `#ifdef HAS_STDATOMIC_H` first and then `#ifdef __cplusplus` including `<atomic>` and keep the compatibility stuff surrounded by `extern "C"`?

Since the C++ part is not used internally and we can't test whether or not `<atomic>` is present at use time (to use `__has_include` we'd need C++17) I'm not sure this is desirable.

kayceesrk:
I'm merging this in so that we plod forward :-)

Issue #453, opened 2020-12-08
"malloc(): unsorted double linked list corrupted" errors


Khady:
I've just installed a 4.10 multicore switch successfully. But after that, I struggled a lot to compile any ocaml code in it. Almost every time I try to build something, I get a `malloc(): unsorted double linked list corrupted` errors.

All of this happens on a server with 40 cores, 256gb of ram, debian 10.3.

Switch installation command:
```
opam switch create 4.10.0+multicore --packages=ocaml-variants.4.10.0+multicore --repositories=multicore=git+https://github.com/ocaml-multicore/multicore-opam.git,default
```

My goal was to compile https://github.com/shubhamkumar13/effects-examples/tree/migrate-to-dune. So once I got my switch I tried to install lwt. And got those errors. I had to run the `opam install lwt --yes` until it would actually pass. Basically every package that was being installed in the switch failed to build at least once.

```
#=== ERROR while compiling lwt.5.3.0 ==========================================#
# context     2.0.6 | linux/x86_64 | ocaml-variants.4.10.0+multicore | git+https://github.com/ocaml-multicore/multicore-opam.git
# path        ~/.opam/4.10.0+multicore/.opam-switch/build/lwt.5.3.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p lwt -j 39
# exit-code   1
# env-file    ~/.opam/log/lwt-7224-6d07ae.env
# output-file ~/.opam/log/lwt-7224-6d07ae.out
### output ###
# malloc(): unsorted double linked list corrupted
# [...]
# (cd _build/default/src/unix && /usr/bin/gcc -O2 -fno-strict-aliasing -fwrapv -fPIC -D_FILE_OFFSET_BITS=64 -D_REENTRANT -I. -g -I /home/louis/.opam/4.10.0+multicore/lib/ocaml -I /home/louis/.opam/4.10.0+multicore/lib/bytes -I /home/louis/.opam/4.10.0+multicore/lib/domainslib -I /home/louis/.opam/4.10.0+multicore/lib/mmap -I /home/louis/.opam/4.10.0+multicore/lib/ocaml/threads -I /home/louis/.o[...]
# malloc(): unsorted double linked list corrupted
#          gcc src/unix/unix_readable.o (got signal ABRT)
[...]

PR #455, opened 2020-12-10
No blocking section on fork

abbysmal:
This PR addressed #453, and brings back dune in working condition in Multicore.

The gist of the issue is that `unix_fork` enters a blocking section before running `fork`, leading to a rogue thread taking over the thread `masterlock` and pushing the child in an invalid state.
#439 helped uncover this issue by actually meddling with the thread chaining (by freeing now defunct thread entries in the thread chaining.), and trying to allocate some data. (and `dune` was a victim because after forking a multithreaded program, it does run `execve`, which allocates data directly on a now possibly corrupted heap)

This PR also brings in a testcase that can reproduce reliably the issue on the current `parallel_minor_gc` branch, replicating the problematic profile for this issue.

I also attempted building and using `dune` with success.


ctk21:
LGTM, merging.

Issue #457, opened 2020-12-14
no-effect-syntax still contains the effect syntax on the Typedtree


EduardoRFS:
By trying to use merlin with the no-effect-syntax I noticed that the current implementation still has the effect system on the typed AST and was pointed out by @kayceesrk that it should be removed

Ideally from a user-space perspective the `.cmi` file would have the same runtime representation as the one upstream, as it uses the same cmi magic number

Sudha247:
Fixed by #458.

PR #458, opened 2020-12-15
no-effect-syntax: remove effects from typedtree

Sudha247:
This patch removes effect syntax from typedtree, addressing #457.  The typed AST does not contain effects anymore, enabling external applications using the AST to work with domains-only Multicore (`no-effect-syntax` branch).


EduardoRFS:
Tried with ocaml-lsp and it doesn't crashes anymore, which is great, still not working for stdlib value, probably because there is still some difference between the cmi on 4.10 and cmi on the no-effect-syntax

EduardoRFS:
There is two additional commits on this branch, it keeps merlin working, the important thing is the typedtree have a compatible format also the ids for the predef types are identical 

https://github.com/EduardoRFS/ocaml/commits/4.10.0%2Btries

Sudha247:
Thanks @EduardoRFS! Added the patches to this PR.

kayceesrk:
There is a CI failure, but this is unrelated to the PR. I've restarted the job, but merging this anyway. Thanks @Sudha247 and @EduardoRFS!

Issue #424, opened 2020-10-23
Issues with global roots management


kayceesrk:
There are several issues currently with global roots management. I thought it might be best to post a single issue that gathers everything together and then make separate issues for each of them. 

## Global roots

OCaml allows the creation of global roots, which are considered to be roots for the GC in addition to the current stack, the live registers, and the C local roots. The global roots in Multicore OCaml come in 4 variants. These variants share some of the implementations, but it is good to classify them separately for ease of understanding. 

### 1: Roots used by the compiler itself

`caml_create_root`, `caml_delete_root`, `caml_read_root`, `caml_modify_root`.

These roots are initialised with `caml_create_root` which returns a `caml_root` value. The roots can be read and deleted. The root must be modified with `caml_modify_root`. This is Multicore OCaml only, and is meant to operate nicely with the concurrent GC. 

### 2: Global roots referred to by C global variables

`caml_register_global_root`, `caml_remove_global_root`

This API allows registering C global variables of type `value` to be considered as a root. In particular, the global variable can be modified directly and the GC ensures that the new value remains reachable. 

### 3: Global roots referred to by C global variables but not modified directly 

`caml_register_generational_global_root`, `caml_remove_generational_global_root`, `caml_modify_generational_global_root`

The global variable must not be modified directly but must go through `caml_modify_generational_global_root`. This is done for performance.

### 4: Native dynlinked globals

`caml_register_dyn_global`

### 5: caml_globals created by the native compiler

These globals are created by the `asmcomp` compiler and do not have an API. There is a global table of such globals named `caml_globals` that is accessed by the runtime. 

## Implementation

(1) is implemented by having a linked list `roots_all` that is a valid OCaml object, whose spine is in the major heap, and each value is the registered root either in the major or the minor heap. The clever trick is to that `caml_modify_root` uses write barrier to update the root and hence, the usual inter-generational pointer mechanism kicks in for making sure that the root is kept alive. The linked list of roots is only scanned during the major GC. 

Having `roots_all` appear as a regular OCaml linked list has advantages for the concurrent mark-sweep collection. At the beginning of the major cycle, all of the domains try to mark the head of `roots_all` linked list. In the common case, one of the domains will mark it and push the head into its mark stack. The rest of the domains will see the head of `roots_all` marked and will not push it into their mark stacks. Either way, the roots are safely marked by concurrent domains without much wasted work. 

[...]

PR #459, opened 2020-12-17
Replace caml_root API with global roots

Sudha247:
This is one of the patches towards #424. In this one, variables of type `caml_root` are converted to `value` type and they are registered as global root or generational global root depending on the variable. 

There's one more item of type `caml_root` in `domain_startup_params`, `callback` which also needs to be converted to `value`, after which `caml_root` API can be removed entirely. I'm currently working on this and shall update the patch when it's done.

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -64,7 +64,7 @@ struct caml_thread_struct {
   struct longjmp_buffer *exit_buf;
   int backtrace_pos;
   code_t * backtrace_buffer;
-  caml_root backtrace_last_exn;
+  value backtrace_last_exn;
   value * gc_regs;
   value * gc_regs_buckets;
   value ** gc_regs_slot;
@@ -217,7 +217,8 @@ static caml_thread_t caml_thread_new_info(void)
   th->exit_buf = NULL;
   th->backtrace_pos = 0;
   th->backtrace_buffer = NULL;
-  th->backtrace_last_exn = caml_create_root(Val_unit);
+  th->backtrace_last_exn = Val_unit;
+  caml_register_global_root(&th->backtrace_last_exn);
   th->domain_id = d->state->id;
 
   #ifndef NATIVE_CODE
@@ -259,7 +260,7 @@ static void caml_thread_remove_info(caml_thread_t th)
   th->next->prev = th->prev;
   th->prev->next = th->next;
   caml_free_stack(th->current_stack);
-  caml_delete_root(th->backtrace_last_exn);
+  caml_remove_global_root(&th->backtrace_last_exn);
   caml_stat_free(th);
   return;
 }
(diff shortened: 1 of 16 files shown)


otherlibs/systhreads/st_stubs.c
```diff
@@ -217,7 +217,8 @@ static caml_thread_t caml_thread_new_info(void)
   th->exit_buf = NULL;
   th->backtrace_pos = 0;
   th->backtrace_buffer = NULL;
-  th->backtrace_last_exn = caml_create_root(Val_unit);
+  th->backtrace_last_exn = Val_unit;
+  caml_register_global_root(&th->backtrace_last_exn);
```
ctk21, line 221: Is this address really safe to take as `th` is stack allocated?

otherlibs/systhreads/st_stubs.c
```diff
@@ -259,7 +260,7 @@ static void caml_thread_remove_info(caml_thread_t th)
   th->next->prev = th->prev;
   th->prev->next = th->next;
   caml_free_stack(th->current_stack);
-  caml_delete_root(th->backtrace_last_exn);
+  caml_remove_global_root(&th->backtrace_last_exn);
```
ctk21, line 263: Are we always sure this address is going to be the one we want, rather than the copy by value for the arg?

runtime/backtrace_byt.c
```diff
@@ -235,9 +235,9 @@ int caml_alloc_backtrace_buffer(void){
 
 void caml_stash_backtrace(value exn, value * sp, int reraise)
 {
-  if (exn != caml_read_root(Caml_state->backtrace_last_exn) || !reraise) {
+  if (exn != Caml_state->backtrace_last_exn || !reraise) {
     Caml_state->backtrace_pos = 0;
-    caml_modify_root(Caml_state->backtrace_last_exn, exn);
+    caml_modify_generational_global_root(&Caml_state->backtrace_last_exn, exn);
```
ctk21, line 240: This looks fishy to me. We register as a `caml_register_global_root` and modify directly in `backtrace_nat` but here we treat it as a generational_global_root.

runtime/debugger.c
```diff
@@ -181,7 +181,8 @@ void caml_debugger_init(void)
   flags = caml_alloc(2, Tag_cons);
   Store_field(flags, 0, Val_int(1)); /* Marshal.Closures */
   Store_field(flags, 1, Val_emptylist);
-  marshal_flags = caml_create_root(flags);
+  marshal_flags = flags;
+  caml_register_generational_global_root(&marshal_flags);
```
ctk21, line 185: We are taking an address of a stack allocated variable here.

ctk21 (commented):
Looks like this one isn't quite ready yet, there are some CI failures.

I noticed some places where we might be doing something unsafe:
- taking addresses of C stack variables and passing them to the GC
- I think there's a place in backtrace where we are mixing generational and non-generational global roots

Sudha247:
Closing in favour of #488.

PR #461, opened 2020-12-24
Remove stw/leader_collision events from eventlog

sadiqj:
For stop-the-world pauses which can be left early it is possible that a domain can actually spin waiting to grab the stw_leader position and this ends up filling the eventlog with `stw/leader_collision` messages. This makes viewing and analysing the logs almost impossible.


ctk21 (commented):
Sounded as if these were spamming the logs in some tests.
This removal looks reasonable, merging!

PR #462, opened 2021-01-08
Move from travis to Github Actions

abbysmal:
This PR aims to reuse the bits from upstream to enable Github actions in our repository.
See: https://github.com/ocaml/ocaml/pull/10043/files

I tried to reuse as many things as possible from the PR to avoid diverging trees, however upstream's CI setup does not include such things as macos testing (I believe it is done in Inria's CI currently?), so I baked it in.

I did not do it very properly, I think the same result could be achieved with a matrix in the action config, but the current setup runs the CI properly on my own branch, we can refine from here. (see https://github.com/Engil/ocaml-multicore/actions/runs/472531411).

I included Linux run (with debug runtime) and macos run (with debug runtime off) for now¸ on x86_64 only.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,37 @@
+name: main
+
+on: [push, pull_request]
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+  # Ensure that make distclean can be run from an empty tree
+    # - name: distclean
+    #   run: |
+    #     MAKE_ARG=-j make distclean
+    - name: configure tree
+      run: |
+        MAKE_ARG=-j XARCH=x64 bash -xe tools/ci/actions/runner.sh configure
+    - name: Build
+      run: |
+        MAKE_ARG=-j bash -xe tools/ci/actions/runner.sh build
+    - name: Run the testsuite
+      run: |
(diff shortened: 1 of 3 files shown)


abbysmal:
Also to note: I commented out various things to leave only the testsuite.
It's seems the sanity check on `make distclean` does not hold in our branch. (it is not possible to run distclean on an empty tree because Makefile.config and friends are absent.)

abbysmal:
I reworked a bit the jobs description, and tested if we could run jobs for both debug and non debug builds.
https://github.com/Engil/ocaml-multicore/actions/runs/477554275
I decided to opt out of running the macos-debug build, as a 51 minutes running time is a bit too much, but I'll leave the linux, non debug build.

ctk21 (commented):
I've taken a look - I think its ok, but I'm not sure I'm 100% qualified to understand how the new world works. 

I have one query:
 - Previously we had the compiler being explicitly set to `clang` for Mac OS and `gcc` for Linux, but I don't see these in the new `main.yml` file. It would be good to test on Mac OS with `clang`, is it just an oversight or is there no way to configure the compiler in the new world?

abbysmal:
I don't think it is required to set the compiler to `clang` here as it is the default on the runner for `macos-latest`, see https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

The `configure` check for clang does indeed detect it:
https://github.com/Engil/ocaml-multicore/runs/1681225717?check_suite_focus=true#step:3:246

I think this is reliable enough to not attempt to explicitly set it up?

PR #463, opened 2021-01-12
Minor GC: restrict global roots scanning to one domain

Sudha247:
Currently, all live domains scan all the global roots during a minor collection. This patch restricts the global root scanning in minor GC to just one domain. The first domain to reach the scanning part carries on the scanning, with a lock around it. Once it is done, a flag is updated, which indicates that other domains need not scan the global roots.
This is of the patches addressing #424.

The sequential and parallel macro benchmark results for this, are below.

**Sequential - Normalised time**
image

**Parallel - Time**
image

Results for the Global roots parallel micro benchmark `globroots_mp.ml` with `10000` roots from https://github.com/ocaml-bench/sandmark/pull/200:

| Number of domains | parallel_minor_gc | minor-gc-globroots |
|-------------------|-------------------|--------------------|
| 1                 | 0.254             | 0.256              |
| 2                 | 0.443             | 0.439              |
| 4                 | 0.556             | 0.54               |
| 8                 | 0.815             | 0.812              |
| 12                | 1.11              | 1.039              |
| 16                | 1.88              | 1.88               |
| 20                | 2.099             | 2.1                |
| 24                | 2.549             | 2.518              |

There is no huge difference between the two branches, this could be due to overheads lying elsewhere.

Examining the eventlogs produced by produce by the same microbenchmark on 12 domains (for a representative value), the event `minor_gc/global_roots` was found 68081 times in `parallel_minor_gc` and only 5983 times with this patch.

### DIFF
--- a/runtime/minor_gc.c
+++ b/runtime/minor_gc.c
@@ -45,8 +45,11 @@ extern value caml_ephe_none; /* See weak.c */
 struct generic_table CAML_TABLE_STRUCT(char);
 
 static atomic_intnat domains_finished_minor_gc;
+static atomic_intnat domain_finished_root;
 
 static atomic_uintnat caml_minor_cycles_started = 0;
+
+static caml_plat_mutex global_roots_lock = CAML_PLAT_MUTEX_INITIALIZER;
 
 double caml_extra_heap_resources_minor = 0;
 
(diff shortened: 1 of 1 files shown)


ctk21 (approved):
LGTM

PR #464, opened 2021-01-18
Replace Field_imm with Field

Sudha247:
This PR aims to remove all the leftover Field immediate from the concurrent minor collector, which are no longer necessary.


ctk21:
I think this is good to go, will merge

Issue #469, opened 2021-02-03
Systhreads created from new domains causing segfault


Sudha247:
The following testcase results in a segfault (increase the number of domains in case it doesn't).

```ocaml
let n = 42
let num_domains = try int_of_string Sys.argv.(1) with _ -> 4

let rec fib n =
  if (n < 2) then n
  else fib (n - 1) + fib (n - 2)

let th_create () =
  let t = Thread.create fib in
  Thread.join (t n)

let _ =
  let domains = Array.init num_domains (fun _ -> Domain.spawn(th_create)) in
  Array.iter Domain.join domains
```
Looking at it on gdb, it fails in [this place](https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/runtime/roots.c#L69), in a minor collection when it is scanning a systhread's local state. This may be related to the CI failure on `multicore_systhreads` like in [this one](https://github.com/ocaml-multicore/ocaml-multicore/runs/1793451367).

abbysmal:
Thanks for the report.

As it turns out, the logic in systhreads in `caml_thread_stop` is invalid.
`Current_thread = Current_thread->next;` is done after `caml_thread_remove_info`, which obviously yields bogus results since `th`  (or `Current_thread`) is freed. 
This triggers segfaults when the backup thread kicks in next because it reuses the `Current_thread` slot. (which is the sole purpose of this assignment in the first place.)
This is also not a common occurrence because it specifically requires backup threads to be running and the proper timing in the running threads.
I think this broke back in december, when adjustments were made to the lifecycle of threads.
I'll also take time to push more assertions on the thread chaining manipulation functions.

Working on a fix.

PR #470, opened 2021-02-03
Systhreads: Current_thread->next value should be saved before setting it to Current_thread after freeing the thread decriptor.

abbysmal:


ctk21:
This fixes #469  (see https://github.com/ocaml-multicore/ocaml-multicore/issues/469#issuecomment-772456332)

Sudha247 (commented):
Good catch! The changes look good to me.

PR #472, opened 2021-02-16
Major GC: scan global roots from one domain

Sudha247:
As a first step towards parallelizing global roots scanning, this patch scans the global roots from only one domain in a major  cycle. Whichever domain reaches that point first carries on with scanning. Some benchmark results for this patch are below.

**Sequential Benchmarks - Normalised time**

normalised

**Parallel Benchmarks - Time**

image

#### Micro Benchmarks

[`globroots_sp`](https://github.com/ocaml-bench/sandmark/blob/master/benchmarks/multicore-gcroots/globroots_sp.ml) - one domain does all the work in the benchmark, while other domains keep waiting on a channel. Domains other than parent domain do not perform any OCaml work, they do only GC work.

`n = 100000`, Time in seconds

| Domains | Old   | New  |
|---------|-------|------|
| 1       | 2.16  | 2.14 |
| 2       | 3.18  | 2.61 |
| 4       | 4.47  | 2.93 |
| 8       | 6.83  | 3.46 |
| 12      | 9.37  | 4.02 |
| 16      | 16.31 | 5.34 |
| 20      | 19.34 | 6.38 |
| 24      | 23    | 8.07 |

[`globroots_mp`](https://github.com/ocaml-bench/sandmark/blob/master/benchmarks/multicore-gcroots/globroots_mp.ml) - work is split equally among all domains. 

`n = 100000`, Time in seconds

| Domains | Old   | New  |
|---------|-------|------|
| 1       | 2.16  | 2.14 |
| 2       | 3.61  | 3.07 |
| 4       | 6.33  | 4.68 |
| 8       | 9.2   | 5.4  |
| 12      | 11.71 | 5.6  |
| 16      | 18.57 | 6.72 |
| 20      | 20.31 | 7.19 |
| 24      | 22.34 | 7.52 |

---

Ultimately, the aim is to parallelize global roots scanning in the major GC. Since we see a sizeable performance improvement in the micro benchmarks while scanning from one domain, it might be useful to add this, and it can act as a sequential baseline to measure the performance improvement of the upcoming parallel version.

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -66,8 +66,11 @@ static atomic_uintnat num_domains_to_final_update_last;
 
 static atomic_uintnat terminated_domains_allocated_words;
 
-static atomic_uintnat domain_finished_global_roots;
-static caml_plat_mutex global_roots_lock = CAML_PLAT_MUTEX_INITIALIZER;
+enum global_roots_status{
+  WORK_UNSTARTED,
+  WORK_STARTED
+};
+static atomic_uintnat domain_global_roots_started;
 
 gc_phase_t caml_gc_phase;
 
@@ -992,6 +995,8 @@ static void cycle_all_domains_callback(struct domain* domain, void* unused,
       atomic_store_rel(&num_domains_to_ephe_sweep, num_domains_in_stw);
       atomic_store_rel(&num_domains_to_final_update_first, num_domains_in_stw);
       atomic_store_rel(&num_domains_to_final_update_last, num_domains_in_stw);
+
+      atomic_store(&domain_global_roots_started, WORK_UNSTARTED);
     }
     // should interrupts be processed here or not?
     // depends on whether marking above may need interrupts
@@ -1021,13 +1026,11 @@ static void cycle_all_domains_callback(struct domain* domain, void* unused,
 
   caml_ev_begin("major_gc/roots");
   caml_do_roots (&caml_darken, NULL, domain, 0);
-  if (atomic_load_acq(&domain_finished_global_roots) == 0){
-    if(caml_plat_try_lock(&global_roots_lock)){
+  uintnat work_unstarted = WORK_UNSTARTED;
+  if(atomic_compare_exchange_strong(&domain_global_roots_started, &work_unstarted, WORK_STARTED)){
       caml_scan_global_roots(&caml_darken, NULL);
-      atomic_store(&domain_finished_global_roots, 1);
-      caml_plat_unlock(&global_roots_lock);
-    }
   }
+  
   caml_ev_end("major_gc/roots");
 
   if (domain->state->mark_stack->count == 0) {
@@ -1292,8 +1295,7 @@ mark_again:
       cycle simultaneously, we loop until the current cycle has ended,
       ignoring whether caml_try_run_on_all_domains succeeds. */
 
-    atomic_store(&domain_finished_global_roots, 0);
-    
+
     while (saved_major_cycle == caml_major_cycles_completed) {
       if (barrier_participants) {
         cycle_all_domains_callback(d, (void*)0, participant_count, barrier_participants);
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -66,6 +66,9 @@ static atomic_uintnat num_domains_to_final_update_last;
 
 static atomic_uintnat terminated_domains_allocated_words;
 
+static atomic_uintnat domain_finished_global_roots;
+static caml_plat_mutex global_roots_lock = CAML_PLAT_MUTEX_INITIALIZER;
+
 gc_phase_t caml_gc_phase;
 
 uintnat caml_get_num_domains_to_mark () {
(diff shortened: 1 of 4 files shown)


runtime/major_gc.c
```diff
@@ -66,6 +66,9 @@ static atomic_uintnat num_domains_to_final_update_last;
 
 static atomic_uintnat terminated_domains_allocated_words;
 
+static atomic_uintnat domain_finished_global_roots;
+static caml_plat_mutex global_roots_lock = CAML_PLAT_MUTEX_INITIALIZER;
```
ctk21: I don't think we need this lock and the atomic variable to get the exclusion we need (more is always more complex and potential for bugs in future). Please can we use just the atomic variable?

runtime/major_gc.c
```diff
@@ -1017,6 +1021,13 @@ static void cycle_all_domains_callback(struct domain* domain, void* unused,
 
   caml_ev_begin("major_gc/roots");
   caml_do_roots (&caml_darken, NULL, domain, 0);
+  if (atomic_load_acq(&domain_finished_global_roots) == 0){
```
ctk21: I'm not a fan of having both the lock and the atomic variable acting as synchronization. Can we just use atomic `domain_finished_global_roots` as the only synchronization primitive?

ctk21 (commented):
I'm not sure this code has the same guarantees on the global roots and the mutators running as in the previous code. 

In the previous code all domains will have been sure to have visited (and so called `caml_darken` on) all global roots. In this new code, one domain will be marking the global roots, but it is now possible for another domain to:
- run through the global roots and see someone is handling them
- progress forwards and re-enter ocaml while the other domain is still handling the global roots
- we now have mutators running which could alter the global roots while there are global roots which have not been visited by `caml_darken`

Is that safe? (If it is, are we able to document the argument for why it is safe?)

runtime/major_gc.c
```diff
@@ -1280,6 +1291,9 @@ static intnat major_collection_slice(intnat howmuch,
     /* To handle the case where multiple domains try to finish the major
       cycle simultaneously, we loop until the current cycle has ended,
       ignoring whether caml_try_run_on_all_domains succeeds. */
+
+    atomic_store(&domain_finished_global_roots, 0);
```
ctk21: Not really sure why we are doing this here. Might it be better to do it inside the `cycle_all_domains_callback`, that way all the code for synchronizing is in the same place?

runtime/major_gc.c
Sudha247: Agree that having both lock and atomic variable adds more complexity. The issue I think we might have with using just the atomic variable is this:
* Domain 1 goes into `caml_scan_global_roots` and is not done yet.
* Domain 2 reaches the same point. Seeing that the atomic variable is still 0, it starts scanning the global roots, meaning we could end up with multiple domains scanning global roots.

runtime/major_gc.c
Sudha247: If we do it inside `cycle_all_domains_callback`, is there a possibility that the variable set to 1 by one domain could be reset to 0 by some other domain? I thought there was, hence I did it outside `cycle_all_domains_callback`. If that's not an issue, then it is better to put this inside the function.

ctk21: so within `cycle_all_domains_callback` you have the `caml_global_barrier_is_final` block which gives you one domain to do any setup of the variables you need, knowing that all the other domains are waiting for you.

runtime/major_gc.c
ctk21: You might need to alter the meaning of the atomic. One way to do it might be to have `domain_global_roots_started` as an enum:
- `0` for `work_unstarted`
- `1` for `work_started`

Then a domain can CAS to claim the work when it is `work_unstarted`, with everybody else skipping it if the work has been started.  

The variable can be initialized in the `caml_global_barrier_is_final` block by the last entrant to the section.

ctk21: As it happens, the current code isn't enforcing only one domain does the work. Consider this sequence:
 - domain A sees `domain_finished_global_roots` is `0` but then suspends
 - domain B sees `domain_finished_global_roots` is `0` takes the lock, does all the work and then releases the lock
 - domain A goes to take the lock, gets it and then does all the work again

Sudha247: Yes, got it. This is a great suggestion, thanks for the help! I've added a patch with a similar approach.

runtime/major_gc.c
Sudha247: Done.

Sudha247:
> Is that safe? (If it is, are we able to document the argument for why it is safe?)

If a mutator resumes in one domain while another one is still darkening the global roots, it is potentially not safe. It is possible that the mutator alters the set of global roots. If say the domain that's marking global roots has already finished marking and a new root added is still `UNMARKED`, this could lead to some unsafe GC operation.

Despite that, I think it is not possible for a mutator to resume its operation while a domain is still scanning global roots. Perhaps it'd be useful to add assertions to make sure that this really is the case.

ctk21:
Adding the assert satisfies me; we should only ever enter `cycle_all_domains_callback` if the stop-the-world `leave_when_done` parameter is set to `0`.

This PR looks good to go for me.

PR #476, opened 2021-02-22
global roots parallel tests

Sudha247:
Adding some of the parallel global roots programs from sandmark to the testsuite, aiming to keep in check global roots interaction with domain lifecycle.

There are two new tests in this patch:

* `globroots_parallel_single.ml` - In this test, one domain does all the global roots work like registering roots, removing roots and calling Gc functions. Other domains just keep spinning till the first one's done. The idle domains do undertake GC work.

* `globroots_parallel_multiple.ml` - Global roots work is shared equally in terms of count amongst all domains.

The parallel tests don't print anything, because the printing order differs between different instances of their execution. The sequential test that is `globroots.ml` is unchanged; I've only removed some redundant definitions that are now present in `globroot.ml`.


ctk21:
Adding coverage where we have global root operations and parallel domains is a good idea.

So I have a couple of questions:
 - does `globroots_parallel_single.ml` add much over `globroots_parallel_multiple.ml` given that the spawned domains in `globroots_parallel_single.ml` don't cause any allocations or much domain lifecycle pressure?  
 - is there a way to get more 'domain lifecycle pressure' into these tests? What I'm thinking is something that is spawning a bunch of domains where each does a bit of global roots and then comes down in a loop; maybe something along these lines:
 ```ocaml
for i = 1 to 25 do
  let burn = (fun () -> TestClassic.test 10; TestGenerational.test 10) in
  let a = Array.init 4 (fun _ -> Domain.spawn burn) in
  Array.iter Domain.join a
done
 ```

Sudha247:
>does globroots_parallel_single.ml add much over globroots_parallel_multiple.ml given that the spawned domains in globroots_parallel_single.ml don't cause any allocations or much domain lifecycle pressure?

Right, `globroots_parallel_single.ml` doesn't add much over the test `globroots_parallel_multiple.ml` allocating from multiple domains. I have removed it now. It may be useful for performance considerations, for which it is in Sandmark anyway.

> is there a way to get more 'domain lifecycle pressure' into these tests?

Yes, this is a good idea! Something similar to the suggested code snippet is in `globroots_parallel_spawn_burn.ml`. Its running time wasn't too long when I tested.

ctk21:
Great. 

I guess the only thing that we aren't covering with these is if we have global root operations across domains; such as:
- one domain creates the root but others use it
- one domain creates the root, another deletes it

Not sure if that is easy to factor into the code that is there because you would have to be careful with the register and remove operations. (Which also raises the question if we produce the correct exceptions/errors when multiple domains do something bad by accident - such as trying to remove a root twice at the same time, or register the same root twice).

testsuite/tests/gc-roots/globroot.ml
```diff
@@ -0,0 +1,77 @@
...
+
+module Test(G: GLOBREF) = struct
+
+  let size = 1024
+
+  let vals = Array.init size Int.to_string
+
+  let a = Array.init size (fun i -> G.register (Int.to_string i))
+
+  let check () =
```
ctk21, line 37: This function doesn't actually seem to get called. Not sure that's intended, it feels like it should be called after each `change ()` call in `test`

Sudha247:
I have added some more cases in `globroots_parallel_spawn_burn` to register roots from one domain, modify and remove them from other domains.

testsuite/tests/gc-roots/globroot.ml
Sudha247: That's something I have been wondering myself. The original test in trunk (https://github.com/ocaml/ocaml/blob/trunk/testsuite/tests/gc-roots/globroots.ml) too doesn't call `check` function. But I think you're right, `check` is now called after `change`.

ctk21:
Adds some coverage in here which will be good if we change how the roots are managed between domains, merging!

PR #477, opened 2021-02-25
Move tls areas to a dedicated memory space

abbysmal:
This PR changes the way we allocate individual domain's tls areas.

Currently, tls areas are allocated alongside the minor heaps, and the minor heaps segment is effectively divided as such:

`Domain 0 tls area -> guard page -> d0 minor heap -> Domain 1 tls area ... ect`

This is not an optimal layout for Domain Local Allocation buffers, we need to be able to bump the allocation pointer without any extra book keeping, besides knowing the end of the segment.

This PR move the tls areas to their own, malloc'ed space, and the minor_heaps segment is now only contiguously storing the minor heaps and nothing else.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -115,6 +115,7 @@ CAMLexport atomic_uintnat caml_num_domains_running;
 
 CAMLexport uintnat caml_minor_heaps_base;
 CAMLexport uintnat caml_minor_heaps_end;
+CAMLexport uintnat caml_tls_areas_base;
 static __thread dom_internal* domain_self;
 
 static int64_t startup_timestamp;
@@ -217,13 +218,6 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       d = &all_domains[i];
       if (!d->interrupt_word_address) {
         /* never been started before, so set up minor heap */
-        if (!caml_mem_commit((void*)d->tls_area, (d->tls_area_end - d->tls_area))) {
-          /* give up now: if we couldn't get memory for this domain, we're
-             unlikely to have better luck with any other */
-          d = 0;
-          caml_plat_unlock(&s->lock);
-          break;
-        }
         caml_domain_state* domain_state =
           (caml_domain_state*)(d->tls_area);
         atomic_uintnat* young_limit = (atomic_uintnat*)&domain_state->young_limit;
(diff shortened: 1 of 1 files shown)


runtime/domain.c
```diff
@@ -333,24 +327,30 @@ CAMLexport void caml_reset_domain_lock(void)
...
+  void* tls_base;
 
   /* sanity check configuration */
   if (caml_mem_round_up_pages(Minor_heap_max) != Minor_heap_max)
     caml_fatal_error("Minor_heap_max misconfigured for this platform");
 
-  /* reserve memory space for minor heaps */
+  /* reserve memory space for minor heaps and tls_areas */
   size = (uintnat)Minor_heap_max * Max_domains;
+  tls_size = sizeof(caml_domain_state) * Max_domains;
```
ctk21: Can we round the `sizeof(caml_domain_state)` up to a whole page? 
Avoids any possibility of false sharing of cache lines between domains and keeps the pages between domains separate.

runtime/domain.c
```diff
@@ -333,24 +327,30 @@ CAMLexport void caml_reset_domain_lock(void)
...
 
   heaps_base = caml_mem_map(size*2, size*2, 1 /* reserve_only */);
-  if (!heaps_base) caml_raise_out_of_memory();
+  tls_base = malloc(tls_size);
```
ctk21: Might be better to use `caml_stat_alloc_noexc` than `malloc` for this.

ctk21 (commented):
One thing I couldn't see in this patch is where `caml_tls_areas_base` gets deallocated. Shouldn't we deallocate that on shutdown?

The other comment is more general, we're allocating the whole of `caml_tls_areas_base` in one `malloc` call from the startup pthread. This might have a bad NUMA outcome compared to the existing code where we are commiting (with `caml_mem_commit`) in `create_domain` from the pthread with which a domain is spawned. I admit the existing behaviour is weak in the sense that the pthread may subsequently be moved by the OS. However the behaviour we are introducing will very likely place all the TLS areas into one NUMA zone on large machines, which will certainly be bad for pthreads spawned in the other NUMA zones.

Is there a reason to strongly prefer using a single `malloc` for the `tls_area` over using the  `caml_mem_commit` for it in `create_domain`?

abbysmal:
> Is there a reason to strongly prefer using a single `malloc` for the `tls_area` over using the `caml_mem_commit` for it in `create_domain`?

There is no specific reason, we would need to round up each tls_area up to a page (in order to be able to commit them properly). This sounds reasonable to me.

ctk21:
LGTM

PR #480, opened 2021-02-26
Remove leave_when_done and friends from STW API

Sudha247:
After a discussion with @ctk21 on #472, we figured `stw_request.leave_when_done` is best gone. This patch attempts that, by removing barriers from `caml_try_run_on_all_domains*` and `stw_request`.

So, the change this patch brings is, `caml_try_run_on_all_domains*` no longer has a barrier inside, therefore by default the callback function would leave as soon as it's done. When a barrier is present inside `caml_try_run_on_all_domains*` with `leave_when_done` as the parameter to keep spinning or not, it is not straightforward to find out whether the callback function is going to exit when done or will keep spinning till other domains are done. In particular, the outcome could be different at different entry points for the same function.

Wherever barriers are necessary, they would have to be added in the callback function itself, making it easier to reason about whether a function is going to leave, or is it going to wait until other domains are done. This is already done for existing functions requiring a barrier.

### DIFF
--- a/runtime/caml/domain.h
+++ b/runtime/caml/domain.h
@@ -129,9 +129,8 @@ int caml_domain_is_in_stw();
 void caml_run_on_all_running_domains_during_stw(void (*handler)(struct domain*, void*), void* data);
 int caml_try_run_on_all_domains_with_spin_work(void (*handler)(struct domain*, void*, int, struct domain**), void*, void (*leader_setup)(struct domain*),
   void (*enter_spin_callback)(struct domain*, void*), void*,
-  void (*leave_spin_callback)(struct domain*, void*), void*,
-  int);
-int caml_try_run_on_all_domains(void (*handler)(struct domain*, void*, int, struct domain**), void*, void (*leader_setup)(struct domain*), int);
+  void (*leave_spin_callback)(struct domain*, void*), void*);
+int caml_try_run_on_all_domains(void (*handler)(struct domain*, void*, int, struct domain**), void*, void (*leader_setup)(struct domain*));
 
 void caml_global_barrier();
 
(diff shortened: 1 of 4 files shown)


runtime/caml/domain.h
```diff
@@ -129,9 +129,8 @@ int caml_domain_is_in_stw();
 void caml_run_on_all_running_domains_during_stw(void (*handler)(struct domain*, void*), void* data);
 int caml_try_run_on_all_domains_with_spin_work(void (*handler)(struct domain*, void*, int, struct domain**), void*, void (*leader_setup)(struct domain*),
   void (*enter_spin_callback)(struct domain*, void*), void*,
-  void (*leave_spin_callback)(struct domain*, void*), void*,
-  int);
-int caml_try_run_on_all_domains(void (*handler)(struct domain*, void*, int, struct domain**), void*, void (*leader_setup)(struct domain*), int);
+  void (*leave_spin_callback)(struct domain*, void*), void*);
```
ctk21: Given that there is no longer a leave spin in the API, we should also remove `leave_spin_callback` as that is unused.

runtime/major_gc.c
```diff
@@ -1053,6 +1052,7 @@ static void cycle_all_domains_callback(struct domain* domain, void* unused,
   domain->state->final_info->updated_first = 0;
   domain->state->final_info->updated_last = 0;
 
+  caml_global_barrier();
```
ctk21, line 1058: Might be nice to signpost that we need this for correctness in a comment, at the very least for the global root marking.

ctk21 (commented):
I always enjoy a PR that removes un-necessary and potentially confusing bits of an API. 

It looks good, although we should remove the leave spin work because that is unused after removing `leave_when_done`.

abbysmal:
I also took a look and it looks good to me, thank you for this, this is a neat cleanup.

ctk21:
LGTM, merging

PR #481, opened 2021-03-02
Don't share array amongst domains in gc-roots tests

Sudha247:
The parallel global roots tests share one array amongst all domains. But for the `change` and `check` function to work as intended, every domain should have its own array. This patch attempts to fix it. It was an oversight on my part.

Now, every domain has its own instance of modules `TestClassic` and `TestGenerational`, so they have their own arrays.


ctk21 (commented):
Great, this fixes the shared `a` array.

However I think there's still a bit more unintended shared state with `Random`. For code running in multiple domains, we should be using domain-local storage when calling `Random.int`. Might matter more for benchmarking.

Also, is it possible to clean up the duplication between `globroot.ml` and `globroots.ml`? I've slipped up a couple of times mistaking one for the other.

Sudha247:
>However I think there's still a bit more unintended shared state with Random. For code running in multiple domains, we should be using domain-local storage when calling Random.int. Might matter more for benchmarking.

Yes, done.

>Also, is it possible to clean up the duplication between globroot.ml and globroots.ml? I've slipped up a couple of times mistaking one for the other.

The naming was indeed convoluted, I've tried to fix that. The sequential test previously `globroots.ml` is renamed to `globroots_sequential.ml`, and the parent module with common structures is named `globroots.ml`.

As far the duplication, there's a separate functor defined in `globroots_sequential` because the test prints a `.` for every iteration. Parallel tests don't print anything inside the loop since the order of printing is non-deterministic. In order to preserve the printing part present in the original upstream test, the Test functor is defined again in the sequential test. So, now we have:
* `globroots.ml` -> Module with common structs
* `globroots_sequental.ml` -> Sequential test
* `globroots_parallel.ml` -> Parallelized version of `globroots_sequental.ml`
* `globroots_parallel_spawn_burn.ml` -> Captures more interactions with domain life-cycle and mutating same roots from multiple domains.

There could be a better way to organize this, I'll be happy to update it if that's the case.

ctk21:
Thanks for fixing it up.

I think this is good to go and should help make the CI more reliable. Merging!

PR #482, opened 2021-03-05
check for -lexecinfo in order to build on musl/alpine

### LINKED ISSUE #473: Building on musl requires dynamically linked execinfo
multicore fails to build without libexecinfo. The dependency seems to be the `backtrace_symbols` function in [`misc.c`](../blob/e6017/runtime/misc.c#L59), usually expected to exist in glibc itself without external linkage. I'm not sure if `#include <execinfo.h>` was introduced in multicore or if it comes from upstream, but it dates back to 2016. Either way, this issue is unique to multicore now.

Attempting to pass `-lexecinfo` to `configure` via the `CFLAGS`, `LIBS`, `DLLIBS` seems to have no effect. Reading through the configure script I saw an undocumented variable which may be used to inject the flag: `cclibs`.

```sh
export cclibs=-lexecinfo
./configure ...
make
```

Doing it like this seems to work.
Tested on Alpine 3.12.1

PS: what's the best way to streamline this into the build process? I'm hoping to write an OPAM file for a musl compiler variant. And `cclibs` feels like a hackish approach.

avsm:
fixes #473

### DIFF
--- a/runtime/misc.c
+++ b/runtime/misc.c
@@ -28,8 +28,10 @@ __declspec(noreturn) void __cdecl abort(void);
 #include <stdio.h>
 #include <string.h>
 #include <stdarg.h>
-#include <execinfo.h>
 #include <stdlib.h>
+#if defined(DEBUG) || defined(NATIVE_CODE)
+#include <execinfo.h>
+#endif
 #include "caml/config.h"
 #include "caml/misc.h"
 #include "caml/memory.h"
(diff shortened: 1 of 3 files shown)


runtime/misc.c
```diff
@@ -28,8 +28,10 @@ __declspec(noreturn) void __cdecl abort(void);
 #include <stdio.h>
 #include <string.h>
 #include <stdarg.h>
+#if defined(DEBUG) || defined(NATIVE_CODE)
```
avsm, line 32: @ctk21 asked about this: I just guarded the includes introduced in https://github.com/ocaml-multicore/ocaml-multicore/commit/a8f1f6cd669#diff-38bae17d49966000ccbcbf7e7660416400d6503c999207067902ea9f55cdd0c5

ctk21:
LGTM, merging

PR #484, opened 2021-03-05
Thread local allocation buffers

abbysmal:
(See #398 and the [draft on the wiki](https://github.com/ocaml-multicore/ocaml-multicore/wiki/Domain-Local-Allocation-Buffers))
This PR introduce thread local allocation buffers. (or Domain local allocation buffers.)

With this PR, we change the way we currently allocate minor heaps segment for each domains.

In the current setting, a minor heap segment is given to each domains as we initialize the `Max_domains` slots:

`caml_minor_heaps_start` and `caml_minor_heaps_end` delimits the minor heaps space, and contains exactly `Max_domains` segments, each one being set to each domains slot.

When a minor heap is full (any domain's), we proceed to do a minor collection, and reallocate each individual minor heap segments.

This means that each domains currently possess their own address range in the minor heaps area, and they keep the same range during their lifetime.

This PR introduces a new mechanism to allocate minor heaps segments, by using an allocation pointer in the global minor heap segment:
- Domains won't have a dedicated address range on the global minor heap, but will instead bump this allocation pointer to reserve a new minor heap segment from the global heap.
- When an individual minor heap runs out, a domain is allowed to fetch a new minor heap segment from the global_minor_heap instead.
- The new condition for organically triggering a minor collection (outside of a `finish_major_gc` cycle), is for the `global_minor_heap_ptr` to grow past the end of the global minor heaps segment.

3526033 addresses one existing issue where the current sizing for the minor heaps space is not properly sized.

The current state of this PR:
- The bump pointer allocation logic is implemented
- Greedily reallocating minor heap segments (ie. one domain can request another segment when it runs out of minor heap space) is done

Missing parts:
- Terminology adjustment: for the sake of keeping the core logic easy to review, renaming has not yet taken place. (we may want to rename `caml_reallocate_minor_heaps`, `caml_minor_heaps_base`, `caml_minor_heaps_end`... ect)
- Testsuite should be extended to include:
   - Spawning up to `Max_domains` and see if minor heaps math does checks out under various circumstances
   - More organic versions of the `spawn_burn` tests, that does not manually trigger GC, but rather let it happen.
- Benchmarking, and tweaking sizes: this PR takes not time to revisit the currents sizes: a minor heap segment is equal to the current default size. Which means a single busy domain can right now make usage of a minor heap space up to `Max_domains * Max_minor_heap_size` (so 128 * 256k words).
   - Related point: dynamically resizing the minor heap segment size?

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -180,7 +180,8 @@ int caml_reallocate_minor_heap(asize_t wsize)
 
     global_minor_heap_ptr = atomic_load_explicit(&caml_global_minor_heap_ptr, memory_order_acquire);
 
-    CAMLassert(global_minor_heap_ptr != 0x0);
+    CAMLassert(caml_minor_heaps_base <= global_minor_heap_ptr);
+    CAMLassert(global_minor_heap_ptr <= caml_minor_heaps_end);
 
     new_alloc_ptr = global_minor_heap_ptr + Bsize_wsize(wsize);
 
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -179,8 +179,6 @@ int caml_reallocate_minor_heap(asize_t wsize)
 
   caml_ev_begin("reallocate");
 
-  Assert(domain_state->young_ptr == domain_state->young_end);
-
   while (1) {
 
     global_minor_heap_ptr = atomic_load_explicit(&caml_global_minor_heap_ptr, memory_order_acquire);
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -43,9 +43,6 @@
 #include "caml/finalise.h"
 #include "caml/gc_ctrl.h"
 
-#define REALLOCATE_OOM -1
-#define REALLOCATE_HEAP_FULL -2
-
 #define BT_IN_BLOCKING_SECTION 0
 #define BT_ENTERING_OCAML 1
 #define BT_TERMINATE 2
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -115,6 +115,8 @@ CAMLexport atomic_uintnat caml_num_domains_running;
 
 CAMLexport uintnat caml_minor_heaps_base;
 CAMLexport uintnat caml_minor_heaps_end;
+CAMLexport atomic_uintnat caml_global_minor_heap_ptr;
+
 CAMLexport uintnat caml_tls_areas_base;
 static __thread dom_internal* domain_self;
 
@@ -169,13 +171,41 @@ asize_t caml_norm_minor_heap_size (intnat wsize)
 int caml_reallocate_minor_heap(asize_t wsize)
 {
   caml_domain_state* domain_state = Caml_state;
+  uintnat global_minor_heap_ptr;
+  uintnat new_alloc_ptr;
+
+  caml_ev_begin("reallocate");
+
   Assert(domain_state->young_ptr == domain_state->young_end);
+
+  while (1) {
+
+    global_minor_heap_ptr = atomic_load_explicit(&caml_global_minor_heap_ptr, memory_order_acquire);
+
+    #ifdef DEBUG
+    CAMLassert(global_minor_heap_ptr != 0x0);
+    #endif
+
+    new_alloc_ptr = global_minor_heap_ptr + Bsize_wsize(wsize);
+
+    /* CAS away and bump the allocation pointer: if it fails a domain likely
+       snatched the requested segment. Restart the loop, else break. */
+    if(atomic_compare_exchange_strong(&caml_global_minor_heap_ptr,
+				      &global_minor_heap_ptr,
+				      new_alloc_ptr)) {
+      break;
+    };
+
+  }
 
   /* free old minor heap.
      instead of unmapping the heap, we decommit it, so there's
      no race whereby other code could attempt to reuse the memory. */
   caml_mem_decommit((void*)domain_self->minor_heap_area,
                     domain_self->minor_heap_area_end - domain_self->minor_heap_area);
+
+  /* new minor heap area is the previous (pre-CAS) global_minor_heap_ptr */
+  domain_self->minor_heap_area = global_minor_heap_ptr;
 
   wsize = caml_norm_minor_heap_size(wsize);
 
@@ -197,6 +227,7 @@ int caml_reallocate_minor_heap(asize_t wsize)
   domain_state->young_end = (char*)(domain_self->minor_heap_area + Bsize_wsize(wsize));
   domain_state->young_limit = (uintnat) domain_state->young_start;
   domain_state->young_ptr = domain_state->young_end;
+  caml_ev_end("reallocate");
   return 0;
 }
 
@@ -354,11 +385,11 @@ void caml_init_domains(uintnat minor_heap_wsz) {
 
   caml_minor_heaps_base = (uintnat) heaps_base;
   caml_minor_heaps_end = (uintnat) heaps_base + size;
+  caml_global_minor_heap_ptr = (uintnat) heaps_base;
   caml_tls_areas_base = (uintnat) tls_base;
 
   for (i = 0; i < Max_domains; i++) {
     struct dom_internal* dom = &all_domains[i];
-    uintnat domain_minor_heap_base;
     uintnat domain_tls_base;
 
     caml_plat_mutex_init(&dom->interruptor.lock);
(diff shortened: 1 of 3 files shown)


abbysmal:
For the `pr7798` failure, it is related to one stat breaking (`minor_words`).
https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/runtime/minor_gc.c#L530
The reason why is because the computation for this is not reliable anymore as one domain may have cycled through a few minor heaps segments before this statistic is collected.
One way to handle it could be to store it in the domain state (and update it each time we cycle through a new minor heap).
I do not quite like this as it adds requirements on identifying why a minor heap segment was fetched. (domain spawn vs domain consuming all its minor heap vs end of a minor collection)

runtime/domain.c
```diff
@@ -43,6 +43,9 @@
 #include "caml/finalise.h"
 #include "caml/gc_ctrl.h"
 
+#define REALLOCATE_OOM -1
+#define REALLOCATE_HEAP_FULL -2
```
ctk21: Do we need these? Aren't the definitions up in `domain.h`?

runtime/domain.c
```diff
@@ -169,18 +174,47 @@ asize_t caml_norm_minor_heap_size (intnat wsize)
...
+  uintnat global_minor_heap_ptr;
+  uintnat new_alloc_ptr;
+
+  caml_ev_begin("reallocate");
+
+  while (1) {
+
+    global_minor_heap_ptr = atomic_load_explicit(&caml_global_minor_heap_ptr, memory_order_acquire);
+
+    #ifdef DEBUG
```
ctk21: Shouldn't need the `DEBUG` guard for a `CAMLassert`

runtime/domain.c
```diff
@@ -169,18 +174,47 @@ asize_t caml_norm_minor_heap_size (intnat wsize)
...
+    CAMLassert(global_minor_heap_ptr != 0x0);
```
ctk21: Can we strengthen the asserts to 
`CAMLassert(caml_minor_heaps_start <= global_minor_heap_ptr)`
`CAMLassert(global_minor_heap_ptr <= caml_minor_heaps_end)`

runtime/domain.c
```diff
@@ -169,18 +174,47 @@ asize_t caml_norm_minor_heap_size (intnat wsize)
...
+      break;
+    };
+
+  }
 
   /* free old minor heap.
      instead of unmapping the heap, we decommit it, so there's
      no race whereby other code could attempt to reuse the memory. */
   caml_mem_decommit((void*)domain_self->minor_heap_area,
                     domain_self->minor_heap_area_end - domain_self->minor_heap_area);
```
ctk21, line 205: I'm not sure this is safe. Are we really allowed to decommit an area which we might then go and read when doing the minor collection?

Also, what is going to happen for the decommit on the domain that hits the `REALLOCATE_HEAP_FULL` codepath above?

runtime/domain.c
```diff
@@ -169,18 +174,47 @@ asize_t caml_norm_minor_heap_size (intnat wsize)
 int caml_reallocate_minor_heap(asize_t wsize)
 {
   caml_domain_state* domain_state = Caml_state;
-  Assert(domain_state->young_ptr == domain_state->young_end);
+  uintnat global_minor_heap_ptr;
+  uintnat new_alloc_ptr;
+
+  caml_ev_begin("reallocate");
```
ctk21, line 177: This ev event isn't being closed on every exit from the function.

runtime/domain.c
```diff
@@ -905,6 +940,34 @@ static void caml_poll_gc_work()
   if (((uintnat)Caml_state->young_ptr - Bhsize_wosize(Max_young_wosize) <
        (uintnat)Caml_state->young_start) ||
       Caml_state->requested_minor_gc) {
+
+    /* not a requested_minor_gc, so minor_heap ran out, but can fetch more from
+       the global minor heap segment. */
+    /* TODO(engil): rework logic, double checking requested_minor_gc is odd */
+    if (!Caml_state->requested_minor_gc &&
+	(atomic_load_explicit(&caml_global_minor_heap_ptr, memory_order_acquire)
```
ctk21, line 944: Do we need this guard here? Can't we just let `caml_reallocate_minor_heap` deal with all of the logic and parallelism of `caml_global_minor_heap_ptr`.

runtime/domain.c
```diff
@@ -905,6 +940,34 @@ static void caml_poll_gc_work()
...
+
+      switch (caml_reallocate_minor_heap(caml_params->init_minor_heap_wsz)) {
+
+      /* if reallocation didn't work, we keep executing this branch and proceed
+	 with a minor collection */
+      case REALLOCATE_HEAP_FULL:
+	break;
+
+      case REALLOCATE_OOM:
+	caml_raise_out_of_memory();
```
ctk21: I don't think it is safe to raise exceptions in certain contexts; for example we can't raise from the middle of a STW section and re-enter the mutator. I wonder if we just bomb completely with `caml_fatal_error`, since not being able to allocate the minor heap is pretty dreadful.

runtime/domain.c
```diff
@@ -905,6 +940,34 @@ static void caml_poll_gc_work()
...
+	break;
+
+      /* TODO(engil): same as earlier, rework, because in case we did reallocate
+	 we skip the requested_major_slice branch this time around. */
```
ctk21, line 959: We need to get the control flow to work here. Maybe we do the 'try to get a new heap slice' before the existing code and have it always run through the rest of the function, with it tripping the `young_ptr`/`young_start` test if getting a new slice fails because the heap is exhausted. That will then also check the `request_major_slice` condition and any further conditions or functionalities that we add in the future.

runtime/domain.c
```diff
@@ -271,9 +306,12 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
...
     }
 
-    if(caml_reallocate_minor_heap(initial_minor_heap_wsize) < 0) {
-      goto reallocate_minor_heap_failure;
-    }
+    /* setting young_limit and young_ptr to minor_heaps_base
+       to trigger minor_heaps reallocation on GC poll */
+    domain_state->young_start = (char*)caml_minor_heaps_base;
+    domain_state->young_end = (char*)caml_minor_heaps_base;
+    domain_state->young_limit = (uintnat) caml_minor_heaps_base;
```
ctk21, line 309: We need to have a way to avoid losing interrupts if another domain attempts to signal us via `young_limit`. I think we can do this by checking that `young_limit` hasn't been set to `INTERRUPT_MAGIC` and then atomically updating `young_limit`. If it has been set to `INTERRUPT_MAGIC`, then we leave `young_limit` alone to allow the signal to propagate.

ctk21 (commented):
I think the design (nearly) works. There is a possible race that can lead to us losing interrupts (and so deadlock). We have to be extremely careful whenever we touch `young_limit` because other domains may try to interrupt us using that variable. If we blat the `young_limit` pointer in `caml_reallocate_minor_heap` without first checking there isn't an interrupt race, then we are in trouble. We are going to need to change `young_limit` from the mutator to get new heap segments, I think we just need a CAS that ensures that we only update `young_limit` if it isn't `INTERRUPT_MAGIC`.  

We should also clean up the tests and make sure things work. With the `major_wait_gc_backup` I wonder if it is naturally fixed by getting the flow control right in `caml_poll_gc_work`. An alternative would be to alter the test by changing the assert to look at `minor_collections` (and if needed upping the `make 22` to force a minor collection).  I'm keen we have at least one test with multiple domains that exhausts the whole minor heap bump area naturally with multiple domains going hard at their minor heaps. A lot of our tests in `tests/parallel` are triggering forced collections, which might not exercise the code paths in `caml_reallocate_minor_heap` that trigger a minor collection. 

I have some niggles around safety with commit and exceptions. We also need to make sure we get the flow control right in `caml_poll_gc_work` and ensure that marking is not starved.

runtime/domain.c
abbysmal: Right, I think `caml_fatal_error` is fine here, good call

runtime/domain.c
abbysmal: I think this works out. This borrows the same logic as the current reallocate code, which is that a segment is decommited before reallocation.
Before it worked because there was full ownership of a specific minor heap segment. (one segment per domain, always the same).

In this branch, every segments that one domain gets is in the rightful position to decommit the memory: everything beyond the global allocation pointer is unused memory (because resetting the global allocation always involves a minor collection).
So everything we decommit has supposedly been collected in the previous minor cycle (and values promoted if need be).
I don't think there is a path where we would touch values in `to-be-decommited` (beyond the allocation pointer) areas, but I may be missing a subtlety

As for the `REALLOCATE_HEAP_FULL`, hitting that codepath means there is no segment available, so there is nothing to decommit.

I think the confusing part of this logic is, do we actually need to decommit every segment we fetch?
I kept it the same as our current branch (+ accounting for the fact that there is no ownership of individual minor heaps segments per domains). The current code mentions this:
https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/runtime/domain.c#L174

```
 /* free old minor heap.
     instead of unmapping the heap, we decommit it, so there's
     no race whereby other code could attempt to reuse the memory. */
```

I'm slightly confused by this comment, is there a scenario where other code could attempt to reuse the memory, after performing a minor collection? (which is the pre-requisite for resetting the global allocation pointer)

abbysmal: (or maybe it is a concurrent minor quirk?)

ctk21:
Closing as superseded by #508

Issue #485, opened 2021-03-05
Segfault while running the layers benchmark.


shubhamkumar13:
**Bug Descripiton :**  This segfault occurs when I try to run the [layers benchmark](https://github.com/shubhamkumar13/sandmark/blob/layers-bench/benchmarks/irmin/layers.ml) with larger values like `-n 1000` or the default value which is used when no argument is passed (4096) in this case

**Reproduction Steps :**

Step 1 - Clone the [repo](https://github.com/shubhamkumar13/sandmark/tree/layers-bench) and checkout the `layers-bench` branch. 
```
git clone https://github.com/shubhamkumar13/sandmark.git &&
cd sandmark &&
git checkout layers-bench
```

Step 2 - This is an optional step to add the multicore switch (goto Step 3 if you already have a multicore switch but using a fresh switch is recommended). To add the mulitcore switch you can follow the the [README](https://github.com/ocaml-multicore/multicore-opam/blob/master/README.md) file on the multicore opam.

Step 3 - Before starting with adding all the dependencies one needs to clone `ocaml-migrate-parsetree` (from [here](https://github.com/Sudha247/ocaml-migrate-parsetree.git)) and `ppxlib` (from [here](https://github.com/Sudha247/ppxlib.git)) because the ones on the opam repository do not build with multicore. 

Then checkout the `mulitcore` branch for both repos. 

After checking out that particular branch pin the packages using the following commands - 
- `opam pin ocaml-migrate-parsetree.dev .` when you are inside the `ocaml-migrate-parsetree` directory 
- `opam pin ppxlib.dev .`  when you are in the `ppxlib` directory. 

After pinning the 2 packages we can move forward with installing the depedencies for the benchmark.

Step 4 - Download the opam export file from [here](https://gist.githubusercontent.com/shubhamkumar13/515a221e2f66d729b2dacb0f55095b5a/raw/99d25456ad12c22e2d5eb02a91cc3e922774b6ca/layers.export) and import it to the current multicore switch.
```
wget https://gist.githubusercontent.com/shubhamkumar13/515a221e2f66d729b2dacb0f55095b5a/raw/99d25456ad12c22e2d5eb02a91cc3e922774b6ca/layers.export
opam sw import layers.export
```

Step 5 - Finally, goto the irmin benchmark directory `cd ./benchmarks/irmin/`

Step 6 - Try to build the benchmark to manually - `dune build layers.exe`

If you get an error similar to this 
```
[...]

PR #487, opened 2021-03-08
systhreads: set gc_regs_buckets and friends to NULL at thread startup

abbysmal:
This fixes the issue encountered on #485 

Those fields were ignored during the initialization phase, and it is not correct.

I re-checked the initialization of the thread and I think these were the only missing fields, maybe this was a mishap while aligning our systhreads implementation to trunk's.

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -215,6 +215,9 @@ static caml_thread_t caml_thread_new_info(void)
   th->exn_handler = NULL;
   th->local_roots = NULL;
   th->exit_buf = NULL;
+  th->gc_regs_buckets = NULL;
+  th->gc_regs = NULL;
+  th->gc_regs_slot = NULL;
   th->backtrace_pos = 0;
   th->backtrace_buffer = NULL;
   th->backtrace_last_exn = caml_create_root(Val_unit);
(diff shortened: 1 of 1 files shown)


abbysmal:
This does bring the `layers.exe` test in check:
```
●●●● ../../_build/default/benchmarks/irmin/layers.exe --ncommits 616
Running benchmarks in benchmarks/irmin/layers.ml:
  
  { ncommits = 616;
    ncycles = 10;
    depth = 10;
    root = "test-bench";
    clear = false;
    no_freeze = false;
    show_stats = false;
    merge_throttle = Block_writes;
    freeze_throttle = Block_writes }
  
+1615242632us  application 9240 commits completed in 55.08s.
[0.006s per commit, 168 commits per second]
```

otherlibs/systhreads/st_stubs.c
```diff
@@ -215,6 +215,9 @@ static caml_thread_t caml_thread_new_info(void)
   th->exn_handler = NULL;
   th->local_roots = NULL;
   th->exit_buf = NULL;
+  th->gc_regs_buckets = NULL;
```
ctk21: Are we able to order the initialisation of the fields to match the declarations in `caml_thread_struct`?

I tried to check we are initialising everything, but bit tricky when they are out of order. I think `next` and `prev` are not being initialised by `caml_thread_new_info`, maybe we just `NULL` them so that the return value is well formed wherever `caml_thread_new_info` is called.

ctk21 (commented):
Well done on fixing that one. 

I'm amazed the testsuite isn't picking this up, but I guess we must be missing some coverage. If we can find a low-hanging way to rectify that while the issue is fresh and we can see it trip, it would be great.

ctk21:
Had a discussion with @Engil, we're going to merge as is to fix the bug.

PR #488, opened 2021-03-11
Remove caml_root API

Sudha247:
This is another attempt at removing the now obsolete `caml_root` API. This PR supersedes #459 which had some pending items.

`caml_root` variables are now changed to `value` type and managed as generational global roots. This patch tries to stay close to trunk on the variables that are common to both trunk and multicore runtime. Checking off one of the items in #424.

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -231,6 +231,7 @@ static caml_thread_t caml_thread_new_info(void)
   th->external_raise = NULL;
   #endif
 
+  caml_register_generational_global_root(&th->backtrace_last_exn);
   return th;
 }
 
@@ -279,6 +280,7 @@ static void caml_thread_reinitialize(void)
   while (th != Current_thread) {
     next = th->next;
     caml_free_stack(th->current_stack);
+    caml_remove_generational_global_root(&th->backtrace_last_exn);
     caml_stat_free(th);
     th = next;
   }
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -64,7 +64,7 @@ struct caml_thread_struct {
   struct longjmp_buffer *exit_buf;
   int backtrace_pos;
   code_t * backtrace_buffer;
-  caml_root backtrace_last_exn;
+  value backtrace_last_exn;
   value * gc_regs;
   value * gc_regs_buckets;
   value ** gc_regs_slot;
@@ -219,7 +219,7 @@ static caml_thread_t caml_thread_new_info(void)
   th->exit_buf = NULL;
   th->backtrace_pos = 0;
   th->backtrace_buffer = NULL;
-  th->backtrace_last_exn = caml_create_root(Val_unit);
+  th->backtrace_last_exn = Val_unit;
   th->gc_regs = NULL;
   th->gc_regs_buckets = NULL;
   th->gc_regs_slot = NULL;
@@ -264,7 +264,6 @@ static void caml_thread_remove_info(caml_thread_t th)
   th->next->prev = th->prev;
   th->prev->next = th->next;
   caml_free_stack(th->current_stack);
-  caml_delete_root(th->backtrace_last_exn);
   caml_stat_free(th);
   return;
 }
@@ -280,7 +279,6 @@ static void caml_thread_reinitialize(void)
   while (th != Current_thread) {
     next = th->next;
     caml_free_stack(th->current_stack);
-    caml_delete_root(th->backtrace_last_exn);
     caml_stat_free(th);
     th = next;
   }
--- a/runtime/backtrace.c
+++ b/runtime/backtrace.c
@@ -30,7 +30,7 @@
 
 void caml_init_backtrace(void)
 {
-  return;
+  caml_register_generational_global_root(&Caml_state->backtrace_last_exn);
 }
 
 /* Start or stop the backtrace machinery */
@@ -42,10 +42,9 @@ CAMLprim value caml_record_backtrace(value vflag)
     Caml_state->backtrace_active = flag;
     Caml_state->backtrace_pos = 0;
     if (flag) {
-      Caml_state->backtrace_last_exn = caml_create_root(Val_unit);
+      Caml_state->backtrace_last_exn = Val_unit;
     } else {
-      caml_delete_root(Caml_state->backtrace_last_exn);
-      Caml_state->backtrace_last_exn = NULL;
+      caml_remove_generational_global_root(&Caml_state->backtrace_last_exn);
     }
   }
   return Val_unit;
@@ -173,8 +172,7 @@ CAMLprim value caml_restore_raw_backtrace(value exn, value backtrace)
 
   caml_domain_state* domain_state = Caml_state;
 
-  if (domain_state->backtrace_last_exn != NULL)
-    caml_modify_root (domain_state->backtrace_last_exn, exn);
+  caml_modify_generational_global_root (&domain_state->backtrace_last_exn, exn);
 
   bt_size = Wosize_val(backtrace);
   if(bt_size > BACKTRACE_BUFFER_SIZE){
--- a/runtime/backtrace.c
+++ b/runtime/backtrace.c
@@ -42,7 +42,7 @@ CAMLprim value caml_record_backtrace(value vflag)
     Caml_state->backtrace_active = flag;
     Caml_state->backtrace_pos = 0;
     if (flag) {
-      Caml_state->backtrace_last_exn = Val_unit;
+      caml_modify_generational_global_root(&Caml_state->backtrace_last_exn, Val_unit);
     } else {
       caml_remove_generational_global_root(&Caml_state->backtrace_last_exn);
     }
--- a/runtime/backtrace.c
+++ b/runtime/backtrace.c
@@ -28,10 +28,6 @@
 #include "caml/fail.h"
 #include "caml/debugger.h"
 
-void caml_init_backtrace(void)
-{
-  caml_register_generational_global_root(&Caml_state->backtrace_last_exn);
-}
 
 /* Start or stop the backtrace machinery */
 CAMLprim value caml_record_backtrace(value vflag)
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -290,6 +290,9 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
     }
 
     domain_state->backtrace_buffer = NULL;
+    domain_state->backtrace_last_exn = Val_unit;
+    caml_register_generational_global_root(&domain_state->backtrace_last_exn);
+
 #ifndef NATIVE_CODE
     domain_state->external_raise = NULL;
     domain_state->trap_sp_off = 1;
@@ -1272,6 +1275,7 @@ static void domain_terminate()
   }
   caml_sample_gc_collect(domain_state);
   caml_remove_generational_global_root(&domain_state->dls_root);
+  caml_remove_generational_global_root(&domain_state->backtrace_last_exn);
 
   caml_stat_free(domain_state->final_info);
   // run the domain termination hook
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -275,10 +275,8 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto reallocate_minor_heap_failure;
     }
 
-    domain_state->dls_root = caml_create_root_noexc(Val_unit);
-    if(domain_state->dls_root == NULL) {
-      goto create_root_failure;
-    }
+    domain_state->dls_root = Val_unit;
+    caml_register_generational_global_root(&domain_state->dls_root);
 
     domain_state->stack_cache = caml_alloc_stack_cache();
     if(domain_state->stack_cache == NULL) {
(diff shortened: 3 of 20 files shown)


runtime/domain.c
```diff
@@ -1226,7 +1224,7 @@ static void domain_terminate()
 
   caml_gc_log("Domain terminating");
   caml_ev_pause(EV_PAUSE_YIELD);
-  caml_delete_root(domain_state->dls_root);
+  caml_remove_generational_global_root(&domain_state->dls_root);
```
ctk21: I know this is where we previously had the `dls_root` removal, but I don't think it is actually correct. The problem is that there are points in `while(!finished)` shutdown loop which can result in mutator code being run by the domain (e.g. finalisers or signal handlers). When this mutator code runs it is reasonable that it might access a domain-local variable. 

Are we able to move the removal of the `dls_root` to after the `while(!finished)` block?

runtime/domain.c
```diff
@@ -533,8 +529,7 @@ static void* domain_thread_func(void* v)
     caml_gc_log("Domain starting (unique_id = %"ARCH_INTNAT_PRINTF_FORMAT"u)",
                 domain_self->interruptor.unique_id);
     caml_domain_start_hook();
-    caml_callback(caml_read_root(callback), Val_unit);
-    caml_delete_root(callback);
+    caml_callback(*domain_callback, Val_unit);
     domain_terminate();
```
ctk21, line 538: I think we need to do something to release the resources associated with `domain_callback` otherwise we will leak here. 

My reasoning is that the function `caml_domain_spawn` has allocated `domain_callback` with `caml_stat_alloc_noexc` and registered it with the runtime with `caml_register_generational_global_root`, the resource ownership is passed to the child thread when it signals `Dom_started` back to the parent.  

Don't we need to have this just before the call to `domain_terminate`:
```C
    caml_remove_generational_global_root(domain_callback);
    caml_stat_free(domain_callback);
```

runtime/interp.c
```diff
@@ -273,8 +273,9 @@ value caml_interprete(code_t prog, asize_t prog_size)
     value raise_unhandled_closure =
       caml_alloc_1(Closure_tag,
                    Val_bytecode(raise_unhandled_code));
-    raise_unhandled = caml_create_root(raise_unhandled_closure);
-    caml_global_data = caml_create_root(Val_unit);
+    raise_unhandled = raise_unhandled_closure;
+    caml_register_generational_global_root(&raise_unhandled);
+    caml_global_data = Val_unit;
```
ctk21, line 278: I don't understand what `caml_global_data` really is. We have places in this PR where it is being altered raw (with the `Field` macro) but we also have other places (for example see `caml_realloc_global` in `meta.c`) where it is being modified with `caml_modify_generational_global_root`. 

If it is a special global root that is operated on raw, then we would need to have something in the root scanning code to operate on it. If it is a generational global root, then we need to register it somewhere and treat it like that everywhere.

runtime/signals.c
```diff
@@ -329,14 +330,14 @@ CAMLprim value caml_install_signal_handler(value signal_number, value action)
...
-    caml_read_field(caml_read_root(caml_signal_handlers), sig, &handler);
+    caml_read_field(caml_signal_handlers, sig, &handler);
     res = caml_alloc_1 (0, handler);
     break;
   default:                      /* error in caml_set_signal_action */
     caml_sys_error(NO_ARG);
   }
   if (Is_block(action)) {
-    caml_modify_field(caml_read_root(caml_signal_handlers), sig, Field(action, 0));
+    caml_modify_field(caml_signal_handlers, sig, Field(action, 0));
```
ctk21: Is this `caml_modify_field` correct? Isn't `caml_signal_handlers` registered as a generational global root that needs to be modified with `caml_modify_generational_global_root`?

ctk21 (commented):
It'll be very good to have the old `caml_root` API gone and this is much better than #459.

I think there are some areas that need resolving (see code comments):
 - it looks to me that there is a resource leak with the callback closure passed to spawned domains in `domain.c`. 
 - in this PR we treat `caml_global_data` sometimes as a generational root and sometimes as a static root. I couldn't figure out where we even register `caml_global_data` with the GC.
 - we are modifying `caml_signal_handlers` with `caml_modify_field` but registering it as a generational global root. 
 - it would be good to move the removal of the `dls_root` in the shutdown sequence to when it is not possible for there to be mutator code running on the domain. 
 
I nice-to-have but potentially useful feature might be:
 - can we write DEBUG asserts into `caml_modify_generational_global_root` that asserts that the block we are modifying has previously been registered with `caml_register_generational_global_root`. This might be more trouble than it is worth depending on if a suitable algorithm for checking can be found, because it may make it too slow to run programs with a lot of generational global roots in DEBUG mode.

runtime/domain.c
Sudha247: Ah yes, missed it, thanks for spotting this!

runtime/interp.c
Sudha247: So `caml_global_data` is treated as a root in multicore and not treated as root in trunk. I mixed up both looking at what trunk does. Hopefully this is fixed now, `caml_global_data` is registered and modified as a generational global root.

runtime/signals.c
Sudha247: Yes, fixed it.

Sudha247:
> can we write DEBUG asserts into `caml_modify_generational_global_root` that asserts that the block we are modifying has previously been registered with `caml_register_generational_global_root`.

We can do this by adding a `caml_skiplist_key_exists` function that checks if a root is present, and an assert to ensure the key exists in the skiplist we're deleting from. But, I'm not sure we can cover all bases with this; there could be some cases in generational global roots where the same root may or may not be present in both young roots and old roots, like mentioned here: https://github.com/ocaml-multicore/ocaml-multicore/blob/parallel_minor_gc/runtime/globroots.c#L180-L181.

otherlibs/systhreads/st_stubs.c
```diff
@@ -219,7 +219,7 @@ static caml_thread_t caml_thread_new_info(void)
   th->exit_buf = NULL;
   th->backtrace_pos = 0;
   th->backtrace_buffer = NULL;
-  th->backtrace_last_exn = caml_create_root(Val_unit);
+  th->backtrace_last_exn = Val_unit;
```
ctk21, line 222: I don't see where `backtrace_last_exn` is now registered with the GC; surely we need to register it as a generational root and then use the correct modification and delete functions?

runtime/backtrace.c
```diff
@@ -30,7 +30,7 @@
 
 void caml_init_backtrace(void)
 {
-  return;
+  caml_register_generational_global_root(&Caml_state->backtrace_last_exn);
```
ctk21: I don't see anywhere in the code where `caml_init_backtrace` is ever called (which is unfortunate as it probably shouldn't be there).

runtime/backtrace.c
```diff
@@ -42,10 +42,9 @@ CAMLprim value caml_record_backtrace(value vflag)
     Caml_state->backtrace_active = flag;
     Caml_state->backtrace_pos = 0;
     if (flag) {
-      Caml_state->backtrace_last_exn = caml_create_root(Val_unit);
+      Caml_state->backtrace_last_exn = Val_unit;
```
ctk21: I'm not convinced that `caml_register_generational_global_root` is called on `Caml_state->backtrace_last_exn` because I don't think `caml_init_backtrace` is called anywhere.

ctk21 (commented):
Had a couple of queries on the backtrace and systhread pieces. Otherwise I think everything is now consistent. 

Glad you looked into the assert, if it's not trivial then happy to leave.

runtime/backtrace.c
Sudha247: Yes, it was not called anywhere, so `Caml_state->backtrace_last_exn` was not being registered as a root. Thanks for finding this! I have attempted to fix this now, `caml_init_backtrace` is called during startup.

Sudha247:
Made some more changes based on suggestions from @engil and @ctk21 .

There's a slight digression from trunk here due to registering `th->backtrace_last_exn` as a generational global root, whereas on trunk it is scanned manually in `caml_thread_scan_roots`. Also, `domain_state->backtrace_last_exn` is handled in `create_domain` and `domain_terminate`.

otherlibs/systhreads/st_stubs.c
```diff
@@ -264,7 +265,6 @@ static void caml_thread_remove_info(caml_thread_t th)
   th->next->prev = th->prev;
   th->prev->next = th->next;
   caml_free_stack(th->current_stack);
-  caml_delete_root(th->backtrace_last_exn);
   caml_stat_free(th);
```
ctk21, line 268: Don't we need a `caml_remove_generational_global_root` here? Otherwise the root will be registered with the GC at a location that is going to disappear after the `caml_stat_free(th)`

ctk21 (commented):
I saw one problem in `caml_thread_remove_info` with the removal of `th->backtrace_last_exn`.

Also I think there is another problem in `caml_thread_save_runtime_state` & `caml_thread_restore_runtime_state`. With both `backtrace_last_exn` in the `caml_thread_struct` and `Caml_state` now both being generational global roots, we need to modify them with `caml_modify_generational_global_root` otherwise the internal generational global root book keeping looks like it will get out of sync. 

I'm wondering if we need to have a systhread with backtraces and exceptions test. Right now I'm not sure anything exercises the logic of this fiddly backtrace state.

kit-ty-kate:
This PR makes `pa_ppx.0.07.02` and `pa_ppx.0.07` segfault during its compilation process.
The following command now segfault reliably with this branch (rebased on top of 4.12+domains) only:
```
$ pwd
/home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils
$ gdb --args ocamlrun /home/opam/.opam/test/bin/not-ocamlfind reinstall-if-diff pa_ppx_testutils -destdir /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib META pa_ppx_testutils.cma pa_ppx_testutils.cmxa pa_ppx_testutils.a pa_ppx_testutils.cmx pa_ppx_testutils.cmi
[...]
Starting program: /home/opam/.opam/test/bin/ocamlrun /home/opam/.opam/test/bin/not-ocamlfind reinstall-if-diff pa_ppx_testutils -destdir /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib META pa_ppx_testutils.cma pa_ppx_testutils.cmxa pa_ppx_testutils.a pa_ppx_testutils.cmx pa_ppx_testutils.cmi
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
ocamlfind: [WARNING] No such directory: /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils
reinstall-if-diff: remove and install lists have differing lengths
+ ocamlfind remove -destdir /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib pa_ppx_testutils
[Detaching after vfork from child process 28788]
ocamlfind: [WARNING] No such file: /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/META
+ ocamlfind install pa_ppx_testutils -destdir /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib META pa_ppx_testutils.cma pa_ppx_testutils.cmxa pa_ppx_testutils.a pa_ppx_testutils.cmx pa_ppx_testutils.cmi
[Detaching after vfork from child process 28789]
Installed /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/pa_ppx_testutils.cmi
Installed /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/pa_ppx_testutils.cmx
Installed /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/pa_ppx_testutils.a
Installed /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/pa_ppx_testutils.cmxa
Installed /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/pa_ppx_testutils.cma
Installed /home/opam/.opam/test/.opam-switch/build/pa_ppx.0.07.02/testutils/../local-install//lib/pa_ppx_testutils/META

Program received signal SIGSEGV, Segmentation fault.
caml_darken (v=0, ignored=0x0, state=0x0) at major_gc.c:721
721     major_gc.c: No such file or directory.
(gdb) bt
#0  caml_darken (v=0, ignored=0x0, state=0x0) at major_gc.c:721
#1  0x00005595ea78aae0 in caml_darken (state=state@entry=0x0, v=<optimized out>, ignored=ignored@entry=0x0) at major_gc.c:719
#2  0x00005595ea78da50 in write_barrier (new_val=0, old_val=<optimized out>, field=0, obj=140542418582960) at memory.c:142
#3  caml_modify (fp=0x7fd294ef7db0, val=0) at memory.c:153
#4  0x00005595ea790e7c in caml_install_signal_handler (signal_number=<optimized out>, action=<optimized out>) at signals.c:307
#5  0x00005595ea7ae5aa in caml_interprete (prog=<optimized out>, prog_size=<optimized out>) at interp.c:1029
#6  0x00005595ea7afe42 in caml_main (argv=0x7ffc113c2448) at startup_byt.c:381
#7  0x00005595ea78984c in main (argc=<optimized out>, argv=<optimized out>) at main.c:41
```
The rest of opam-repository worked fine however.

Sudha247:
I have rebased this PR on top of `4.12+domains+effects` in #516. Closing this one.

Issue #471, opened 2021-02-03
Unix.fork fails with "unlock: Operation not permitted"


kit-ty-kate:
Full logs:
* http://check.ocamllabs.io:8082/log/1610933656-df43578ae336c14c63a6d491d406bb8da762395d/4.10+multicore/bad/dune.1.9.1
* http://check.ocamllabs.io:8082/log/1610933656-df43578ae336c14c63a6d491d406bb8da762395d/4.10+multicore/bad/omake.0.10.3
* http://check.ocamllabs.io:8082/log/1610933656-df43578ae336c14c63a6d491d406bb8da762395d/4.10+multicore/bad/rdf.0.11.0

All end up failing with:
```
# Fatal error during unlock: Operation not permitted
```
This is an issue with `Unix.fork` and seems to have been caused by https://github.com/ocaml-multicore/ocaml-multicore/pull/455/files#diff-f7ddfd5d7014ce34ffd8b1c4431a6e201fda084ff6b890933bef07e2f71a305a
Here is the strace:
```
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fc329e5ae90) = 17919
[pid 17918] fstat(4, {st_mode=S_IFIFO|0600, st_size=0, ...}) = 0
[pid 17918] lseek(4, 0, SEEK_CUR)       = -1 ESPIPE (Illegal seek)
[pid 17919] set_robust_list(0x7fc329e5aea0, 24 <unfinished ...>
[pid 17918] read(4,  <unfinished ...>
[pid 17919] <... set_robust_list resumed>) = 0
[pid 17919] write(2, "Fatal error during unlock", 25Fatal error during unlock) = 25
[pid 17919] write(2, ": Operation not permitted\n", 26: Operation not permitted
) = 26
```
_Sorry I meant to create this issue earlier but forgot._

abbysmal:
So our intuition that it was related to 455 was founded.
The idea is that 455 is still a valid fix, however it did break the state of locks within the runtime after the fork: the domain lock is not in a valid state after the fork (in the child process), leading to any subsequent use of this lock to fail. (in this specific scenario, `Unix.close` enters a blocking section and thus releases the domain lock, failing.)
Smaller repro:

```
let () =
  let fd = Unix.dup Unix.stdout in
  let ret = Unix.fork () in
  if ret = 0 then
    Unix.close fd
  else
[...]

PR #491, opened 2021-03-12
Reinitialize child locks after fork

abbysmal:
This PR aims to fix fork issues, following up on #455 and #471.

Currently, forking on OCaml Multicore is only allowed when there is only one domain running.
However, specific care is required to allow the runtime to operate properly after a fork.

Various locks need to be reset in the child for it to operate properly (in this case, the domain lock).

#455 observed that under specific conditions, entering a blocking section during fork could allow a systhread to take over the lock and execute code or runtime duty while the process is being forked, leading to mayhem.

#455 aligned with trunk and opted out of entering a blocking section, fixing this issue, while also making sure the locks were properly reset in `caml_thread_reinitialize`, called through `pthreads_atfork1.

This however left a problem in the case a program is not using systhreads, where the domain lock would never be reset nor acquired, meaning that the next time a program tries to release it (when entering a blocking section in #471, for example), the program would crash.

This PR attempt at finding a solution that can account for both cases, and reset the runtime locks after fork.

After talking a bit with @ctk21, we opted to try to introduce an `atfork` hook in the runtime, that will be called by `unix_fork`.
This way the `atfork` duties can be handled in a simple way on a case by case basis. (and extended further)
 
 This also means that we do not use `pthreads_atfork` anymore, and I'm unsure of the implications for this. (this may be a problem for programs forking in a convoluted way, clearing away from the path of using the core wrapper shipped with the OCaml distribution.)
 
 Testcases were also added to cover the bug in #471, and some more.


abbysmal:
@ctk21 also mentioned that we should straighten the invariant on `unix_fork` and check the program has also no backup_threads running.
This will be done in a later PR with the appropriate testcases.

(and also thank you @kit-ty-kate for investigating that issue!)

Engil:
@ctk21 also mentioned that we should straighten the invariant on `unix_fork` and check the program has also no backup_threads running.
This will be done in a later PR with the appropriate testcases.

(and also thank you @kit-ty-kate for investigating that issue!)

ctk21:
We had a side discussion because there is a slight change in how things might behave. With this PR if the program calls fork from the FFI instead of with `unix_fork`, then things might not work out well. One option might be to have a `caml_plat_atfork` which could be used for the runtime lock and called again to add the systhread locks, but there was a concern that would not play well on non-Unix platforms. We will go with `caml_atfork_hook` and calling it from `unix_fork` for now as in this PR. 

This PR fixes #471 and has been tested with opam-ci (thanks to @kit-ty-kate), so we will merge and move on for now.

PR #492, opened 2021-03-13
Make the implementation of lazy aware of safe points.

### LINKED ISSUE #483: Lazy.try_force segfaults on multiple domains
This testcase tweaked from `lazy5.ml` in the testsuite, results in a segmentation fault, especially when the number of domains is high. The failure was first observed in this [CI failure](https://github.com/ocaml-multicore/ocaml-multicore/runs/2005271740). The code below fails more reliably than `lazy5.ml`

```ocaml
let num_domains = 16

let rec safe_force l =
  match Lazy.try_force l with
  | Some x -> x
  | None -> (
      Domain.Sync.cpu_relax () ;
      safe_force l
    )

let f count =
  let _n = (Domain.self ():> int) in
  let r = ref 0 in
  for _ = 1 to count do
    incr r;
  done;
  !r

let l = lazy (f 1_000_000_000)
let d1 = Array.init (num_domains - 1) (fun _->
  Domain.spawn (fun () ->
      let _n = (Domain.self ():> int) in
      safe_force l))
let n2 = safe_force l
let n1 = Array.map Domain.join d1

let _ = Printf.printf "n1=%d n2=%d\n" n1.(0) n2
```

kayceesrk:
The introduction of safe points at function entry may cause GC to run,
which may short circuit lazy values. This commit removes an assumption
about the structure of the lazy value `update_tag_forcing` and makes it
handle the possibility of GC running in the function prologue.

Fixes #483.

### DIFF
--- a/stdlib/camlinternalLazy.ml
+++ b/stdlib/camlinternalLazy.ml
@@ -53,13 +53,19 @@ let do_force_val_block blk =
   make_forward b (Obj.repr result);
   result
 
-type status = Racy | Forcing
+type 'a status = Racy | Forcing | Done of 'a
 
-(* Assumes [blk] is a block with tag lazy *)
+(* [blk] might either be a lazy block with any tag (due to parallelism) or a
+ * short-circuited object (when the poll point inserted at the entry to this
+ * function performs a GC, which short-circuits the lazy). Hence, we need to
+ * repeat the tag checks again. *)
 let update_tag_forcing (blk : 'arg lazy_t) =
   let b = Obj.repr blk in
-  if not (update_tag b Obj.lazy_tag Obj.forcing_tag) then
-    (* blk has tag either
+  let t = Obj.tag b in
+  if t <> Obj.lazy_tag && t <> Obj.forcing_tag && t <> Obj.forward_tag then
+    Done (Obj.obj b : 'arg)
+  else if not (update_tag b Obj.lazy_tag Obj.forcing_tag) then begin
+    (* [update_tag] has failed. [blk] has tag that is either
         + Obj.forcing_tag -- currently being forced by this domain or
                              another concurrent domain (or)
         + Obj.forward_tag -- was being forced by another domain which has since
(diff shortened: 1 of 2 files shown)


stdlib/camlinternalLazy.ml
```diff
@@ -53,31 +53,40 @@ let do_force_val_block blk =
...
+ * short-circuited object (when the poll point inserted at the entry to this
+ * function performs a GC, which short-circuits the lazy). Hence, we need to
+ * repeat the tag checks again. *)
 let update_tag_forcing (blk : 'arg lazy_t) =
   let b = Obj.repr blk in
-  if not (update_tag b Obj.lazy_tag Obj.forcing_tag) then
-    (* blk has tag either
+  let t = Obj.tag b in
+  if t <> Obj.lazy_tag && t <> Obj.forcing_tag && t <> Obj.forward_tag then
+    Done (Obj.obj b : 'arg)
```
Sudha247, line 66: This takes care of the case where the object is no longer a block. Does it make sense to add an assert here or in `obj_update_tag` to be sure we are operating with a block?

Sudha247 (commented):
The changes look good to me and I believe this fixes #483. 

It'll be useful to have a testcase with a relatively larger number of domains (say 16) trying to force the same lazy value. `lazy5` currently calls `try_force` from two domains. Could we edit `lazy5` or add a new test to cover this?

stdlib/camlinternalLazy.ml
```diff
@@ -53,31 +53,40 @@ let do_force_val_block blk =
...
+  else if not (update_tag b Obj.lazy_tag Obj.forcing_tag) then begin
```
ctk21, line 67: Surely we would have the same bug if the safepoint gets inserted just before the `if not ...` test?

ctk21 (commented):
I think this fixes the problem if the safepoint is inserted at the entry point to the function. There is a gnarly thing here, I'm not convinced the implementation is right if a safepoint can be inserted anywhere. 

Are we able to:
- put the testcode in #483 into the testsuite just in case someone moves something around in the safepoints, Lazy GC or lazy stdlib code
- document the assumption on the safepoint inserts in the problematic `update_tag_forcing` function
- open an issue highlighting that we need a less brittle implementation for Lazy in the presence of arbitrary safepoints; e.g. we could explore pushing a bit more into the C function using the atomics to avoid the issue and have a bit less in the stdlib side.

stdlib/camlinternalLazy.ml
kayceesrk: Unfortunately, `assert` in OCaml always runs. It would be nice if it were to run in a hypothetical `debug` model. But it should be ok to do that here since this is the slow path; fast path is where the lazy has already been forced.

kayceesrk:
Thanks for the useful comments. It may be useful to what @ctk21 suggests, which is to move more functionality to C where we know safepoints will not interfere. Let me clean up the code in this PR to make it more robust.

kayceesrk:
Closing in favour of #501.

PR #493, opened 2021-03-14
Revert remove RacyLazy

kayceesrk:
With multicore, lazy values can also be forced by multiple domains at the same time and hence, we need to distinguish recursive forcing of lazy values from parallel forcing. In PR#338, we introduced `Lazy.try_force_val` which returns with [None] immediately if a lazy is concurrently forced. But the PR also removed `RacyLazy` exception. It was agreed in the PR discussion to reinstate `RacyLazy` exception https://github.com/ocaml-multicore/ocaml-multicore/pull/338#issuecomment-635144020. This PR does that.


kayceesrk:
Must be merged after #492.

kayceesrk:
Closing in favour of #501.

PR #494, opened 2021-03-15
Stronger invariants on unix_fork

abbysmal:
This PR is a follow up to #491 and aims to have a stronger invariant on `unix_fork`, as suggested by @ctk21 
The current situation is that `unix_fork` will not permit a fork if a domain is currently running alongside the domain 0. (`caml_domain_alone`).

This PR pushes it a bit further, by enforcing that no other domain ever ran alongside domain 0. (by looking for the backup thread).
The idea is that, a runtime that has been _tainted_ by multicore activities is a source of bugs when trying to reset to the proper state in the child process after a fork.
I think it is fair that supporting `fork` for legacy programs is fine, and such programs should preferably either keep the API usage to Multicore (and deprecate fork usage), or not use any of Domain's API at all.

This PR introduce two tests to check that we enforce this property properly, as well as a new `caml_is_multicore` utility in the runtime to check that no other domain was either spawned.
I'm not too fond of the name for this utility, suggestions are welcome. :)


ctk21:
I'm fine with this as is. Might prevent someone chasing round the houses in future and documents the conditions when fork is expected work.

PR #496, opened 2021-03-16
Replace caml_initialize_field with caml_initialize

Sudha247:
This patch replaces `caml_initialize_field` - one of the leftover items from concurrent minor collector that are no longer necessary, with `caml_initialize`.

Note: Some minor conflicts may arise with #488, I shall rebase on top of whichever is landed first.

### DIFF
--- a/runtime/memory.c
+++ b/runtime/memory.c
@@ -179,9 +179,6 @@ CAMLexport CAMLweakdef void caml_modify (value *fp, value val)
                         memory_order_release);
 }
 
-/* Compatability with old C-API
-   bit of a HACK as less Assert possible here
- */
 CAMLexport CAMLweakdef void caml_initialize (value *fp, value val)
 {
 #ifdef DEBUG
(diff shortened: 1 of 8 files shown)


runtime/memory.c
```diff
@@ -179,31 +179,12 @@ CAMLexport CAMLweakdef void caml_modify (value *fp, value val)
...
-    Assert(Op_val(obj)[field] == Debug_uninit_major ||
-           Op_val(obj)[field] == Val_unit);
-#endif
-
-  write_barrier(obj, field, Op_val(obj)[field], val);
-  Op_val(obj)[field] = val;
-}
-
 /* Compatability with old C-API
    bit of a HACK as less Assert possible here
```
ctk21: Minor nit, I think we remove this comment which is historic now?

ctk21 (commented):
Thanks, this will clear up some confusion by removing historic additions to the C-API. 
This looks good to me!

Minor minor nit on the cleaning up the comment for `caml_initialize`.

runtime/memory.c
Sudha247: Yeah, makes sense to remove comments that are no longer relevant. Done.

ctk21 (approved):
This looks good to me and gets rid of a (now unnecessary) C-API function.

PR #500, opened 2021-03-17
Replace caml_modify_field with caml_modify

Sudha247:
One more patch removing historic C-functions that are not required anymore. This one replaces all usages of `cam_modify_field` with `caml_modify` and remove `cam_modify_field`.

### DIFF
--- a/runtime/memory.c
+++ b/runtime/memory.c
@@ -96,7 +96,7 @@
    Using release stores for all writes also ensures publication safety
    for newly-allocated objects, and isn't necessary for initialising
    writes. The cost is free on x86, but requires a fence in
-   caml_modify_field on weakly-ordered architectures (ARM, Power).
+   caml_modify on weakly-ordered architectures (ARM, Power).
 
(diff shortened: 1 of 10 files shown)


runtime/memory.c
```diff
@@ -313,7 +295,7 @@ CAMLexport void caml_set_fields (value obj, value v)
   Assert (Is_block(obj));
```
ctk21, line 296: I think we should keep the mutable_stores stat collection:
```C
#if defined(COLLECT_STATS) && defined(NATIVE_CODE)
  Caml_state->mutable_stores++;
#endif
```
Cheap to maintain here and useful while `COLLECT_STATS` exists as an option.

ctk21 (commented):
Thanks, this looks good in the whole. 

One minor things:
 - I found one more instance of `caml_modify_field` in the memory model comment in memory.c; are we able to just rename that one to `caml_modify` as well?

ctk21:
Thanks, LGTM.

PR #501, opened 2021-03-17
Safepoints lazy fix -- Attempt 2

kayceesrk:
This supplants #492 and #493. This also addresses comments in https://github.com/ocaml-multicore/ocaml-multicore/pull/492#pullrequestreview-612034443.

* Adds the test from #483 as `lazy7.ml`
* The PR moves more of the internal implementation of lazy into the runtime so that we don't make assumptions about atomicity w.r.t the GC. 

## Notes for reviewing

The first commit does the bulk of the work. It may be useful to review that commit first.


stdlib/camlinternalLazy.ml
```diff
@@ -50,43 +83,37 @@ let do_force_val_block blk =
...
 (* used in the %lazy_force primitive *)
 let force_lazy_block blk = force_gen_lazy_block ~only_val:false blk
 
-let try_force_gen_lazy_block ~only_val blk =
-  match update_tag_forcing blk with
-  | Racy -> None
-  | Forcing when only_val -> Some (do_force_val_block blk)
-  | Forcing -> Some (do_force_block blk)
+let try_force_gen_lazy_block ~only_val (blk : 'arg lazy_t) =
+  match update_to_forcing (Obj.repr blk) with
```
Sudha247, line 109: I suspect it's possible to run into the same error we were previously witnessing here.

runtime/obj.c
```diff
@@ -250,6 +212,74 @@ CAMLprim value caml_lazy_make_forward (value v)
...
+    return Field(v,0);
+  return v;
+}
+
+CAMLprim value caml_lazy_update_to_forcing (value v)
+{
+  tag_t tag;
+  value field0;
+
+  if (obj_update_tag (v, Lazy_tag, Forcing_tag)) {
```
Sudha247, line 272: I think we should be making sure `v` is still a block before we pass it to `obj_update_tag`. In the case that `v` is not a block, this could lead to the same error we were witnessing previously. We should probably go over the other checks first before calling `obj_update_tag`.

Sudha247 (commented):
Thank you, this is a great improvement over the existing implementation! Moving bits to the runtime is helpful. 

I just had one suggestion on `update_to_forcing`.

runtime/obj.c
kayceesrk: That's correct. Good catch.

kayceesrk:
@Sudha247 I've fixed the bug. I've also made other improvements in https://github.com/ocaml-multicore/ocaml-multicore/pull/501/commits/8217553e5934feff38e90b63faea52815f2cf221. Please review.

stdlib/camlinternalLazy.ml
```diff
@@ -50,58 +83,56 @@ let do_force_val_block blk =
...
+       * in [force_gen_lazy_block]. *)
+      None
 
 (* [force_gen ~only_val:false] is not used, since [Lazy.force] is
    declared as a primitive whose code inlines the tag tests of its
    argument. This function is here for the sake of completeness, and
    for debugging purpose. *)
 let force_gen ~only_val (lzv : 'arg lazy_t) =
   let x = Obj.repr lzv in
+  (* XXX No safe points start *)
```
Sudha247, line 124: Just so I'm not misreading this comment: Does this mean we are assuming there are no safe points in this range or that the compiler will not insert any safe points?

Sudha247 (approved):
Great! This fixes the bug in #483. The changes look good to me.

I had a question on a comment, but that shouldn't block merging this.

stdlib/camlinternalLazy.ml
kayceesrk: We are assuming that the compiler will not insert safe points in this block. The compiler only inserts safe points at function entries and back edges of loops now. If this behaviour were to change, then this code will break. This is documenting the assumptions that hold now.

PR #505, opened 2021-03-22
Add a unique domain token to distinguish lazy forcing failure

### LINKED ISSUE #504: Using domain id (int) to identify forcing domain of lazy block is wrong
@Sudha247 Observed a failure in `lazy6.ml` in the CI run during merge of #501. https://github.com/ocaml-multicore/ocaml-multicore/runs/2147827717. This is a real bug and has to do with the value we store in a lazy block currently being forced. 

## Bug

In order to handle racy access by multiple mutators, lazy blocks use compare-and-swap (CAS) to update the tag of the block from `Lazy_tag` to `Forcing_tag`. Once successful, the first field of the object, which used to contain the closure for the lazy computation, is updated with the current domain id (int). This information is used to distinguish between recursive and concurrent forcing of lazy values. For example, when forcing a lazy value, if the compare-and-swap (CAS) fails and the tag is `Forcing_tag`, we perform a **racy** read of the first field, and compare that with the current domain id. If the domain id, is the same, then it indicates recursive forcing (`Lazy.force` raises `Undefined`). Otherwise, it is parallel forcing (`Lazy.force` raises `RacyLazy`). 

The trouble is when the domain that is currently forcing the lazy computation computes the result, it first updates the first field to the result and then updates the `Forcing_tag` to `Forward_tag`. In `lazy6.ml`, the result is `()`, which has the same representation as the integer value `0`. The second domain mistakes the final `()` value for domain id `0`. This domain now raises `Undefined` exception.

## Proposed fix

The crux of the problem is that we need to distinguish the domain id that we store in the lazy value under evaluation from any result value. Since the result could be **any** value, the domain id has to be different from all of the possible values.

The trick would be to use the freshness offered by the OCaml heap. Instead of using the integer domain id, we shall allocate a 1-word object on the heap when the domain is created. Let's call this `unique_token`. This unique token cannot be accessed by regular OCaml programs. The only way to access this token would be through `caml_domain_unique_id`. This token is stored in the Lazy block that is currently being forced rather than the integer domain id. In order to check whether the current domain is the same as the one that forced this lazy earlier, we compare the addresses of the saved token and the current domain's token. 

The unique token need not be kept alive at domain termination. Consider the case when a domain is in the middle of forcing a lazy and terminates. The token is kept alive thanks to that being referenced from the lazy object. As long as the token is alive, no other domain will get the same unique token. Since we use physical (address) equality to compare the token, by definition, if the token has been GCed, then there are no reference to such a token from a live domain or any lazy values. And hence, it is safe to leave the token to be GCed at domain termination.

Regarding implementation, one may implement the token completely from the OCaml side using domain-local state. But I wonder if this is a good idea to make the performance of lazy depend on the number of domain local variables.

@ctk21 @Sudha247 thoughts?

kayceesrk:
Fixes #504.

### DIFF
--- a/runtime/caml/domain_state.tbl
+++ b/runtime/caml/domain_state.tbl
@@ -116,7 +116,7 @@ DOMAIN_STATE(int, pools_to_rescan_count)
 DOMAIN_STATE(caml_root, dls_root)
 /* Domain-local state */
 
-DOMAIN_STATE(caml_root, unique_token_root)
+DOMAIN_STATE(value, unique_token_root)
 
 /*****************************************************************************/
 /* GC stats (see gc_ctrl.c and the Gc module) */
--- a/runtime/caml/domain_state.tbl
+++ b/runtime/caml/domain_state.tbl
@@ -114,6 +114,9 @@ DOMAIN_STATE(int, pools_to_rescan_len)
 DOMAIN_STATE(int, pools_to_rescan_count)
 
 DOMAIN_STATE(caml_root, dls_root)
+/* Domain-local state */
+
+DOMAIN_STATE(caml_root, unique_token_root)
 
 /*****************************************************************************/
 /* GC stats (see gc_ctrl.c and the Gc module) */
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -278,7 +278,13 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
 
     domain_state->dls_root = caml_create_root_noexc(Val_unit);
     if(domain_state->dls_root == NULL) {
-      goto create_root_failure;
+      goto create_dls_root_failure;
+    }
+
+    domain_state->unique_token_root =
+      caml_create_root_noexc(caml_alloc_1(Abstract_tag,Val_unit));
+    if(domain_state->unique_token_root == NULL) {
+      goto create_unique_token_root_failure;
     }
 
     domain_state->stack_cache = caml_alloc_stack_cache();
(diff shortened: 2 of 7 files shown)


runtime/domain.c
```diff
@@ -278,7 +278,13 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
 
     domain_state->dls_root = caml_create_root_noexc(Val_unit);
     if(domain_state->dls_root == NULL) {
-      goto create_root_failure;
+      goto create_dls_root_failure;
+    }
+
+    domain_state->unique_token_root =
+      caml_create_root_noexc(caml_alloc_1(Abstract_tag,Val_unit));
```
Sudha247: Can we use generational global roots here please? With #488 `caml_root` should be gone.

runtime/domain.c
```diff
@@ -1249,6 +1264,7 @@ static void domain_terminate()
 
   caml_gc_log("Domain terminating");
   caml_delete_root(domain_state->dls_root);
+  caml_delete_root(domain_state->unique_token_root);
```
Sudha247: As @ctk21 pointed out in #488, the while block below this could run mutator code, so is it better to move the deletion to after the while block?

runtime/caml/domain_state.tbl
```diff
@@ -114,6 +114,9 @@ DOMAIN_STATE(int, pools_to_rescan_len)
 DOMAIN_STATE(int, pools_to_rescan_count)
 
 DOMAIN_STATE(caml_root, dls_root)
+/* Domain-local state */
+
+DOMAIN_STATE(caml_root, unique_token_root)
```
Sudha247: This needs to be a `value` if we're using generational global roots.

Sudha247 (commented):
Thanks, this should fix the issue in #504. I concur on not using domain-local state for storing the unique token, since that would make performance unpredictable.

Left some comments below, it's majorly along the lines of sticking to using generational global roots to be in sync with #488.

Sudha247 (approved):
Thanks for the updates! This looks good to me.

PR #506, opened 2021-03-23
Replace `Op_val` with `Field`

Sudha247:
Changes usages of `Op_val (x)[i]` to `Field (x, i)`. `Op_val` was supposedly used at a time `Field` was not working with multicore. They are both same and the latter seems to be more widely used in trunk.

### DIFF
--- a/runtime/floats.c
+++ b/runtime/floats.c
@@ -82,8 +82,8 @@ CAMLexport double caml_Double_val(value val)
   union { value v[2]; double d; } buffer;
 
   CAMLassert(sizeof(double) == 2 * sizeof(value));
-  buffer.v[0] = Op_val(val)[0];
-  buffer.v[1] = Op_val(val)[1];
+  buffer.v[0] = Field(val, 0);
+  buffer.v[1] = Field(val, 0);
   return buffer.d;
 }
 
@@ -93,8 +93,8 @@ CAMLexport void caml_Store_double_val(value val, double dbl)
 
   CAMLassert(sizeof(double) == 2 * sizeof(value));
   buffer.d = dbl;
-  Op_val(val)[0] = buffer.v[0];
-  Op_val(val)[1] = buffer.v[1];
+  Field(val, 0) = buffer.v[0];
+  Field(val, 0) = buffer.v[1];
 }
 
 #endif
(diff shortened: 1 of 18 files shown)


runtime/floats.c
```diff
@@ -82,8 +82,8 @@ CAMLexport double caml_Double_val(value val)
   union { value v[2]; double d; } buffer;
 
   CAMLassert(sizeof(double) == 2 * sizeof(value));
-  buffer.v[0] = Op_val(val)[0];
-  buffer.v[1] = Op_val(val)[1];
+  buffer.v[0] = Field(val, 0);
+  buffer.v[1] = Field(val, 0);
```
ctk21: This looks like a typo, shouldn't it be:
```c
buffer.v[1] = Field(val, 1);
```

runtime/floats.c
```diff
@@ -93,8 +93,8 @@ CAMLexport void caml_Store_double_val(value val, double dbl)
 
   CAMLassert(sizeof(double) == 2 * sizeof(value));
   buffer.d = dbl;
-  Op_val(val)[0] = buffer.v[0];
-  Op_val(val)[1] = buffer.v[1];
+  Field(val, 0) = buffer.v[0];
+  Field(val, 0) = buffer.v[1];
```
ctk21: This looks like a typo, shouldn't it be:
```C
Field(val, 1) = buffer.v[1];
```

runtime/minor_gc.c
```diff
@@ -341,16 +341,16 @@ static void oldify_one (void* st_v, value v, value *p)
     st->live_bytes += Bhsize_hd(hd);
     result = alloc_shared(sz, tag);
     for (i = 0; i < sz; i++) {
-      value curr = Op_val(v)[i];
-      Op_val (result)[i] = curr;
+      value curr = Field(v, i);
```
ctk21: Do we need the local variable `value curr`?

ctk21 (commented):
Couple of places that need fixing which look like typos. Otherwise good.

runtime/floats.c
Sudha247: They were typos indeed, well spotted, thanks!

runtime/minor_gc.c
Sudha247: You're right, we don't need it.

ctk21:
Thanks! merging

PR #508, opened 2021-03-23
Domain local allocation buffers

sadiqj:
This PR extends https://github.com/ocaml-multicore/ocaml-multicore/pull/484 

1. We now resize the global minor heap lazily based on the number of running domains participating in a minor collection
2. Memory is only committed/released when the number of running domains changes, which should be rarely at steady state
3. Domain local allocation buffers are set to 1/8th of the global minor per _domain_ heap size and are replenished as necessary
4. It has been rebased on to 4.12

There is currently one failing test locally, `timing.ml`. I'm still trying to figure out what's going on with that.


runtime/minor_gc.c
```diff
@@ -518,13 +519,97 @@ void caml_empty_minor_heap_domain_clear (struct domain* domain, void* unused)
...
   clear_table ((struct generic_table *)&minor_tables->custom);
 
+  return;
+}
+
+void caml_adjust_global_minor_heap(int participating_domains) {
+  // Calculate the size of the existing mapping
+  int global_minor_heap_size = caml_global_minor_heap_limit - caml_global_minor_heap_start;
+  // Now using the number of participating domains, we calculate the new size
+  int new_global_minor_heap_size = participating_domains*Bsize_wsize(global_minor_heap_wsz_per_domain) + caml_global_minor_heap_start;
```
abbysmal: I think adding `caml_global_minor_heap_start` wasn't meant here
It just works because everything is page aligned (and thus, is effectively adding zero to the number).

runtime/minor_gc.c
```diff
@@ -129,11 +131,10 @@ void caml_set_minor_heap_size (asize_t wsize)
...
   caml_domain_state* domain_state = Caml_state;
   struct caml_minor_tables *r = domain_state->minor_tables;
-  if (domain_state->young_ptr != domain_state->young_end) caml_minor_collection ();
 
-  if(caml_reallocate_minor_heap(wsize) < 0) {
-    caml_fatal_error("Fatal error: No memory for minor heap");
-  }
+  global_minor_heap_wsz_per_domain = wsize;
+
+  if (domain_state->young_ptr != domain_state->young_end) caml_minor_collection ();
```
abbysmal: I think we should always trigger a minor collection here, there are conditions where a minor collection won't be hit on next gc poll

abbysmal (commented):
Thanks for this work!
I think the changes are fine, I found a thing or two that looks weird, but overall the core logic looks sane to me.
The minor heap setting bits are a bit tricky, I will address these as I take over the branch. (and I think the testsuite deserves more coverage for these)

Naming wise, I think things make sense now, there's a few config options we may want to change, I will take care of this later as well.

runtime/domain.c
```diff
@@ -272,10 +243,18 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto init_major_gc_failure;
     }
 
-    if(caml_reallocate_minor_heap(initial_minor_heap_wsize) < 0) {
-      goto reallocate_minor_heap_failure;
+    /* setting young_limit and young_ptr to minor_heaps_base
+       to trigger minor_heaps reallocation on GC poll */
+    domain_state->young_start = (char*)caml_global_minor_heap_start;
+    domain_state->young_end = (char*)caml_global_minor_heap_start;
```
gasche: the `minor_heaps_base` naming in the comment is now out of date

runtime/caml/minor_gc.h
```diff
@@ -54,9 +54,8 @@ struct caml_minor_tables {
 
 struct domain;
 
-CAMLextern void caml_minor_collection (void);
+CAMLextern asize_t global_minor_heap_wsz_per_domain;
 
-#ifdef CAML_INTERNALS
```
ctk21, line 59: Is the removal of this `CAML_INTERNALS` guard intentional? 
Aren't some of these functions meant to be guarded to not be visible elsewhere.

runtime/domain.c
```diff
@@ -272,10 +243,18 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto init_major_gc_failure;
     }
 
-    if(caml_reallocate_minor_heap(initial_minor_heap_wsize) < 0) {
-      goto reallocate_minor_heap_failure;
+    /* setting young_limit and young_ptr to minor_heaps_base
+       to trigger minor_heaps reallocation on GC poll */
+    domain_state->young_start = (char*)caml_global_minor_heap_start;
```
ctk21: This function `create_domain` can execute in parallel with STW segments but the domain is not yet part of the STW. It will be brittle to read `caml_global_minor_heap_start` multiple times should we ever change how that variable is assigned. 
Is there any reason to read this variable at all? Can't we just set all the `young_*` values to a suitable `null` value and trigger the collection we need when the values are used?

runtime/domain.c
```diff
@@ -272,10 +243,18 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
...
+    domain_state->young_end = (char*)caml_global_minor_heap_start;
+
+    young_limit = atomic_load_acq((atomic_uintnat*)&domain_state->young_limit);
+    if( young_limit != INTERRUPT_MAGIC ) {
+      atomic_compare_exchange_strong((atomic_uintnat*)&domain_state->young_limit, &young_limit, caml_global_minor_heap_start);
```
ctk21: I think this snippet to update the `young_limit` should use `caml_update_young_limit`

runtime/domain.c
```diff
@@ -932,32 +924,57 @@ void caml_request_minor_gc (void)
...
+  { // No GC in this block
+    int domain_minor_heap_full = ((uintnat)Caml_state->young_ptr - Bhsize_wosize(Max_young_wosize) <
+        (uintnat)Caml_state->young_start);
+
+    int need_minor_gc = Caml_state->requested_minor_gc;
+
+    // Check if our minor heap is full. If it is then we need to try to grab a
+    // new one from the global minor heap.
+    if( domain_minor_heap_full && !need_minor_gc ) {
+      uintnat global_ptr =
```
ctk21: Is there a reason why we need to do this check with the `global_ptr`? We have to deal with that case inside `caml_replenish_minor_heap` already. 

I didn't understand why the body of this `if` block wasn't just:
```C
/* try to get a new minor heap */
if(!caml_replenish_minor_heap()) {
   need_minor_gc = 1;
}
```

runtime/domain.c
```diff
@@ -932,32 +924,57 @@ void caml_request_minor_gc (void)
 static void caml_poll_gc_work()
 {
   CAMLalloc_point_here;
+  { // No GC in this block
```
ctk21: I didn't understand this comment, as there is a call to `caml_empty_minor_heaps_once` inside the block!

runtime/minor_gc.c
```diff
@@ -518,13 +519,97 @@ void caml_empty_minor_heap_domain_clear (struct domain* domain, void* unused)
...
+            memory_order_release);
+}
+
+int caml_replenish_minor_heap()
+{
+  caml_domain_state* domain_state = Caml_state;
+  uintnat cached_global_minor_heap_ptr;
+  uintnat new_alloc_ptr;
+
+  uintnat minor_buffer_wsize = global_minor_heap_wsz_per_domain >> 3;
```
ctk21: I had a concern here that a user might somehow set the minor heap size small and then we would have a minor heap that couldn't hold even the smallest allocation. Do we need a guard here or in the initialization of `global_minor_heap_wsz_per_domain`?

runtime/minor_gc.c
```diff
@@ -518,13 +519,97 @@ void caml_empty_minor_heap_domain_clear (struct domain* domain, void* unused)
...
+
+  // global_minor_heap_ptr is now our new minor heap for this domain
+
+  // I can't think of a situation where requested_buffer_size is not a word multiple
+  domain_state->minor_heap_wsz = Wsize_bsize(requested_buffer_size);
+
+  caml_update_young_limit(cached_global_minor_heap_ptr);
+
+  domain_state->young_start = (char*)cached_global_minor_heap_ptr;
+  domain_state->young_end = (char*)(cached_global_minor_heap_ptr + minor_buffer_bsize);
```
ctk21, line 602: Is this `minor_buffer_bsize` really correct? 
Aren't there code paths here where the `requested_buffer_size` is not equivalent to `minor_buffer_bsize`?

runtime/minor_gc.c
```diff
@@ -683,18 +767,35 @@ void caml_empty_minor_heap_promote (struct domain* domain, int participating_cou
...
+
   domain_state->stat_minor_collections++;
   domain_state->stat_promoted_words += domain_state->allocated_words - prev_alloc_words;
 
   CAML_EV_END(EV_MINOR);
   caml_gc_log ("Minor collection of domain %d completed: %2.0f%% of %u KB live, rewrite: successes=%u failures=%u",
                domain->state->id,
-               100.0 * (double)st.live_bytes / (double)minor_allocated_bytes,
-               (unsigned)(minor_allocated_bytes + 512)/1024, rewrite_successes, rewrite_failures);
+               100.0 * (double)st.live_bytes / (double)(domain_state->stat_minor_words),
```
ctk21: I don't understand this change from `minor_allocated_bytes` to `stat_minor_words`.
The change makes the fraction in units `bytes`/`words`, rather than dimensionless; i.e. wouldn't we need to alter `st.live_bytes` to be _live words_.
The second change probably needs us to alter `KB live` to `K words live`.

runtime/domain.c
abbysmal: Good catch, this case can be simplified indeed.

runtime/minor_gc.c
abbysmal: This change is to take into account that now, we may have used more minor heap space than just one segment (the one currently set to the domain.)
If the domain cycled through many minor heap segment with DLABs, we keep track of these in the replenish cycle: every time we are done doing replenish, we set `stat_minor_words` to `+=` the previous minor heap segment.
This effectively allow us to keep track of the previous minor heap segments used by this domain.
However it means we do need to use `stat_minor_words` instead of the more local `minor_allocated_bytes` (that only knows of the current minor heap segment.).
I think a simpler fix would be to change `(domain_state->stat_minor_words)` to `Bsize_wsize(domain_state->stat_minor_words)` ?

runtime/domain.c
abbysmal: I'm unsure as well, I think it is only meant for the first part of this function? (and a Gc should only be triggered after this preambule, through the `need_minor_gc` flag?) @sadiqj

sadiqj: I've read that comment a few times and I don't understand what it means. Suggest we get rid of it.

ctk21:
I have run this PR (on 20210504 with sandmark) using a large two socket Zen2 machine and some larger than normal benchmark workloads. These are the results I got:
20210504_big_dlabs_time_sec
20210504_big_dlabs_speedup
20210504_big_dlabs_minor_collections

There are some odd things here. We seem to cut the number of minor GCs quite a bit with some of the DLABs run, however in terms of execution time at higher core counts the PR does not seem to improve our runtime. In some instances it is running much slower (e.g. `lu_decomposition`, `minilight`, `spectralnorm2`).

abbysmal:
Here's an excerpt of the bench results (ran by @shubhamkumar13 on godal (IIT machine)).
Provided results are for `256k` sized minor heaps (speedup, time_sec and minor_collections)
speedup
time_secs
minor_collections

ctk21:
Some experiments were conducted to see if the way we were handling the memory allocations of the minor heaps were responsible for the slowdowns on the big instances. The tags in the results below are:
- `4.12.0+domains+effects` baseline
- `4.12.0+domains+effects_dlabs` this PR
- `4.12.0+domains+effects_dlabs_malloc` this PR but where we make use of glibc malloc to allocate all the minor heap and TLS areas. See this [branch](https://github.com/ctk21/ocaml-multicore/tree/ctk21_dlabs_4.12_malloc), results for this [tree 427b02d](https://github.com/ctk21/ocaml-multicore/tree/427b02db319d122aa9eb7a8db54dfed6746c57a2).
- `4.12.0+domains+effects_dlabs_commit_decommit` this PR but where we commit each minor heap segment as we get it and decommit on minor GC (_this is not necessarily sensible, but we wanted to know how bad that is_). See this [branch](https://github.com/ctk21/ocaml-multicore/commits/dlabs_4.12_decommit_commit_experiment), results for this [tree 7e1ca31](https://github.com/ctk21/ocaml-multicore/tree/7e1ca31837c46638b091206217f21945e5427d49).

20210511_commit_decommit_time
20210511_commit_decommit_speedup

The one idea not tried, but might be is to have a 'minor heap stealing' strategy that would be more like HPC 'workstealing':
- each domain has its own minor heap as in the existing baseline code
- when a domain runs out of minor heap, it tries to steal from another domains minor heap
- when all the minor heap space (or a significant fraction) is used, do the minor collection

It does beg the question as to if using non-local minor heap space is ever better on larger machines than just having an early collection.

ctk21:
I have run the benchmarks with a prefetching experiment:
- `4.12+domains+effects` baseline (`b23a41`).
- `4.12+domains+effects+dlabs` this PR (`099a65`).
- `4.12+domains+effects+dlabs_prefetch` this [branch](https://github.com/sadiqj/ocaml-multicore/commits/dlabs_prefetch) with this [tree](https://github.com/sadiqj/ocaml-multicore/tree/f25633bcfba473f79f551905fb226c8d690752c4) (`f25633`)

Here are results with 'big' instances on a detuned Zen2:
20210512_dlabs_prefetch_time
20210512_dlabs_prefetch_speedup
20210512_dlabs_prefetch_mgc

The prefetching is having a beneficial effect. We even see a datapoint with binarytrees where dlabs has out performed at high core counts, offset against it underperforming in the same benchmark at lower core counts without prefetching. Worth noting that the baseline does not perform prefetching on its minor heap, although unclear if that is important. 

(NB: sandmark is slightly updated in this run vs the previous for matrix_multiplication and floyd_warshall; apologies that it is not fixed code there)

ctk21:
We are closing this PR. However we have summarised where we got to [here](https://github.com/ocaml-multicore/ocaml-multicore/wiki/Domain-Local-Allocation-Buffers-Addendum).

PR #511, opened 2021-03-24
Allocate unique root token on the major heap instead of the minor

sadiqj:
Fixes bug in https://github.com/ocaml-multicore/ocaml-multicore/pull/505

In `create_domain` we are allocating the unique root token on the minor heap, this potentially opens us up to ending up in a stop-the-world pause with a partially constructed Domain.

This PR changes that to a major heap allocation that does not raise an exception and bails on Domain creation if it fails.

Noticed because it causes the DLABs PR to deadlock in tests.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -281,7 +281,11 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto create_dls_root_failure;
     }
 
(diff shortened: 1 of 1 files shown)


Sudha247 (commented):
Looks good, thanks! This should fix the testsuite hanging, I'll merge this.

PR #513, opened 2021-03-25
Clear the minor heap at the end of a collection if we're in the debug runtime

sadiqj:
It is currently possible to miss bugs where values in the minor heap that should be promoted are not. This happens because the data is still present in the minor heap before being over-written by a future minor allocation.

This PR clears the minor heap at the end of a minor collection and so enables us to catch this class of bugs. We actually write a debug value in to every element of the minor heap, which is useful for debugging failures.

There will be two commits on this PR. One that clears the minor heap and a second that disables the two failing tests. I'll link an issue to investigate these.

### DIFF
--- a/runtime/minor_gc.c
+++ b/runtime/minor_gc.c
@@ -517,6 +517,14 @@ void caml_empty_minor_heap_domain_clear (struct domain* domain, void* unused)
   clear_table ((struct generic_table *)&minor_tables->major_ref);
   clear_table ((struct generic_table *)&minor_tables->ephe_ref);
   clear_table ((struct generic_table *)&minor_tables->custom);
+
+#ifdef DEBUG
+  {
+    uintnat* p = (uintnat*)domain_state->young_start;
+    for (; p < (uintnat*)domain_state->young_end; p++)
+      *p = Debug_uninit_align;
+  }
+#endif
 }
 
 void caml_empty_minor_heap_promote (struct domain* domain, int participating_count, struct domain** participating, int not_alone)
--- a/testsuite/disabled
+++ b/testsuite/disabled
@@ -85,4 +85,4 @@ tool-debugger
 tests/promotion/bigrecmod.ml
 
 # disabled as there's a bug with signal handlers
-tests/callback/test_signalhandlers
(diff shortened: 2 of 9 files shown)


runtime/minor_gc.c
```diff
@@ -682,6 +682,15 @@ void caml_empty_minor_heap_promote (struct domain* domain, int participating_cou
   domain_state->stat_minor_collections++;
   domain_state->stat_promoted_words += domain_state->allocated_words - prev_alloc_words;
 
+#ifdef DEBUG
+  caml_global_barrier();
```
ctk21: I'm a bit worried about changes in behaviour by having barriers when in debug mode. 
Can we move this block into `caml_empty_minor_heap_domain_clear` (which is called after the `EV_MINOR_LEAVE_BARRIER`) and do away with the barrier in debug mode?

ctk21 (commented):
I'm in favour of scrubbing the minor heap in DEBUG mode to flush out bugs. 

I have a concern about introducing barriers when in DEBUG mode. I think we can avoid it by moving where the scrub happens.

runtime/minor_gc.c
sadiqj: That was my initial plan but I couldn't satisfy myself there weren't any references in to the minor heap from other domains at that point - as there isn't a barrier.

Though the fact there could still be references (e.g from the finalisers) is a bug in itself?

ctk21: So `caml_empty_minor_heap_domain_clear` is called on the other side of the minor GC leave barrier (see `EV_MINOR_LEAVE_BARRIER`) with the opportunistic work. Anything on the other side of that barrier can't reference data in any minor heap and is a bug. 

`caml_empty_minor_heap_domain_clear` currently scrubs the remembered set structures, updates which finalisers might be young, etc. So I think it is the right place to do more debug scrubbing.

sadiqj: Ok, that makes sense. Will move it over there.

ctk21:
LGTM. We have issue #517 which this extra debug code has revealed.

Merging so that this helps with DLABs.

PR #514, opened 2021-03-25
Update instructions in ocaml-variants.opam

dra27:
- Use a local switch instead of a global one
- Use the multicore opam repo
- Tweak the make command, now multicore is on 4.12

These instructions should probably be somewhere else!

cc @ctk21


configure.ac
```diff
@@ -568,7 +568,7 @@ AS_CASE([$ocaml_cv_cc_vendor],
   cc_warnings='-Wall -Wdeclaration-after-statement'])
 
 AS_CASE([$enable_warn_error,AC_PACKAGE_VERSION],
-  [yes,*|,*+dev*|,*+multicore*],
+  [yes,*|,*+dev*|,*+domains*],
```
ctk21, line 571: Wouldn't we need to also update the string in the `VERSION` file to match?

There is a decision here, if we change the `VERSION`:
- we might break things downstream expecting `+multicore` in the string (might not be a concern)
- we would want different version strings for `+domains` and `+domains+effects` on the branches

dra27: Yes - I think originally that was expected in another PR... it's now done here too!

dra27:
I added a commit removing the addition of `--disable-warn-error` from the opam file - i.e. if you've pinned multicore, you get warnings as errors, if you've installed it from the opam-repository package, you don't.

ctk21:
I think this one is good to merge.

PR #516, opened 2021-03-26
Remove caml_root API

Sudha247:
This is essentially #488 rebased to `4.12+domains+effects`.

Thanks to @engil, the patch now includes a test for backtrace last exception.

### DIFF
--- a/runtime/backtrace.c
+++ b/runtime/backtrace.c
@@ -29,11 +29,6 @@
 #include "caml/debugger.h"
 #include "caml/startup.h"
 
-void caml_init_backtrace(void)
-{
-  return;
-}
-
 /* Start or stop the backtrace machinery */
 CAMLprim value caml_record_backtrace(value vflag)
 {
@@ -43,10 +38,9 @@ CAMLprim value caml_record_backtrace(value vflag)
     Caml_state->backtrace_active = flag;
     Caml_state->backtrace_pos = 0;
     if (flag) {
-      Caml_state->backtrace_last_exn = caml_create_root(Val_unit);
+      caml_modify_generational_global_root(&Caml_state->backtrace_last_exn, Val_unit);
     } else {
-      caml_delete_root(Caml_state->backtrace_last_exn);
-      Caml_state->backtrace_last_exn = NULL;
+      caml_remove_generational_global_root(&Caml_state->backtrace_last_exn);
     }
   }
   return Val_unit;
@@ -205,8 +199,7 @@ CAMLprim value caml_restore_raw_backtrace(value exn, value backtrace)
 
   caml_domain_state* domain_state = Caml_state;
 
-  if (domain_state->backtrace_last_exn != NULL)
-    caml_modify_root (domain_state->backtrace_last_exn, exn);
+  caml_modify_generational_global_root (&domain_state->backtrace_last_exn, exn);
 
   bt_size = Wosize_val(backtrace);
   if(bt_size > BACKTRACE_BUFFER_SIZE){
--- a/runtime/backtrace.c
+++ b/runtime/backtrace.c
@@ -41,6 +41,7 @@ CAMLprim value caml_record_backtrace(value vflag)
       caml_modify_generational_global_root(&Caml_state->backtrace_last_exn, Val_unit);
     } else {
       caml_remove_generational_global_root(&Caml_state->backtrace_last_exn);
+      Caml_state->backtrace_last_exn = (value) NULL;
     }
   }
   return Val_unit;
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -276,10 +276,11 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto reallocate_minor_heap_failure;
     }
 
-    domain_state->dls_root = caml_create_root_noexc(Val_unit);
-    if(domain_state->dls_root == NULL) {
+    domain_state->dls_root = Val_unit;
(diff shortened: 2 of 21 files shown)


kit-ty-kate:
Does this also include the segfault described in https://github.com/ocaml-multicore/ocaml-multicore/pull/488#issuecomment-805136874 or has this been fixed?

Sudha247:
>Does this also include the segfault described in #488 (comment) or has this been fixed?

The issue is still there, I've added a description in #518. Feel free to comment there if you'd like to add anything else there.

runtime/signals.c
```diff
@@ -296,14 +297,16 @@ CAMLprim value caml_install_signal_handler(value signal_number, value action)
     res = Val_int(1);
     break;
   case 2:                       /* was Signal_handle */
-    caml_read_field(caml_read_root(caml_signal_handlers), sig, &handler);
+    caml_read_field(caml_signal_handlers, signal_number, &handler);
```
ctk21: I think we might have a typo here. Shouldn't this be:
`caml_read_field(caml_signal_handlers, sig, &handler);`

runtime/backtrace.c
```diff
@@ -43,10 +38,9 @@ CAMLprim value caml_record_backtrace(value vflag)
     Caml_state->backtrace_active = flag;
     Caml_state->backtrace_pos = 0;
     if (flag) {
-      Caml_state->backtrace_last_exn = caml_create_root(Val_unit);
+      caml_modify_generational_global_root(&Caml_state->backtrace_last_exn, Val_unit);
     } else {
-      caml_delete_root(Caml_state->backtrace_last_exn);
-      Caml_state->backtrace_last_exn = NULL;
+      caml_remove_generational_global_root(&Caml_state->backtrace_last_exn);
```
ctk21, line 43: Might be a good idea to NULL `Caml_state->backtrace_last_exn` at the point we delete it. 
Just makes it a bit easier to debug things should a mistake occur somewhere.

runtime/domain.c
```diff
@@ -276,10 +276,11 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto reallocate_minor_heap_failure;
     }
 
-    domain_state->dls_root = caml_create_root_noexc(Val_unit);
-    if(domain_state->dls_root == NULL) {
+    domain_state->dls_root = Val_unit;
+    if(domain_state->dls_root == (value)NULL) {
```
ctk21: Can't we remove this branch and the label `create_dls_root_failure`?
(`domain_state->dls_root` will always be `Val_unit` and never `NULL`)

ctk21 (commented):
I think this is nearly there. 

Correctness wise, it looks like a typo in the `caml_install_signal_handler` where we should use `sig`.

Sudha247:
Looks like https://github.com/ocaml-multicore/ocaml-multicore/pull/516/commits/98a7c3a066864ea76e0f8e3b8d916f3087bfb953 fixes #518.

ctk21 (approved):
Visually I didn't spot any correctness problems. I think this is now good to merge.

runtime/interp.c
```diff
@@ -687,23 +690,25 @@ value caml_interprete(code_t prog, asize_t prog_size)
...
+      accu = Field(caml_global_data, *pc);
       pc++;
       accu = Field(accu, *pc);
       pc++;
       Next;
     }
 
     Instruct(SETGLOBAL):
-      caml_modify(&Field(caml_read_root(caml_global_data), *pc), accu);
+      global_data = caml_global_data;
```
kayceesrk: This look "interesting". Perhaps a comment would be useful. Is this the only place where `global_data` is used?

runtime/interp.c
```diff
@@ -218,7 +218,8 @@ sp is a local copy of the global variable Caml_state->extern_sp. */
 static __thread intnat caml_bcodcount;
 #endif
 
-static caml_root raise_unhandled;
+static value raise_unhandled;
+value global_data;
```
kayceesrk: I might have read it too fast, but is there a reason for `global_data` to be a C global?

runtime/interp.c
Sudha247: Yeah, this is the only place `global_data` is used. It's just an intermediate location to store `caml_global_data` and call `caml_modify`, since we can't use `caml_modify` directly on `caml_global_data` and will need to use `caml_modify_generational_global_root` to update it. Will add a comment conveying this.

runtime/interp.c
Sudha247: You're right, this need not be a C global. Will move its declaration to where it's used.

runtime/interp.c
```diff
@@ -687,26 +689,31 @@ value caml_interprete(code_t prog, asize_t prog_size)
...
       pc++;
       Next;
     }
 
-    Instruct(SETGLOBAL):
-      caml_modify(&Field(caml_read_root(caml_global_data), *pc), accu);
+    Instruct(SETGLOBAL):  {
+      /* temporary variable used in caml_modify,
+         caml_modify cannot be called directly on generational roots */
+      value global_data = caml_global_data;
```
ctk21: I'm a bit confused why we need to do the juggling with `global_data` and `caml_modify_generational_global_root`.
Surely we can alter the field of any OCaml object without worrying if there is a reference registered as a generational global root from another place?

That is, can't we reduce this block to:
```C
caml_modify(&Field(caml_global_data, *pc), accu);
```

PR #519, opened 2021-03-26
Make timing test more robust

sadiqj:
d1 can enter a STW pause before it gets in to the critical section and can actually miss the notification from d2. This is highly unlikely on the current 4.12+domains+effects branch but because much more likely with DLABs and caused intermittent failures of this test.


ctk21:
LGTM

PR #521, opened 2021-03-30
-O2 -> -O0 in configure causes testsuite failures

sadiqj:


ctk21:
We actually don't need to hack the script to get the behaviour we want.

```sh
./configure CFLAGS='-O0' && make -j
```
will build a compiler with `-O0` at the end of the arguments which will apply over the `-O2` default (at least for gcc and clang as seen on [the compiler explorer](https://godbolt.org/))

PR #523, opened 2021-03-31
Systhreads Mutex raises `Sys_error`

Sudha247:
This patch attempts to bring Systhreads Mutex error checks in line with https://github.com/ocaml/ocaml/pull/9846.

For the cases mentioned in the PR, currently on multicore they are reported as fatal errors. On stock OCaml 4.12.0, the same errors are reported as `Sys_error`. With this patch, multicore compiler also reports a `Sys_error` when those errors are encountered. This is corroborated by the test in https://github.com/ocaml/ocaml/pull/9846, also added in this patch.


abbysmal:
This PR looks good to me, thanks!

There's still remnant of platform stuff, I think we may want to ideally move everything to `pthreads` functions instead of using the platform wrappers, but I think this can come as a follow up PR, this one addresses a specific issue.

Going back to last meeting's discussions, I think it may be fine to move everything to directly using the pthread's API in systhreads, however maybe we could reconsider either the design of the platform API (making it more user friendly, returning errors in an usable way.), or explore an user level API (in both C, and stdlib).
The former could be useful for external C libraries (and otherlibs stuff), and the later is basically #390.

Any opinion @ctk21 @kayceesrk ?

ctk21 (commented):
I think the code fixes the user visible problem.

Please can we not add `tests/lib-systhreads/mutex_error.ml` but actually enable the existing `tests/lib-threads/mutex_errors.ml` from upstream in `tests/disabled`? (or is there something subtly different about those tests?)

I'm not wild about us mixing raw `pthread_mutex_*` functions with `caml_plat_*` functions. Although this could be done as a follow up.

I think that we will end up with two sets of functions for synchronisation: _runtime internal_ (e.g. `caml_plat_*` series) which may be more optimized and _user facing_ (e.g. `Condvar`, `Mutex`, etc.) which may have more checks.

Sudha247:
>Please can we not add tests/lib-systhreads/mutex_error.ml but actually enable the existing tests/lib-threads/mutex_errors.ml from upstream in tests/disabled? (or is there something subtly different about those tests?)

Oops, sorry they are the same test. My grep failed, didn't notice the disabled test. It's enabled now, thanks!

ctk21:
Merging this one. We can come back to any work needed for `caml_plat_*` mixing with `pthread_*` when we get `Mutex` and `Condition` working independent of systhreads.

PR #527, opened 2021-03-31
Port eventlog to CTF

abbysmal:
Still some work to do on this one.

Some bits to refactor:
- I think the core logic is fine, but I may need to stress test it.
- I need to make sure we get similar logs between the previous (`parallel_minor_gc`) eventlog and this one.
- I need to push changes to https://github.com/ocaml-multicore/eventlog-tools

I think I may be missing some events that existed in our previous eventlog implementation, I need to check that.

### DIFF
--- a/runtime/caml/domain_state.tbl
+++ b/runtime/caml/domain_state.tbl
@@ -129,23 +129,6 @@ DOMAIN_STATE(intnat, stat_major_collections)
 DOMAIN_STATE(intnat, stat_forced_major_collections)
 DOMAIN_STATE(uintnat, stat_blocks_marked)
 
-/*****************************************************************************/
-/* Detailed stats */
-/*****************************************************************************/
-DOMAIN_STATE(uintnat, allocations)
-DOMAIN_STATE(uintnat, mutable_loads)
-DOMAIN_STATE(uintnat, immutable_loads)
-DOMAIN_STATE(uintnat, mutable_stores)
-DOMAIN_STATE(uintnat, immutable_stores)
-DOMAIN_STATE(uintnat, extcall_noalloc)
-DOMAIN_STATE(uintnat, extcall_alloc)
-DOMAIN_STATE(uintnat, extcall_alloc_stackargs)
-DOMAIN_STATE(uintnat, tailcall_imm)
-DOMAIN_STATE(uintnat, tailcall_ind)
-DOMAIN_STATE(uintnat, call_imm)
-DOMAIN_STATE(uintnat, call_ind)
-DOMAIN_STATE(uintnat, stackoverflow_checks)
-
 DOMAIN_STATE(uintnat, eventlog_paused)
 DOMAIN_STATE(uintnat, eventlog_enabled)
 DOMAIN_STATE(FILE*, eventlog_out)
--- a/runtime/caml/domain_state.tbl
+++ b/runtime/caml/domain_state.tbl
@@ -129,11 +129,27 @@ DOMAIN_STATE(intnat, stat_major_collections)
 DOMAIN_STATE(intnat, stat_forced_major_collections)
 DOMAIN_STATE(uintnat, stat_blocks_marked)
 
-DOMAIN_STATE(uintnat, eventlog_startup_timestamp)
-DOMAIN_STATE(long, eventlog_startup_pid)
+/*****************************************************************************/
+/* Detailed stats */
+/*****************************************************************************/
+DOMAIN_STATE(uintnat, allocations)
+DOMAIN_STATE(uintnat, mutable_loads)
+DOMAIN_STATE(uintnat, immutable_loads)
+DOMAIN_STATE(uintnat, mutable_stores)
+DOMAIN_STATE(uintnat, immutable_stores)
+DOMAIN_STATE(uintnat, extcall_noalloc)
+DOMAIN_STATE(uintnat, extcall_alloc)
+DOMAIN_STATE(uintnat, extcall_alloc_stackargs)
+DOMAIN_STATE(uintnat, tailcall_imm)
+DOMAIN_STATE(uintnat, tailcall_ind)
+DOMAIN_STATE(uintnat, call_imm)
+DOMAIN_STATE(uintnat, call_ind)
+DOMAIN_STATE(uintnat, stackoverflow_checks)
+
 DOMAIN_STATE(uintnat, eventlog_paused)
 DOMAIN_STATE(uintnat, eventlog_enabled)
 DOMAIN_STATE(FILE*, eventlog_out)
+DOMAIN_STATE(struct event_buffer *, eventlog_buffer)
 /* See eventlog.c */
 
 // whether or not a domain is inside of a stop-the-world handler
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -305,11 +305,6 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
     domain_state->trap_sp_off = 1;
 #endif
 
-    domain_state->eventlog_enabled = 0;
-    domain_state->eventlog_paused = 0;
-    domain_state->eventlog_startup_pid = 0;
-    domain_state->eventlog_startup_timestamp = 0;
-    domain_state->eventlog_out = NULL;
 
 #if defined(NAKED_POINTERS_CHECKER) && !defined(_WIN32)
     domain_state->checking_pointer_pc = NULL;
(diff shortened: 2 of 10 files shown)


runtime/eventlog.c
```diff
@@ -69,13 +69,37 @@ struct event {
...
+  uintnat ev_flushed;
+  int domain_unique_id;
+  atomic_uintnat ev_generated;
+  FILE *out;
   struct event events[EVENT_BUF_SIZE];
 };
 
-static struct event_buffer* evbuf;
+static __thread struct event_buffer* evbuf;
+static __thread int is_backup_thread = 0;
```
sadiqj: I'm a bit concerned about having thread locals hanging around in random places. Is the reason these can't be on Caml_state because the backup thread requires it's own buffer?

runtime/eventlog.c
```diff
@@ -69,13 +69,37 @@ struct event {
   uint64_t count; /* for misc counters */
 };
 
-#define EVENT_BUF_SIZE 4096
+/* events are stored thread-locally in event_buffers,
+   which are registered in a global doubly-linked list,
```
sadiqj: If this is only for the purposes of flushing on exit, wouldn't Caml_state be a better place to put them? The cost of iterating Max_domains is relatively small if we're only doing it once at exit.

runtime/domain.c
```diff
@@ -1272,6 +1269,7 @@ static void domain_terminate()
     handover_ephemerons(domain_state);
     handover_finalisers(domain_state);
 
+    CAML_EVENTLOG_TEARDOWN();
```
ctk21: In this loop, we may still take part in STW sections.

I think we should put the `CAML_EVENTLOG_TEARDOWN` after the domain is no longer part of STW sections; say after `caml_teardown_major_gc`

ctk21:
Minor nit, but can we clean up the `instrumented_runtime` changes to `configure.ac`?

Right now vs mainline OCaml we have the following confusing diff in `configure.ac`:
```diff
@@ -61,7 +61,7 @@ toolchain="cc"
 profinfo=false
 profinfo_width=0
 extralibs=
-instrumented_runtime=true
+instrumented_runtime=false
 instrumented_runtime_ldlibs=""

 # Information about the package
@@ -227,12 +227,11 @@ AC_ARG_ENABLE([dependency-generation],
 AC_ARG_VAR([DLLIBS],
   [which libraries to use (in addition to -ldl) to load dynamic libs])

-# Disable instrumented-runtime with multicore
 AC_ARG_ENABLE([instrumented-runtime],
   [AS_HELP_STRING([--enable-instrumented-runtime],
     [build the instrumented runtime @<:@default=auto@:>@])],
   [],
-  [enable_instrumented_runtime=no])
+  [enable_instrumented_runtime=auto])

 AC_ARG_ENABLE([vmthreads], [],
   [AC_MSG_ERROR([The vmthreads library is no longer available. \
```

I'm presuming we can revert the changes in the block with the `# Disable instrumented-runtime with multicore`?

Also it might be nice to comment the line where we have enabled the `instrumented_runtime` in multicore, saying that we have the instrumented runtime always enabled by default.

ctk21:
Another thing from a user perspective, do we have instructions on how to switch on and generate a trace using CTF which can be viewed by a user? 
(presumably switch on is still `OCAMLRUNPARAM=e` as that hasn't been changed in `startup_aux.c`?)
It's possible the instructions belong in the Wiki and not the PR, but we should try to get that lined up at the point of merge for the PR.

runtime/caml/domain_state.tbl
```diff
@@ -129,11 +129,27 @@ DOMAIN_STATE(intnat, stat_major_collections)
 DOMAIN_STATE(intnat, stat_forced_major_collections)
 DOMAIN_STATE(uintnat, stat_blocks_marked)
 
-DOMAIN_STATE(uintnat, eventlog_startup_timestamp)
-DOMAIN_STATE(long, eventlog_startup_pid)
+/*****************************************************************************/
+/* Detailed stats */
```
sadiqj: How have these reappeared? They were removed in https://github.com/ocaml-multicore/ocaml-multicore/pull/535

runtime/eventlog.c
```diff
@@ -67,15 +66,28 @@ struct event {
...
+  int domain_unique_id;
   uintnat ev_generated;
+  uint64_t alloc_buckets [20]; /* allocation stats, by size */
   struct event events[EVENT_BUF_SIZE];
 };
 
-static struct event_buffer* evbuf;
+#define evbuf Caml_state->eventlog_buffer
+
+static int64_t startup_timestamp;
```
sadiqj: Is this better as `time_t`?

runtime/eventlog.c
```diff
@@ -243,22 +295,15 @@ void caml_eventlog_init()
   char_os *toggle = caml_secure_getenv(T("OCAML_EVENTLOG_ENABLED"));
 
   if (toggle != NULL) {
-    Caml_state->eventlog_enabled = 1;
-    if (*toggle == 'p')
-      Caml_state->eventlog_paused = 1;
+    eventlog_enabled = 1;
+    // FIXME(engil): disabling eventlog pause on Multicore for now
```
sadiqj: This probably needs to be fixed?

runtime/eventlog.c
```diff
@@ -305,49 +350,55 @@ void caml_ev_counter(ev_gc_counter counter, uint64_t val)
...
   } else {
-    ++alloc_buckets[19];
+    ++evbuf->alloc_buckets[19];
   }
 }
 
 /*  Note that this function does not trigger an actual disk flush, it just
     pushes events in the event buffer.
 */
+// FIXME(engil): alloc events are not flushed on Multicore for now.
```
sadiqj: Is this still an issue?

runtime/eventlog.c
```diff
@@ -358,26 +409,29 @@ void caml_ev_flush()
 
 void caml_eventlog_disable()
 {
-  Caml_state->eventlog_enabled = 0;
-  teardown_eventlog();
+  // FIXME(engil): disabling runtime eventlog disable on Multicore for now
```
sadiqj: Likewise, this needs to be fixed.

runtime/eventlog.c
```diff
@@ -358,26 +409,29 @@ void caml_ev_flush()
...
+  // eventlog_enabled = 0;
+  // teardown_eventlog();
 }
 
 CAMLprim value caml_eventlog_resume(value v)
 {
-  CAMLassert(v == Val_unit);
-  if (Caml_state->eventlog_enabled)
-    Caml_state->eventlog_paused = 0;
+  // FIXME(engil): disabling eventlog pause on Multicore for now
```
sadiqj: Ditto

runtime/eventlog.c
```diff
@@ -358,26 +409,29 @@ void caml_ev_flush()
...
 
 CAMLprim value caml_eventlog_pause(value v)
 {
-  CAMLassert(v == Val_unit);
-  if (Caml_state->eventlog_enabled) {
-    Caml_state->eventlog_paused = 1;
-    if (evbuf && Caml_state->eventlog_out)
-      flush_events(Caml_state->eventlog_out, evbuf);
-  };
+  // FIXME(engil): disabling eventlog pause on Multicore for now
```
sadiqj: And this one.

sadiqj (changes requested):

sadiqj:
I think we need to address these `FIXME`s before merging this. I'm happy to see we've eliminated all but one of the thread locals (and the remaining one makes sense).

abbysmal:
Thanks for your review. I addressed the outstanding issues.

I implemented pausing and resume in a way that looks alright.
I introduced the `eventlog_paused` global variable so any domain requesting a pause effectively interrupt the eventlog machinery.
We could also introduce it in a domain local way but I think the former is a more common usecase than the later.
`eventlog_teardown` will take care of flushing remaining event on domain termination even if paused, so there's no need for the explicit flush from trunk.
Sorry for the detailed stats, this was a rebase mishap.

There is one FIXME that I did turn into a TODO, and it's related to the `alloc` counter events.
It is a metric that was found in trunk, and not on our implementation of eventlog. Considering how different the allocators are, I'm not sure how and where we want them.
I think it is fine to be left as a TODO item, though I did made sure the mechanism is now working and flushed correctly if we ever put them in use again.

sadiqj:
This addresses me earlier concerns. Thanks @Engil

sadiqj (approved):

Issue #524, opened 2021-03-31
Improve CI runs


abbysmal:
Miscellaneous run settings to add to the CI matrix:
- Run multiple time specific parts of the testsuite (`parallel` and such)
- tiny minor heap runs
- `-O0` runs

sadiqj:
Restricting the benchmarks to a single core is another good way of shaking out parallel issues.

abbysmal:
Another idea:

We may want to introduce a wider parameter to control `num_domains` in some tests.
This would also to use as many core as available and to control a bit better the kind of codepaths we want to explore.
In some cases we want to keep a fixed number of domains to demonstrate certain specific properties, however for many things we may as well allow to spawn as many domains as possible.

ctk21:
Closing as #528 is merged

PR #528, opened 2021-03-31
Expand CI runs

abbysmal:
This PR is aiming to improve our CI runs to make it easier for us to spot bugs in the wild.
This is a follow up to #524 

Currently implemented:
- [x] Running "parallel" "callback" "gc-roots" "effects" "lib-threads" "lib-systhreads" "weak-ephe-final" in a loop, 3 times for now.
- [x] Running the testsuite with `taskset -c 0` (following @sadiqj suggestion, I assume his finding was the `domain_id` test failing in such context.)
- [x] Running the testsuite with a small minor heap (4096 words), debug runtime.
- [x] `-O0` runs 

What needs to be done:
- [ ] maybe add more variants?
- [ ] Maybe reduce the testsuite run to the select few directories, for both the `taskset` and the small minor heap runs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,20 +3,10 @@ name: main
 on: [push, pull_request]
 
 jobs:
-  build:
-    name: ${{ matrix.name }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        include:
-          - name: linux-debug
-            os: ubuntu-latest
-            env: OCAMLRUNPARAM=v=0,V=1 USE_RUNTIME=d
-          - name: linux
-            os: ubuntu-latest
-          - name: macos
-            os: macos-latest
 
+  build:
+    name: 'linux'
+    runs-on: 'ubuntu-latest'
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


abbysmal:
We also need to add a run, compiling the runtime with `-O0` instead of the default `-O2`.
see #521

abbysmal:
@ctk21 I just pushed a change to this branch.
This change introduce a new way to handle the build and testing process:

There is now a `build` job that will take care of building the compiler, and upload the compiler as an artifact to be used by subsequent, parallelized jobs. (in the `testsuite` job matrix.)

There is also another `build-misc` job that will keep active other, separate compiler build instances, namely for `macos`, and two other linux build jobs to test building the compiler with the debug runtime on (`linux-debug`), and another to build the compiler using `CLFLAGS= -O0`.
Both `build-misc` jobs will also run the testsuite, in a non-specialized way.

The `testsuite` job will run different testsuite run profiles in parallel, based on the result of the `linux` build.
- `normal`: just a simple testsuite run. (equivalent to the previous `linux` build job.
- `super`: a loop of a selection of our favorite tests, running without extra parameters. (3 times.)
- `taskset`: runs our favorite tests, on only one core.
- `debug-s4096`: runs our favorite tests, with debug runtime and a small minor heap setting.

This is a first attempt at this approach, the numbers are looking better, though the small minor heap and single core runs are the ones taking the most time still right now.
I think we can bring them down to under the total running time of the `macos` run (which is historically the longest running one) if we are a bit more restrictive on our testsuite selection.

(and thank you @smorimoto for giving me some advice and resources to do this!)

abbysmal:
I think this is the definitive version of this PR. (in terms of features.)

It wraps up everything we discussed, and I addressed a few issues with the previous versions. (for instance, `make one DIR` does not report error in exit code, I went on to use `LIST` instead.)
(though ocaml/ocaml#10150 address this specific issue with `DIR`, I just felt that `LIST` was better suited for this job anyway.)
Now the testsuite is failing on at least `linux-O0`, on the `register_c_threads` failure (which is a known one on `O0`.).

I will take some time to investigate this failure and fix it in the coming days.

ctk21:
Looks like all the testsuite fails that were blocking this are fixed. 
Merging to get the extra coverage. Forwards!

PR #533, opened 2021-04-12
Systhreads synchronizations use pthread functions

Sudha247:
Extending #523, this patch aims to get systhreads error checking in-line with trunk. Currently we use `caml_plat_*` functions in systhreads for synchronizations. They have slightly different error codes than `pthread_*` functions.

In this patch, they are made to use `pthread_*` functions directly and raise `Sys_error` when something goes wrong, in place of `Fatal error` that's raised now.

### DIFF
--- a/otherlibs/systhreads/st_posix.h
+++ b/otherlibs/systhreads/st_posix.h
@@ -121,16 +121,16 @@ Caml_inline void st_thread_set_id(intnat id)
    threads. */
 
 typedef struct {
-  caml_plat_mutex lock;           /* to protect contents */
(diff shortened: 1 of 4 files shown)


ctk21 (commented):
I took a look, I believe the changes to do what is intended. 

Please can we remove `caml_plat_cond_init_no_mutex` and `caml_plat_cond_set_mutex` from `platform.c`/`platform.h`; these were introduced for `st_posix` but will be no longer needed after this change.

ctk21:
LGTM, merging!

PR #534, opened 2021-04-12
Move allocation size information to frame descriptors

sadiqj:
This PR implements the changes in https://github.com/ocaml/ocaml/pull/8805 

The allocation size information is now propagated using the frame descriptors, which enables it to be tracked by statmemprof (when that finally works on multicore).

To simplify the changes, I've also unified the runtime entry so polls now go through the same mechanism but with no allocations in the debug information. This is the same thing we do on https://github.com/ocaml/ocaml/pull/10039 and so should make that eventual merge back into multicore easier as well.

### DIFF
--- a/asmcomp/amd64/emit.mlp
+++ b/asmcomp/amd64/emit.mlp
@@ -277,27 +277,16 @@ let record_frame live dbg =
 (* Record calls to the GC -- we've moved them out of the way *)
 
 type gc_call =
-  { gc_size: int;                       (* Allocation size, in bytes *)
-    gc_lbl: label;                      (* Entry label *)
+  { gc_lbl: label;                      (* Entry label *)
     gc_return_lbl: label;               (* Where to branch after GC *)
     gc_frame: label;                    (* Label of frame descriptor *)
-    is_poll : bool
   }
 
 let call_gc_sites = ref ([] : gc_call list)
 
 let emit_call_gc gc =
   def_label gc.gc_lbl;
-  if gc.is_poll then begin
-    emit_call "caml_call_poll"
-  end else begin
-    match gc.gc_size with
-    | 16 -> emit_call "caml_call_gc1"
-    | 24 -> emit_call "caml_call_gc2"
-    | 32 -> emit_call "caml_call_gc3"
-    | n ->  I.add (int n) r15;
-            emit_call "caml_call_gc"
-  end;
+  emit_call "caml_call_gc";
   def_label gc.gc_frame;
   I.jmp (label gc.gc_return_lbl)
 
@@ -678,8 +667,6 @@ let emit_instr fallthrough i =
       end
   | Lop(Ialloc { bytes = n; dbginfo }) ->
       if !fastcode_flag then begin
-        let lbl_redo = new_label() in
-        def_label lbl_redo;
         I.sub (int n) r15;
         if Config.stats then begin
           I.inc (domain_field Domainstate.Domain_allocations);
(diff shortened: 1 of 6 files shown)


asmcomp/amd64/emit.mlp
```diff
@@ -792,19 +779,17 @@ let emit_instr fallthrough i =
   | Lop (Inop) -> I.nop ()
   | Lop (Ipoll) ->
       I.cmp (domain_field Domainstate.Domain_young_limit) r15;
-      let gc_call_label = new_label () in
-      let label_after_gc = new_label () in
+      let lbl_call_gc = new_label() in
       let lbl_frame =
-        record_frame_label i.live (Dbg_other i.dbg)
+        record_frame_label i.live (Dbg_alloc [])
```
ctk21, line 784: Is this a desired change in behaviour for the debug info argument: `Dbg_other i.dbg` -> `Dbg_alloc []`?
I'm not an expert on if that should be `i.dbg` or `[]`. I notice that for `Ialloc` we are getting a `dbginfo` structure.

asmcomp/emitaux.mli
```diff
@@ -39,6 +39,7 @@ val emit_debug_info_gen :
   (file_num:int -> line:int -> col:int -> unit) -> unit
 
 type frame_debuginfo =
+| Dbg_alloc of Debuginfo.alloc_dbginfo
```
ctk21: strange indentation

runtime/signals_nat.c
```diff
@@ -57,8 +59,52 @@ extern signal_handler caml_win32_signal(int sig, signal_handler action);
...
+
+  { /* Compute the total allocation size at this point,
+       including allocations combined by Comballoc */
+    unsigned char* alloc_len = (unsigned char*)(&d->live_ofs[d->num_live]);
+    int i, nallocs = *alloc_len++;
+    for (i = 0; i < nallocs; i++) {
+      /* Since 2 words is the smallest allocation, sizes are
+         encoded as (wosize - 2).
+         See Emitaux.emit_frames and caml/stack.h */
+      allocsz += alloc_len[i] + 2;
```
ctk21: I know it's the same in the end, but I wonder if it is better to take mainline's:
`Whsize_wosize(Wosize_encoded_alloc_len(alloc_len[i]))`

runtime/signals_nat.c
```diff
@@ -57,8 +59,52 @@ extern signal_handler caml_win32_signal(int sig, signal_handler action);
...
+  }
+
+  whsize = Whsize_wosize(allocsz);
+
+  Caml_state->young_ptr += whsize * sizeof(value);
+
   caml_handle_gc_interrupt();
 
+  /* Re-do the allocation: we now have enough space in the minor heap. */
+  Caml_state->young_ptr -= whsize * sizeof(value);
```
ctk21: Are we able to assert that we really have enough space in the minor heap?

The case I'm worried about is if somehow we end up in a position after `caml_handle_gc_interrupt` where:
- a minor collection happens
- major slice work schedules a finaliser
- we run the finaliser which fills up the minor heap so there isn't enough space but doesn't trigger a collection

runtime/signals_nat.c
```diff
@@ -57,8 +59,52 @@ extern signal_handler caml_win32_signal(int sig, signal_handler action);
...
+
 #ifdef WITH_SPACETIME
```
ctk21: Eak, maybe kill this `WITH_SPACETIME` block while you're in there?

ctk21 (commented):
I think this PR works, but had two concerns. 

I had a concern (which might be unfounded) about the `IPoll` Debuginfo structure.
I also had a concern around the lack of retry after `caml_handle_gc_interrupt` in `caml_garbage_collection`.

asmcomp/amd64/emit.mlp
sadiqj: We're also doing the same thing on the upstream safepoints PR: https://github.com/ocaml/ocaml/pull/10039/files#diff-bcdf301a6eb0f21dfd035884dd0ff535c11995cbca69ef9319a5ffa77114c260R660 .

The label `lbl_frame` is used for the jump back to the allocation site: https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/asmcomp/amd64/emit.mlp#L301 so I don't think it really has any corresponding debug information - it's in the epilogue. There could be an argument that it should really have debug info that locates it at the allocation site.

I think we should leave it here and then see if it needs to be fixed in the upstream PR, which we'll pull in ourselves after it's merged.

ctk21: sounds good.

ctk21:
I think this one is good to merge!

PR #535, opened 2021-04-12
remove multicore stats collection

sadiqj:
This PR removes the configurable stats collection functionality from multicore.

Removing because it shrinks the diffs to trunk in several places and it's unlikely we'll want to keep it when upstreaming.


ctk21:
I make this all of the multicore stats collection removed, merging!

PR #537, opened 2021-04-12
Port stedolan's "Micro-optimise allocations on amd64 to save a register."

sadiqj:
This PR depends on https://github.com/ocaml-multicore/ocaml-multicore/pull/534 and ports @stedolan 's PR https://github.com/ocaml/ocaml/pull/9280 to multicore.

This frees up a register and also brings down our diff on amd64's emit.mlp


ctk21:
LGTM, but we need to merge #534 first.

PR #539, opened 2021-04-16
Forcing_tag invalid argument to `Gc.finalise`

Sudha247:
Lazy values are not valid arguments to the function `Gc.finalise`. The tags `Lazy_tag` and `Forward_tag` are covered in the finaliser function `generic_final_register`. This patch adds `Forcing_tag` which is also a tag lazy values could have when the computation is being forced, so that `Gc.finalise` raises an invalid argument exception when a block with `Forcing_tag` is given as an argument.


ctk21:
LGTM, merging!

PR #542, opened 2021-04-20
Add ephemeron lazy test

Sudha247:
This test covers ephemerons, lazy values and domain life-cycle interspersed with GC.


ctk21:
It looks like the lazy is acting as a guard on `set_values` which will create ephemerons in one domain then we terminate the domains, do a GC and then check the ephemerons constructed in the lazy block. I'm all for tests that get us coverage of subtle code paths we currently don't execute in the testsuite, but I'm a bit unclear what this one is covering that we don't aren't covering elsewhere. 

If we take the features, lazy, ephemerons and domain lifecycle, I can see that there might be something involving ephemerons that have lazy objects in the key or value field. Maybe there is something that could be fiddly with the various lazy tags and how the set/get/update ephemeron code works. With the lifecycle maybe there is something that could go wrong with transferring of an ephemeron on domain termination with some of the lazy tags in the ephemeron key or value. 

I'm not sure if there is something specific you are trying to catch with this one?

Sudha247:
Having lazy objects as key or data in ephemeron sounds good, let me try that out. With this test I was just trying to run ephemeron operations with `Lazy.force`, if it's not testing any new code path then maybe we can drop it in favour of other tests.

Sudha247:
Code in this test is covered elsewhere in the testsuite. #550 adds more coverage for Lazy. Closing this PR.

PR #543, opened 2021-04-20
Parallel version of weaklifetime test

Sudha247:
Adding a parallel version of the `weaklifetime.ml` test, wherein Weak structures are being accessed by multiple domains.


testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -0,0 +1,91 @@
...
+    (* we only read data.(i).objs.(j) in this local binding to ensure
+        that it does not remain reachable on the bytecode stack
+        in the rest of the function below, when we overwrite the value
+        and try to observe its collection.  *)
+    match data.(i).objs.(j), Weak.check data.(i).wp j with
+    | Present x, false -> assert false
+    | Absent n, true -> assert (gc1 <= n+2); No_change
+    | Absent _, false -> Fill
+    | Present _, true ->
+      if Random.int 10 = 0 then Erase else No_change
```
ctk21: Wouldn't this usage of `Random.int` need to use the DLS random state?
I wonder if just wrapping up the DLS boilerplate to give you a `random_int` function all the way through is the cleanest way.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -0,0 +1,91 @@
...
+    let gc2 = gccount () in
+    if gc1 <> gc2 then data.(i).objs.(j) <- Absent gc2;
+;;
+
+let dummy = ref [||];;
+
+let run i () =
+  let s = Domain.DLS.get random_state in
+  while gccount () < 5 do
+    dummy := Array.make (Random.int 300) 0;
```
ctk21: Similar comment about using `Random.int` from multiple domains, we probably want to introduce and use a `random_int` function to avoid all the boilerplate.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -0,0 +1,91 @@
...
+    | Present x, false -> assert false
+    | Absent n, true -> assert (gc1 <= n+2); No_change
+    | Absent _, false -> Fill
+    | Present _, true ->
+      if Random.int 10 = 0 then Erase else No_change
+  in
+  match change with
+  | No_change -> ()
+  | Fill ->
+    let x = Array.make (1 + Random.int 10) 42 in
```
ctk21: Isn't this a usage of `Random.int` from multiple domains?

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -0,0 +1,91 @@
...
+  objs : objdata array;
+  wp : block Weak.t;
+};;
+
+let data =
+  Array.init size (fun i ->
+    let n = 1 + Random.int size in
+    {
+      objs = Array.make n (Absent 0);
+      wp = Weak.create n;
```
ctk21, line 27: Right now this will create all the weak pointers in the initial main domain and so will not undergo a domain lifecycle event. 
Are we able to change the initialization of the data to be done per domain?
Then each domain would have created weak pointers and experienced lifecycle events with that.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -0,0 +1,91 @@
...
+    let i = (Random.State.int s (size/num_domains)) + i * size/num_domains in
+    let j = Random.State.int s (Array.length data.(i).objs) in
+    check_and_change i j;
+  done
+
+let _ =
+  for i = 1 to 5 do
+    let domains = Array.init (num_domains - 1) (fun i -> Domain.spawn(run i)) in
+    run (num_domains - 1) ();
+    Array.iter Domain.join domains
```
ctk21, line 98: I wonder if it is easy to make the test even more challenging. Right now each array element is where things are updated. At this point where only the main domain is running, we shuffle (or rotate) the underlying `data.(i)` index that is used for the next iteration. 
This would then give us an example where multiple domains are calling operations on a weak pointer.

ctk21 (commented):
This is useful because it will introduce weak pointers in multiple domains and we want to get some domain lifecycle events with active weak pointers. 

I had a couple of comments:
 - we need to guard `Random.int` usage in the presence of multiple domains; introducing a `random_int` wrapper for the boilerplate might be the simplest solution in this context.
 - we are creating all the weak pointers in the initial main domain and so these weak pointers will not experience the termination of their owning domain. I think we should fix this so that we get coverage of the orphaning process. 
 - the access pattern for the test is quite partitioned. I wonder if just adding an iteration count index offset would allow for multiple domains operating on the same weak pointer.

Sudha247:
Thanks @ctk21! Adding a `random_int` guard with DLS is a good idea. We need weak pointers created at new domains to cover orphaning, and I think retaining the current one created at main domain could be useful to have same weak array accessed from multiple domains. Will add those.

>the access pattern for the test is quite partitioned. I wonder if just adding an iteration count index offset would allow for multiple domains operating on the same weak pointer.

Access pattern partitioning was intentional so as to not have two domains trying to update the same field. Is `Weak.set` thread-safe to be called from multiple domains on the same field?

ctk21:
> Access pattern partitioning was intentional so as to not have two domains trying to update the same field. Is `Weak.set` thread-safe to be called from multiple domains on the same field?

I'm not suggesting the test be made thread-safe. I'm thinking that when all the domains come to a halt after the `Domain.join` you switch around the mapping for which domain operates on a given slice in the next iteration. This means that a domains will (serially) touch weak pointers owned by other domains and operated on previously by other domains.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -0,0 +1,102 @@
+(* TEST
+*)
+
+Random.init 12345;;
```
abbysmal: I think this is not required anymore since `random_int` use DLS?

abbysmal:
>      I'm not suggesting the test be made thread-safe. I'm thinking that when all the domains come to a halt after the Domain.join > you switch around the mapping for which domain operates on a given slice in the next iteration. This means that a domains will > (serially) touch weak pointers owned by other domains and operated on previously by other domains.

But since domains are effectively re-spawned at every iterations, isn't it effectively the same?
Each iterations re-spawns domain, so domain at index 0 in the next run is effectively a different domain. (but there's slot reuse involved in the same order.)

I just reviewed this PR (thanks for this testcase!), and it looks good to me. (minor a leftover `Random.init` at the top of the file.).
I'm not sure I understood the index shuffling suggestion (and I may be missing subtler implications), if this is something we want to do I can wait before merging, otherwise I think this test can go in.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
ctk21: With the introduction of `index` we have weak pointer operations being done on the same weak pointer by multiple domains. I'm marking this resolved.

Sudha247:
Thanks for taking a look @Engil! Removed the leftover `Random.init`. I think @ctk21 also meant about slot resuse wrt index shuffling, anyway that's been incorporated to the test now.

abbysmal:
Right, I checked the overall changes and did not see this one came in a commit later.
Thanks, merging!

PR #550, opened 2021-05-03
Lazy effects test

Sudha247:
This patch adds a test that covers effects with Lazy computations. The different cases are these:

1. A lazy computation is suspended and later continued in the same domain. It is expected to finish the computation and return the result.
2. Another lazy computation is started then suspended. It is continued in a new domain. As long as no other domain tries to force the same lazy value, the computation will end and return the result.
3. This time the suspended computation is forced again from a new domain. It is expected to raise a `RacyLazy` exception because a different domain is trying to force the same lazy value.


testsuite/tests/effects/test_lazy.ml
```diff
@@ -0,0 +1,34 @@
+(* TEST *)
+
+effect Stop : unit
+
+let f count =
+  let _n = (Domain.self ():> int) in
```
ctk21: Do we need this, it appears unused?

testsuite/tests/effects/test_lazy.ml
```diff
@@ -0,0 +1,34 @@
...
+  done;
+  !r
+
+let _ =
+  let l = lazy (f 1_000) in
+  let v1 =
+    try Lazy.force l with
+    effect Stop k -> continue k () in
+    Printf.printf "%d\n" v1;
+  let l2 = lazy (f 1_000) in
```
ctk21: Might be nice to make this a different number, e.g. `2_000`, so that things in the output tie unambiguously back to the test code.

testsuite/tests/effects/test_lazy.ml
```diff
@@ -0,0 +1,34 @@
...
+  let v2 =
+    try Lazy.force l2 with
+    effect Stop k ->
+      let d = Domain.spawn(fun () -> continue k ()) in
+      Domain.join d in
+    Printf.printf "%d\n" v2;
+  let l3 = lazy (f 1_000) in
```
ctk21: Same again, maybe `3_000`.

ctk21 (commented):
This test looks reasonable to me, aside some minor nits.

ctk21:
LGTM, merging

PR #552, opened 2021-05-18
Add a force_instrumented_runtime option to configure

abbysmal:
This commit aims to introduce an option to configure an OCaml installation to force the usage of the instrumented runtime on linker invocations.

This option can be enabled using `./configure --enable-force-instrumented-runtime`

This is required for ocaml-bench/sandmark#218
This allows a compiler switch to always link in the instrumented runtime in order to extract event logs in an automated fashion, even for packages installed through Opam, without having to modify any `dune`/else build files.

The approach is pretty similar to the current [trick used in sandmark for trunk](https://github.com/Sudha247/ocaml/tree/4.10.0+stock%2Binstrumented).

A few modifications are required for the OCaml distribution to compile itself with a different runtime (having to add explicit `runtime` directories here and there, as with @Sudha247's branch linked previously)

I think this approach works, however I could see an improvement by making it a more general option:
`-default-runtime=$runtime` where `$runtime` could be `i` or `d` (allowing to configure an OCaml installation to default to either the debug or instrumented runtime.)

I am unsure this is worth the extra work, but if it sounds like a good idea I can make it work that way as well.


kayceesrk:
Lgtm. 

One thing to check is whether `make install` actually copies the instrumented version of the runtime to the installation directory. If you can confirm that, I can merge this one.

abbysmal:
@kayceesrk I did gave it a trial run some times ago against my branch of sandmark, it does work as advertised (it can link successfully against it)

I just checked again if it is properly installed:
```
●●●● make install | grep runi
/usr/bin/install -c ocamlrun ocamlrund ocamlruni "/tmp/ocaml/bin"
/usr/bin/install -c -m 644 ld.conf libcamlrun.a libcamlrund.a libcamlruni.a libcamlrun_pic.a "/tmp/ocaml/lib/ocaml"
/usr/bin/install -c -m 644 libasmrun.a libasmrund.a libasmruni.a libasmrun_pic.a "/tmp/ocaml/lib/ocaml"

```

kayceesrk:
Sounds good! 👍

kayceesrk (approved):

PR #555, opened 2021-05-20
runtime: CAML_TRACE_VERSION is now set to a Multicore specific value

abbysmal:
This is required in order to work on ocaml-bench/sandmark#218
We need to distinguish a Multicore OCaml trace and a trunk one, as the format varies ever so slightly.
This difference is handled in ocaml-multicore/eventlog-tools#2
I've set it to a number we are likely not to reach, bumping upstream's trace version, as it is for our own development purposes.


ctk21:
LGTM - merging. 
(The ocaml-multicore-ci CI error is spurious here)

PR #562, opened 2021-06-04
Import fixes to the minor heap allocation code from DLABs.

abbysmal:
This PR provides two fixes to the way the minor heap is allocated on Multicore OCaml.

The first fix removes a multiplication by two to the overall size of the minor, which I assume is related to earlier experiments on our minor heap code.

The second fix make sure that we do convert the `Minor_heap_max` limit from `config.h` toa byte size where appropriate in the codebase.
Before this fix, it turns out we allocate the minor heaps segment in word size and not byte size, which leads to an oversized minor heaps segment

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -150,13 +150,11 @@ CAMLexport __thread caml_domain_state* Caml_state;
 
 asize_t caml_norm_minor_heap_size (intnat wsize)
 {
-  asize_t page_size = caml_mem_round_up_pages(1);
   asize_t bs, max;
   if (wsize < Minor_heap_min) wsize = Minor_heap_min;
   bs = caml_mem_round_up_pages(Bsize_wsize (wsize));
 
-  Assert(page_size * 2 < Minor_heap_max);
-  max = Minor_heap_max - page_size * 2;
+  max = Bsize_wsize(Minor_heap_max);
 
   if (bs > max) bs = max;
 
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -355,7 +355,7 @@ void caml_init_domains(uintnat minor_heap_wsz) {
     caml_fatal_error("Minor_heap_max misconfigured for this platform");
 
   /* reserve memory space for minor heaps and tls_areas */
-  size = (uintnat)Minor_heap_max * Max_domains;
+  size = (uintnat)Bsize_wsize(Minor_heap_max) * Max_domains;
   tls_size = caml_mem_round_up_pages(sizeof(caml_domain_state));
   tls_areas_size = tls_size * Max_domains;
 
@@ -387,12 +387,12 @@ void caml_init_domains(uintnat minor_heap_wsz) {
     dom->backup_thread_msg = BT_INIT;
 
     domain_minor_heap_base = caml_minor_heaps_base +
-      (uintnat)Minor_heap_max * (uintnat)i;
+      (uintnat)Bsize_wsize(Minor_heap_max) * (uintnat)i;
     domain_tls_base = caml_tls_areas_base + tls_size * (uintnat)i;
     dom->tls_area = domain_tls_base;
     dom->tls_area_end = domain_tls_base + tls_size;
     dom->minor_heap_area = domain_minor_heap_base;
-    dom->minor_heap_area_end = domain_minor_heap_base + Minor_heap_max;
+    dom->minor_heap_area_end = domain_minor_heap_base + Bsize_wsize(Minor_heap_max);
   }
 
 
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -617,7 +617,7 @@ struct domain* caml_domain_self()
 struct domain* caml_owner_of_young_block(value v) {
   int heap_id;
   Assert(Is_young(v));
-  heap_id = ((uintnat)v - caml_minor_heaps_base) / Minor_heap_max;
+  heap_id = ((uintnat)v - caml_minor_heaps_base) / Bsize_wsize(Minor_heap_max);
   return &all_domains[heap_id].state;
 }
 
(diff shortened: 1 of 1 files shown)


kayceesrk:
> The first fix removes a multiplication by two to the overall size of the minor, which I assume is related to earlier experiments on our minor heap code.

To clarify, this was necessary for concurrent minor GCs. See sec 4.3.2 of https://arxiv.org/abs/2004.11663. We no longer need this since Multicore now performs stop-the-world parallel minor GC.

runtime/domain.c
```diff
@@ -355,11 +355,11 @@ void caml_init_domains(uintnat minor_heap_wsz) {
     caml_fatal_error("Minor_heap_max misconfigured for this platform");
 
   /* reserve memory space for minor heaps and tls_areas */
-  size = (uintnat)Minor_heap_max * Max_domains;
+  size = (uintnat)Bsize_wsize(Minor_heap_max) * Max_domains;
```
kayceesrk, line 356: It would be better to rename `Minor_heap_max` to `Minor_heap_max_wsize`. And add a comment where `Minor_heap_max_wsize` is defined to say that this would allow for a maximum minor heap size of 2GB.

ctk21:
I think this PR is correct. 

ocaml and ocaml-multicore are in alignment on what the `Minor_heap_max` variable should be; see [ocaml](https://github.com/ocaml/ocaml/blob/trunk/runtime/caml/config.h#L208-L219) and [ocaml-multicore](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/runtime/caml/config.h#L227-L237). There might be an argument that the naming should include units. If we do decide to change the units in multicore from those used in ocaml, I feel we should change the name.

The intent of the code which this PR touches is that we will reserve `Minor_heap_max * Max_domains` words with each domain then committing its chunk. The code before the PR has:
```C
size = (uintnat)Minor_heap_max * Max_domains
...
heaps_base = caml_mem_map(size*2, size*2, 1 /* reserve_only */);
```
from what I see `caml_mem_map` (and `mmap` inside `caml_mem_map`) are expecting the size parameter to be in bytes.

I think there are two bugs in that snippet above which this PR fixes:
 - the need to convert size into bytes for `caml_mem_map`
 - the `spurious size*2` which is probably a hangover from previous experiments

I believe the PR is good in fixing the error; I don’t see an issue with the comments in config.h as there is alignment between stock and ocaml-multicore.
I make:
  `Bsize_wsize(Minor_heap_max) * Max_domains`
= `Minor_heap_max * sizeof(value) * Max_domains`
= `(1<<28) * 8 * 128`
= `274877906944` bytes (or `256GB`)
which is `2GB` per domain

ctk21:
If we do intend to change the naming of `Minor_heap_max` to include `wsize`, please can I request that we change `Minor_heap_min` and `Minor_heap_def` as well to maintain consistency between the three minor heap variables in `config.h`.

I also notice that we have `Stack_threshold_words`. So I wonder if `words` is better than `wsize`. 

Should we go for a rename, my preference is that we try and get as much consistency as makes sense across the `Minor_heap_xxx` and other variables in `config.h`.

abbysmal:
Thanks @ctk21 for your input.
I am pretty hesitant on renaming: it would have made more sense if the underlying units were inconsistent (as we suspected earlier).
However it looks like the maths do check out and the comments in `config.h` are indeed accurate.
I think renaming would be ok and improve clarity to some extent. But it would also clobber the diff with trunk a bit further.
@kayceesrk do you think we should go with renaming nonetheless?

kayceesrk:
Given that both stock and multicore are in agreement, we should retain the names as-is.

kayceesrk (approved):

abbysmal:
I took a look at what would happen if we were to let a program running 128 domains go through a few GC cycle with the minor heap segment configured to `Minor_heap_max`. (as suggested by @ctk21)
Two bugs were spotted by Tom in the process and addressed. (with a small adjustments I will comment on separately.)

runtime/domain.c
```diff
@@ -150,13 +150,11 @@ CAMLexport __thread caml_domain_state* Caml_state;
 
 asize_t caml_norm_minor_heap_size (intnat wsize)
 {
-  asize_t page_size = caml_mem_round_up_pages(1);
   asize_t bs, max;
   if (wsize < Minor_heap_min) wsize = Minor_heap_min;
   bs = caml_mem_round_up_pages(Bsize_wsize (wsize));
 
-  Assert(page_size * 2 < Minor_heap_max);
-  max = Minor_heap_max - page_size * 2;
```
abbysmal, line 159: @kayceesrk I am unsure about the requirement for `- page_size * 2` here.
Is it related to guard pages of some sort that used to be put in place? Or am I missing something?

kayceesrk: IIRC, this was when the DLS area was maintained just beside the minor heap area. Currently, we maintain the Caml_state in r14 register on amd64. Earlier, this was not maintained in a separate register. It was computed from the young_pointer, which is maintained in the r15 register, by applying a mask. Now that we maintain Caml_state in r14 and Caml_state is anyway maintained separately, we don't need this `page_size * 2`.

runtime/domain.c
```diff
@@ -355,11 +353,11 @@ void caml_init_domains(uintnat minor_heap_wsz) {
     caml_fatal_error("Minor_heap_max misconfigured for this platform");
```
ctk21, line 353: We have one place on L352 (just above this one), where we do:
 `caml_mem_round_up_pages(Minor_heap_max) != Minor_heap_max`
I believe we should be doing `Bsize_wsize` on both of the occurrences of `Minor_heap_max` here.

ctk21:
LGTM, merging

PR #573, opened 2021-06-08
Backport trunk safepoints PR to multicore

sadiqj:
This PR backports https://github.com/ocaml/ocaml/pull/10039 to multicore.


sadiqj:
There is currently one failing test:

`tests/lib-systhreads/eintr`

Where a signal seems to be delivered in the wrong place. The eintr part of the test receives Sys.Break which is quite odd.

**Updated**: Found the culprit https://github.com/ocaml-multicore/ocaml-multicore/pull/598

ctk21:
I tried this out with sandmark on a large Xen2 box. Here are the results for big instances:
20210702_sherwood_pr573_time
20210702_sherwood_pr573_speedup

The run has discovered a bug in the PR. For some reason the `allocated_words` in `OCAMLRUNPARAM=v=0x400` is getting corrupted:
```
$ OCAMLRUNPARAM=v=0x400 _build/4.12.0+domains+effects_pr573_1/benchmarks/multicore-numerical/binarytrees5_multicore.exe 16 24
allocated_words: 6917529036583372800
minor_words: 6917529036583368704
promoted_words: 3421608747
major_words: 3421612845
minor_collections: 54627
major_collections: 28
forced_major_collections: 0
heap_words: 854568928
top_heap_words: 940707808
mean_space_overhead: 7.897463

$ OCAMLRUNPARAM=v=0x400 _build/4.12.0+domains+effects_1/benchmarks/multicore-numerical/binarytrees5_multicore.exe 16 24
allocated_words: 8942290884
minor_words: 8942286786
promoted_words: 3839254833
major_words: 3839258931
minor_collections: 72463
major_collections: 34
forced_major_collections: 0
heap_words: 856752096
top_heap_words: 925552608
mean_space_overhead: 12.686927
```

Also there is an interesting thing going on with this code in the binarytrees benchmark for sandmark:
```ocaml
let rec check t =
  Domain.Sync.poll ();
  match t with
  | Empty -> 0
  | Node(l, r) -> 1 + check l + check r
```
by making `Domain.Sync.poll` a no-op, the recursive function `check` no longer gets a safepoint. I couldn't also see an easy way to refactor the code and get a safepoint poll in there. (This may be something we just leave)

kayceesrk:
```ocaml
let rec check t =
  Domain.Sync.poll ();
  match t with
  | Empty -> 0
  | Node(l, r) -> 1 + check l + check r
```

This function is non-tail-recursive. So it cannot run forever without stack overflow. It would be useful to poll in our stack realloc code (if we don't already do).

sadiqj:
Got the stats bug and fixed another asynchronous exceptions one as well: https://github.com/ocaml-multicore/ocaml-multicore/pull/573/commits/a129c9713989c512c7027c801bbe31e677507591 https://github.com/ocaml-multicore/ocaml-multicore/pull/573/commits/09234e03d7ffb7f29888c7921f07ff4f41df03ca

It was possible for `young_ptr` to be less than `young_end` which caused some crazy stats unsigned underflow.

ctk21:
I took a look at those two commits, but I couldn't see how they affects `young_ptr` and `young_end` are you able to point me in the direction of how that works?

sadiqj:
> I took a look at those two commits, but I couldn't see how they affects `young_ptr` and `young_end` are you able to point me in the direction of how that works?

https://github.com/ocaml-multicore/ocaml-multicore/commit/09234e03d7ffb7f29888c7921f07ff4f41df03ca is the crucial one.

Essentially polls continued with the rest of the GC logic. So they'd hit https://github.com/ocaml-multicore/ocaml-multicore/blob/09234e03d7ffb7f29888c7921f07ff4f41df03ca/runtime/signals_nat.c#L106 and we would assume there had been a a block with zero fields allocation (but with a header!) on to the minor heap we needed to undo.

So we'd remove 8 bytes from `young_ptr`. If we did that on an empty heap (where `young_ptr == young_end`) then we'd have `young_ptr > young_end` and subtracting the former from the latter in stats would cause some outrageous numbers.

Issue #376, opened 2020-07-24
Parallel domain unsound functions in stdlib


ctk21:
We have functions in stdlib which can be unsound when two domains use the function at the same time.

For example, `Random.bits ()` currently uses a global state that is shared by multiple domains that is [updated non-atomically](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.10.0/stdlib/random.ml#L76-L83). This makes the function potentially not safe to use with multiple domains (in this case different state sequences could be observed between a sequential program and a parallel one. Imagine one domain suspended (by OS, CPU, etc) after calculating its local `(s.idx + 1) mod 55` on L77 but before storing it to `s.idx` and another domain then executing the whole function from scratch).

One strategy for handling this issue is to sweep the stdlib documentation and flag all functions that are unsafe for multiple domains to use.

ctk21:
We also need to be careful with the usage of `lazy`. Stdlib makes use of lazy in places like `Hashtbl.create` and `Filename.temp_file` which can also be a source of unexpected behaviour between domains. The lazy forcing itself could raise `RacyLazy` which could be unexpected.

kayceesrk:
There is now a written spec around what we expect from `stdlib` in Multicore OCaml: https://github.com/ocaml-multicore/ocaml-multicore/wiki/Safety-of-Stdlib-under-Multicore-OCaml.

kayceesrk:
Fixed by #577 and #582.

PR #577, opened 2021-06-10
Use DLS for default random state

### LINKED ISSUE #564: Default state in Random should be domain-local
The default state in the `Random` module is now global. This makes it unsafe to use of `Random` module by multiple domains in parallel. See https://github.com/ocaml-multicore/ocaml-multicore/issues/376#issue-665008228. The module should be fixed such that the [default state](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/stdlib/random.ml#L150-L164) is made domain-local.

kayceesrk:
Fixes https://github.com/ocaml-multicore/ocaml-multicore/issues/564 and https://github.com/ocaml-multicore/ocaml-multicore/issues/376. 

The main challenge is to ensure that Random functions do not slow down as more domain-local state entries are added. The trick is to change the representation of domain-local state to a record where the default random state and the user-defined state are fields in a record (See `type dls_state` in `obj.ml`). This makes the access to the domain-local random state a constant time operation.

There are a few follow-up tasks when this PR is merged. Firstly, the `caml_domain_dls_get` should be implemented as a primitive rather than a non-allocating external call. This will make the cost of accessing the default Random state as close to stock. Next, the use of explicit DLS for Random state in `domainslib`, `sandmark`, the tutorial, etc, should be reverted. @Sudha247 would you volunteer to do the latter?


Sudha247:
> Next, the use of explicit DLS for Random state in domainslib, sandmark, the tutorial, etc, should be reverted.

Yeah, happy to make the necessary updates after this is merged.

kayceesrk:
Opened #582 instead of this PR.

PR #579, opened 2021-06-10
Primitive for fetching DLS root

kayceesrk:
This PR implements a primitive for fetching DLS. Now fetching the DLS root compiles down to a single `mov` instruction on `amd64`.


asmcomp/CSEgen.ml
```diff
@@ -237,11 +237,12 @@ method class_of_operation op =
   | Ispecific _ -> Op_other
   | Iname_for_debugger _ -> Op_pure
   | Ipoll -> Op_checkbound
-  | Inop -> Op_store true (* not correct, but enough to force CSE from leaving
-                             nops alone *)
+  | Inop ->
+      (* not correct, but enough to force CSE from leaving nops alone *)
+      Op_store true
+  | Idls_get -> Op_pure
```
ctk21: I'm not a deep expert, but is it correct that:
```ocaml
| Iload(_,_) -> Op_load
```
but 
```ocaml
| Idls_get -> Op_pure
```
I'm not sure I understand why `Idls_get` shouldn't be the same as a `Iload`

ctk21 (commented):
I think this does the right thing and is good. 

One query about `Op_pure` for `Idsl_get` in CSEgen.

asmcomp/CSEgen.ml
kayceesrk: I shall change this one to `Op_load`.

sadiqj (approved):
Would be nice if we had a way of translating this to some existing Cmm and Mach, to avoid having to push it through to the emitters.

At the moment there's no concept of `Caml_state` in cmm. If we find further uses of that, we could probably add `Cload_caml_state` or similar at cmm and make this a bit more general.

For now though, this works.

ctk21:
I will merge

Issue #566, opened 2021-06-07
Make `Filename.temp_file` thread-safe


kayceesrk:
`Filename.temp_file` uses [global state](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/stdlib/filename.ml#L330-L334). This state should be made domain-local. This usage is very similar to the issue with [`Hashtbl.create`](https://github.com/ocaml-multicore/ocaml-multicore/issues/565).

kayceesrk:
Fixed by #582.

Issue #565, opened 2021-06-07
Make `Hashtbl.create` thread-safe


kayceesrk:
`Hashtbl.create` uses [global state](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/stdlib/hashtbl.ml#L60-L76). This state should be made domain-local.

kayceesrk:
Fixed by #582.

PR #582, opened 2021-06-11
Make global state domain-local in Random, Hashtbl and Filename

### LINKED ISSUE #564: Default state in Random should be domain-local

kayceesrk:
Fixes issues #564, #565, #566. 

Depends on PR #579.


stdlib/domain.ml
```diff
@@ -120,24 +115,24 @@ module DLS = struct
...
-    match add_or_update_entry key cs vals with
+    match add_or_update_entry key cs st.entry_list with
     | None -> ()
-    | Some e -> set_dls_list (e::vals)
+    | Some e -> st.entry_list <- e::st.entry_list
 
   let get k =
-    let rec search key_id init dls_list l =
+    let st = get_dls_state () in
+    let rec search (key_id, init) l =
```
ctk21: Unfortunately the closure `search` is capturing `st` which is leading to an allocation on every call `get k`. Similar to #405, can we move to:
```ocaml
let rec search st (key_id, init) l =
...
in
let st = get_dls_state () in
Obj.obj @@ search st k st.entry_list 
```

stdlib/hashtbl.ml
```diff
@@ -57,7 +57,13 @@ let randomized = ref randomized_default
 let randomize () = randomized := true
 let is_randomized () = !randomized
 
-let prng = lazy (Random.State.make_self_init())
+let _ = CamlinternalDomain.(register_initialiser (fun st ->
+  st.hashtbl <- Obj.repr (lazy (Random.State.make_self_init()))))
+
+let current_state () =
```
ctk21: I wonder if `prng_state` might be a better name for this.

stdlib/random.ml
```diff
@@ -163,22 +163,30 @@ let default = {
...
+
+let full_init seed = State.full_init (current_state ()) seed
+let init seed = State.full_init (current_state ()) [| seed |]
 let self_init () = full_init (random_seed())
 
 (* Manipulating the current state. *)
 
-let get_state () = State.copy default
-let set_state s = State.assign default s
+let get_state () = State.copy (current_state ())
```
ctk21: We should update the mli documentation to highlight that `get_state` and `set_state` is for the current domain.

stdlib/random.ml
```diff
@@ -163,22 +163,30 @@ let default = {
...
+
+let bits () = State.bits (current_state ())
+let int bound = State.int (current_state ()) bound
+let int32 bound = State.int32 (current_state ()) bound
+let nativeint bound = State.nativeint (current_state ()) bound
+let int64 bound = State.int64 (current_state ()) bound
+let float scale = State.float (current_state ()) scale
+let bool () = State.bool (current_state ())
+
+let full_init seed = State.full_init (current_state ()) seed
```
ctk21: We should update the mli documentation to highlight that `full_init`, `init` and `self_init` update just the current domain's seed value.

ctk21 (commented):
I think this PR fixes the Stdlib issues which are pressing. 

Down the line we want to have a uniform efficient way for user level DLS variables vs the special fast locations for Stdlib variables introduced in this PR. We should probably file an issue for that after we get this one done. 

I had a one place where we should avoid causing an allocation in `DLS.get` and minor naming/documentation nits.

(I ignored the diff coming from #579).

kayceesrk:
I've significantly simplified the PR thanks to making the domain-local state an array allowing constant-time access.

stdlib/hashtbl.ml
```diff
@@ -72,7 +72,8 @@ let rec power_2_above x n =
 
 let create ?(random = !randomized) initial_size =
   let s = power_2_above 16 initial_size in
-  let seed = if random then Random.State.bits (Lazy.force prng) else 0 in
+  let random_state = Domain.DLS.get prng_key in
+  let seed = if random then Random.State.bits random_state else 0 in
```
gasche: it would be nicer if `DLS.get` only occurred under the `if random` test, to avoid initializing the PRNG when it is not used.

stdlib/hashtbl.ml
```diff
@@ -617,7 +618,8 @@ let of_seq i =
 let rebuild ?(random = !randomized) h =
   let s = power_2_above 16 (Array.length h.data) in
   let seed =
-    if random then Random.State.bits (Lazy.force prng)
+    let random_state = Domain.DLS.get prng_key in
+    if random then Random.State.bits random_state
```
gasche: (same here)

stdlib/random.mli
```diff
@@ -18,18 +18,18 @@
...
-(** Initialize the generator with a random seed chosen
-   in a system-dependent way.  If [/dev/urandom] is available on
-   the host machine, it is used to provide a highly random initial
-   seed.  Otherwise, a less random seed is computed from system
-   parameters (current time, process IDs). *)
+(** Initialize the domain-local generator with a random seed chosen
+    in a system-dependent way.  If [/dev/urandom] is available on the host
+    machine, it is used to provide a highly random initial seed.  Otherwise, a
+    less random seed is computed from system parameters (current time, process
+    IDs). *)
```
gasche: When /dev/urandom is not available, we have a randomness problem if many domains are spawned at the same time and initialize their PRNG right at the start. Is it possible to add the domain identifier to the mix?

stdlib/domain.ml
```diff
@@ -9,6 +9,61 @@ module Raw = struct
...
+  let maybe_grow idx =
+    let st = get_dls_state () in
+    if idx < Array.length st then st
+    else begin
+      let sz = Int.shift_left 1 (1 + log2 idx) in
+      let new_st = Array.make sz unique_value in
+      Array.blit st 0 new_st 0 (Array.length st);
+      set_dls_state new_st;
+      new_st
+    end
```
gasche, line 52: The more common way to write this dynamic-growth logic is to double the current size until we get above the target index. I think this more common version is better because (1) it's obvious that the resulting size is above the target index, and (2) in the common case (which is the case here) where the target index is right after the size limit, the new size is computed in O(1) instead of O(log n). Of course, this does not matter very much here, as this operation is infrequent.

stdlib/StdlibModules
```diff
@@ -34,11 +34,12 @@ STDLIB_MODS=\
   stdlib pervasives seq option either result bool char uchar \
   sys list bytes string unit marshal obj array float int int32 int64 nativeint \
   lexing parsing set map stack queue camlinternalLazy lazy stream buffer \
-  camlinternalFormat printf arg atomic \
+  atomic mutex condition semaphore domain \
+	camlinternalFormat printf arg \
```
kayceesrk: Tabs to spaces.

stdlib/hashtbl.ml
kayceesrk: Good catch. Fixed.

stdlib/hashtbl.ml
kayceesrk: Fixed.

stdlib/random.mli
kayceesrk: I've updated the the random state initialisation with a unique token of domain-local state.

stdlib/domain.ml
kayceesrk: Fixed.

runtime/sys.c
```diff
@@ -599,6 +599,7 @@ CAMLprim value caml_sys_random_seed (value unit)
     data[n++] = getpid();
     data[n++] = getppid();
 #endif
+    data[n++] = (intnat)Caml_state->unique_token_root;
```
gasche, line 602: The current implementation is `caml_sys_random_seed` is not very nice (I think); in the current state you also need to add this logic to `caml_win32_random_seed` in `runtime/win32.c`.

Another approach would be to add your write *outside* the `#ifdef WIN32 ... #endif` block, written as `if (n < 16) data[n++] ...`. Then it would happen on both systems, which reduces duplication.

Reading this code made we want to refactor it, so maybe I'll send a PR upstream directly -- without your change.

testsuite/tests/gc-roots/globroots.ml
```diff
@@ -45,23 +43,23 @@ module Test(G: GLOBREF) () = struct
     done
 
   let change () =
-    match Random.State.int (Domain.DLS.get random_state) 37 with
+    match Random.int 37 with
```
gasche: Does this change preserve the behavior of the test? If I read things correctly, previous all domains would start from their own self-initialized seed, while the behavior of `Random.int 37` is to start from the static seed (independently in any domain).
I think that the change is "correct" in the sense that the new version is in-line with the behavior of the test on trunk. But thinking about this code made me realize that working with Random global state in multi-domain programs is not simple.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
```diff
@@ -3,9 +3,6 @@
 
 let size = 1000;;
 let num_domains = 4;;
-let random_state = Domain.DLS.new_key Random.State.make_self_init
-
-let random_int = Random.State.int (Domain.DLS.get random_state)
```
gasche: This test does use several domains, and I'm not sure that removing the random-state initialization completely is a good idea -- at least it is a change in behavior. If one wanted to match the previous behavior, they would need to run `Random.State.make_self_init` for each domain, probably in the driver loop at the end of the test.

Then one could argue that tests should rather be deterministic, and that starting from a fixed seed is better. Still I would expect this fixed seed to be different for each domain; with the current code, if I understand think correctly all domains are going to draw the same random sequence, and this feels fairly wrong.

gasche (commented):
I have reviewed the new runtime code and the use of `Random` in other parts of the stdlib (`Filename.temp_file_name` and randomized hashtables), and I believe that those are correct.

However, I am not sure about the behavior we get for the `Random` module itself.

For users that want self-initialization, `Domain.DLS.new_key Random.State.make_self_init` is a good approach (if a bit low-level), it gives the expected behavior of an independent self-initialized seed for all domains. They can also use `Random.int` and other global-state functions to the same effect, as long as they remember to call `Random.self_init` at the creation of each domain. (Are they going to remember this? I'm a bit worried.)

However, for users that want to initialize their generator with a fixed seed, I don't think that the current state is very satisfying. If they don't do anything, all domains are going to be initialized with the same seed, so all domains are going to draw the same random sequence. This is pretty bad, creating unwanted correlations in the program.

What they probably want is to either:
- Set the seed of new domains by consuming random state from the initial domain (to spawn 9 new domains, draw 9 random integers in the initial domain and set the seed of each new domain by the corresponding random integer using `Random.init`). This requires non-trivial logic, I don't think that most Random users would be careful enough to do this.
- Or call `Random.full_init : int array -> unit` with a seed of their choice *and* the domain identifier. This is simple, but I am not sure it is a good idea: if we stop a domain and start another one right after, are we going to get a "fresh" domain identifier, or the same identifier as some previous domain? (The latter would be my guess from skimming the runtime code.)

I think that a multi-domain program and a global-state random generator are fundamentally incompatible and we are never going to get a stdlib design that feels nice anyway. (In particular, we can't hope that a fixed seed will make random-generation deterministic, as long as random-using program fragments are scheduled non-deterministically on the domains.) Trying to do magic things for user convenience is not a good idea, the model must be simple so that users can easily reason about the sort of randomness they get. But I think it's worth giving some thoughts to what the basic needs are, and how to meet them correctly; they should be documented, and, ideally, on the path of least effort.

For me the basic needs are as follows:
- The user wants self-initialization for a multi-domain program. Currently this requires specific thinking when spawning new domains, but predictable and easy to achieve. This should be documented carefully in the Random documentation. I considered the idea of remembering that `self_init` has been called on the main domain and doing it implicitly on other domains, but that would be too magic. 
- The user wants fixed-seed initialization for the program, but different domains should draw different values. I don't know what is a simple and correct approach to do this currently.

Remark: if we had a random generator supporting splitting (for example @xavierleroy's [pringo](https://github.com/xavierleroy/pringo)), I would expect spawning a new domain to split the random generator state between the spawning domain and the spawned domain. Notice that this does "exactly the right thing" in both the self-initialized and fixed-seed scenarios. But this requires a much tighter coupling between the Random module and the runtime, Random cannot be explained anymore as a user-side library with global state.

kayceesrk:
@gasche Thanks for the detailed reviews. I have a feeling that the question of whether `Random` module does the right thing in a multi-domain setting is a question that seems wider than what this PR is trying to do, which is to fix issues with unsafe sharing of random state between multiple domains. As you suggest, the changes are technically correct but whether they are doing the right thing is still an open question. 

In order to make progress, may I suggest splitting the question of `Random` with multiple domains into a separate issue, and we work to merge this PR in? I also don't understand randomness and its implications under multiple domains well, and I shall defer to others who understand the topic well. Who do you suggest may help with this?

gasche:
Yes, I think the split is very reasonable.

> Who do you suggest may help with this?

I strategically pinged @xavierleroy above, because I think he is the right person to ask about this.

kayceesrk:
Great. I've updated `runtime/win32.c` with corresponding change, though we don't yet support . With that it looks like all of your original concerns had been addressed. Is that a correct reading?

stdlib/domain.ml
```diff
@@ -9,6 +9,62 @@ module Raw = struct
...
+        if idx < s then s else compute_new_size (2 * s)
+      in
+      let new_sz = compute_new_size sz in
+      let new_st = Array.make new_sz unique_value in
+      Array.blit st 0 new_st 0 sz;
+      set_dls_state new_st;
+      new_st
+    end
+
+  let set (idx, _init) x =
```
lpw25, line 54: I'm not sure you are being defensive enough in your implementation of an array of `Obj.t`s. The potential issues with the `Obj.t` being a float need to be handled carefully. See `Obj_array` in Base for an example of what is needed: https://github.com/janestreet/base/blob/master/src/obj_array.ml.

lpw25:
> I would expect spawning a new domain to split the random generator state between the spawning domain and the spawned domain. Notice that this does "exactly the right thing" in both the self-initialized and fixed-seed scenarios. But this requires a much tighter coupling between the Random module and the runtime, Random cannot be explained anymore as a user-side library with global state.

Does this suggest a missing feature of domain local state? Perhaps when a new domain is created it should be able to create its new domain local states from the values of the same states in its parent domain? You could probably implement this fairly efficiently by remembering the tree structure on the domains and copying the state array of the parents when they create their children.

stdlib/domain.ml
kayceesrk: Thanks @lpw25! I read through Base's `Obj_array`. IIUC, since we initialise the array with a non-float value, the array will not have `Double_array_tag`. Hence, `get` should be ok [1]. For `set`, if the element is wrapped in `Sys.opaque_identity`, then this will prevent flambda from incorrectly inferring that the array has `Double_array_tag` [2]. Does that address your concerns? Am I missing anything else?

[1] https://github.com/janestreet/base/blob/master/src/obj_array.ml#L50-L53
[2] https://github.com/janestreet/base/blob/master/src/obj_array.ml#L65-L69

xavierleroy:
I'm catching up on the discussion.  Generally speaking, the usual approaches to pseudo-random number generation in multithreaded programs are
- share a single state between all threads, or
- use a splittable PRNG such as SplitMix and split the state when a thread is created.

The latter gives you reproducibility (from a fixed initial state) if all the threads are created before they start generating random numbers.  However, this reproducibility is lost if the state is attached to a domain and lightweight threads are then interleaved in this domain.

The former sounds expensive in terms of synchronization, but with SplitMix for example you only need to update one 64-bit integer, which you could do with a CAS.

@gasche already pointed to my PRINGO library (https://github.com/xavierleroy/pringo).  It's a preliminary investigation of splittable PRNGs that could one day replace OCaml's Random module.

stdlib/domain.ml
lpw25: I think that's right. The `set` part was the one I was concerned about.

kayceesrk:
Thanks @xavierleroy. I've opened a separate issue to keep track of multi-thread compatible PRNGs https://github.com/ocaml-multicore/ocaml-multicore/issues/585.

stdlib/domain.ml
kayceesrk: Thanks!

kayceesrk:
> Does this suggest a missing feature of domain local state? Perhaps when a new domain is created it should be able to create its new domain local states from the values of the same states in its parent domain? You could probably implement this fairly efficiently by remembering the tree structure on the domains and copying the state array of the parents when they create their children.

Sounds reasonable. I've made an issue to track this: https://github.com/ocaml-multicore/ocaml-multicore/issues/587.

stdlib/domain.ml
```diff
@@ -44,6 +103,7 @@ let spawn f =
   let termination_mutex = Mutex.create () in
   let state = Atomic.make Running in
   let body () =
+    DLS.create_dls ();
```
ctk21: Very unlikely, but we should guard this call to `create_dls` as with `f ()` should an exception ever come out of it. e.g. `match (DLS.create_dls (); f ()) with`

testsuite/tests/gc-roots/globroots.ml
ctk21: I agree that this is changing the behaviour of the test. It will correlate all the domains to the same sequence of updates starting with the static seed in `Random.mk_default`.

Using the old `random_state = Domain.DLS.new_key Random.State.make_self_init` is probably better to keep the test behaviour the same.

testsuite/tests/weak-ephe-final/weaklifetime_par.ml
ctk21: I agree that this is changing the behaviour of the test. It will correlate all the domains to the same sequence of updates starting with the static seed in `Random.mk_default`.

Using the old `random_state = Domain.DLS.new_key Random.State.make_self_init` is probably better to keep the test behaviour the same.

ctk21 (commented):
The update to this PR removes my concern about 'stdlib specific slots' and gives us O(1) lookups for user keys, which is much nicer.

I had some minor nits: one place where we weren't guarding exceptions and I think we should retain the existing random behaviour in two of the testsuite changes. 

I think we should document what users should expect when using `Random` with domains.  For example, I think the following short comment in `Random.mli` might be enough:
 - By default the `Random` seed is static and identical across all domains. This means that calling `Random.int N` repeatedly in two domains will give the _same_ sequence of numbers in each domain.
 - If you wish each domain to have uncorrelated sequences of random numbers then one way to achieve that is to call `Random.self_init ()` on each domain before requesting random numbers.

We have #587 for parent-child DLS and #585 for splittable PRNG but I'm keen in the meantime we document what is there.

testsuite/tests/gc-roots/globroots.ml
gasche: In the case of this test, I think that the proposed change is the correct one:
- the new test coincides with the upstream version of the test
- the test is single-domain anyway, so the problematic correlations are not going to show up

ctk21: Although only executed in the `globroots` tests in a single-domain, I think the `Globroots` module is used in other tests with multiple domains. 
I see the `Globroots` module getting used by [globroots_parallel.ml](https://github.com/ocaml-multicore/ocaml-multicore/blob/b9afce3e9a474682f4003965cbc87bb0413e1e2d/testsuite/tests/gc-roots/globroots_parallel.ml) and [globroots_parallel_spawn_burn.ml](https://github.com/ocaml-multicore/ocaml-multicore/blob/b9afce3e9a474682f4003965cbc87bb0413e1e2d/testsuite/tests/gc-roots/globroots_parallel.ml).

gasche: Ah, I had missed this, thanks!

stdlib/domain.ml
gasche: I think the current PR is fine as it is, but I wonder if we shouldn't consider a safe version of this code, using one of the usual "universal type" embeddings. If I understand correctly, the code is not performance-critical, it should be O(1) but we can afford an extra dynamic check on access.

kayceesrk:
Thanks @ctk21. I believe I have addressed all of your comments.

ctk21:
> Thanks @ctk21. I believe I have addressed all of your comments.

Yes, I think it looks good. 

I think we are saying that the CI fail is #588 and unrelated to this PR, so will merge?

kayceesrk:
Yep. CI failure is unrelated and I am working on a fix for it.

runtime/sys.c
```diff
@@ -586,7 +586,7 @@ CAMLprim value caml_sys_random_seed (value unit)
   /* If the read from /dev/urandom fully succeeded, we now have 96 bits
      of good random data and can stop here.  Otherwise, complement
      whatever we got (probably nothing) with some not-very-random data. */
-  if (n < 12) {
+  if (n < 11) {
```
gasche, line 589: Note: after thinking harder about this, I think that changing `n < 12` to `n < 11` is not the best way to do this change. (If we get exactly 11 bytes from /dev/urandom, we do want to add some non-random seeds; currently the after-PR implementation will lack a bit of entropy in this weird corner case that will probably not occur in practice.)

I proposed a refactoring of the upstream function in https://github.com/ocaml/ocaml/pull/10472 that would have made it easier to correctly extend the function.

PR #586, opened 2021-06-16
Make the state in Format domain-local

### LINKED ISSUE #563: Format.printf raises exception when using multiple domains
Running this code with `4.12.0+multicore` gives `Stdlib.Queue.Empty`:
```ocaml
let () =
  let domains = Array.init 7 (fun i ->
    Domain.spawn (fun () ->
          for j = 1 to 10000000 do () done;
          Format.printf "Running in domain %d@." i;
          flush stdout;
      )
    ) in
  Array.iter Domain.join domains
```

```
$ ocamlopt -o multi.exe multi.ml && ./multi.exe
Running in domain 0
Running in domain 
1
5
Running in domain Running in domain 2Fatal error: exception Stdlib.Queue.Empty
```

kayceesrk:
This fixes #563. 

Depends on #582. 

For the wider discussion about the format state and the right thing to do, see [ocaml#10453](https://github.com/ocaml/ocaml/issues/10453).


stdlib/format.mli
```diff
@@ -929,13 +928,13 @@ val formatter_of_out_channel : out_channel -> formatter
 *)
 
 val std_formatter : formatter
-(** The standard formatter to write to standard output.
+(** The initiali domain's standard formatter to write to standard output.
```
jberdine: initial~i~

ctk21 (commented):
I'm _not_ a `Format` module expert.

I was looking at the intent (proposal section in ocaml#10453) and from my read:
- we should add a testcase based on issue #563 to the PR showing that the bug is fixed
- I didn't see any changes in `format.mli` that would allow for 
```
Provide additional functions to access the domain-local stdout, stderr and string formatters and the string buffer.
```

I also wonder if we add a short sentence or two to the documentation explaining what users should expect if they call print functions in two domains; safety but interleaving is for the user to fix.

ctk21:
Thanks, the changes address all of the concerns I had above.

kayceesrk:
Commit https://github.com/ocaml-multicore/ocaml-multicore/pull/586/commits/a1e5baf371e2b8a13c7157735e5a810efe065e8f fixes #592.

stdlib/domain.ml
```diff
@@ -99,11 +99,51 @@ let rec spin f =
...
+
 let spawn f =
+  do_at_first_spawn ();
   let termination_mutex = Mutex.create () in
   let state = Atomic.make Running in
   let body () =
-    let result = match DLS.create_dls (); f () with
+    DLS.create_dls ();
+    let res = f () in
+    do_at_exit ();
```
ctk21: If `f ()` raises an exception then `do_at_exit ()` will not execute, is that intentional or do we always want to execute `do_at_exit ()`?

kayceesrk:
The last commit takes @jberdine's idea from https://github.com/ocaml-multicore/ocaml-multicore/pull/591 and makes it the default when multiple domains are spawned.

stdlib/domain.ml
kayceesrk: Good point. `do_at_exit` should always execute. I'll change this now.

stdlib/domain.ml
```diff
@@ -137,15 +137,11 @@ let spawn f =
...
-    res
-  in
-  let wrapped_body () =
-    let result = match body () with
+    let result = match DLS.create_dls (); f () with
       | x -> Ok x
-      | exception ex -> Error ex in
+      | exception ex -> Error ex
+    in
+    do_at_exit ();
```
ctk21, line 145: Given that `do_at_exit ()` may also throw an exception (the API exposes this to user defined functions), we probably need to guard it as well. Otherwise an at_exit handler which throws in a child domain is going to hang domains that try to join it.

kayceesrk: I've changed the behaviour in https://github.com/ocaml-multicore/ocaml-multicore/pull/586/commits/4ce92d14376ffca1ed2580f5b0150a61d7059891 such that exceptions from `Domain.at_exit` callbacks are ignored. The documentation also mentions this. Sadly, this is not the case with `Stdlib.at_exit`, which can prevent other `at_exit` functions from running since the functions are evaluated in LIFO order. 

@talex5 @craigfe Wondered if you had an opinion on this?

testsuite/tests/lib-format/mc_pr586_par2.ml
```diff
@@ -0,0 +1,12 @@
+(* TEST *)
+
+let () =
+  let domains = Array.init 7 (fun i ->
+    Domain.spawn (fun () ->
+          for j = 1 to 10000000 do () done;
+          for j = 1 to 100 do
+            Format.printf "21. +. 21. is %f\n" (21. +. 21.);
```
Sudha247: It looks like `mc_pr586_par2.ml` covers what's in `mc_pr586_par.ml` too. Would it make a difference if we retain only one of them?

stdlib/format.mli
```diff
@@ -1210,10 +1237,16 @@ val fprintf : formatter -> ('a, formatter, unit) format -> 'a
 *)
 
 val printf : ('a, formatter, unit) format -> 'a
-(** Same as [fprintf] above, but output on [std_formatter]. *)
+(** Same as [fprintf] above, but output on [get_std_formatter ()].
+
+    It is defined as [fun fmt -> fprintf (get_std_formatter ()) fmt].
```
jberdine: Perhaps it would be better to add something to reflect the delaying in the latest implementation. Maybe: "It is defined similarly to [fun fmt -> fprintf (get_std_formatter ()) fmt] but delays calling [get_std_formatter] until after the final argument required by the [format] is received."

stdlib/format.ml
```diff
@@ -1102,82 +1154,84 @@ let formatter_of_symbolic_output_buffer sob =
...
+and close_tag v = pp_close_tag (DLS.get std_formatter_key) v
+and open_stag v = pp_open_stag (DLS.get std_formatter_key) v
+and close_stag v = pp_close_stag (DLS.get std_formatter_key) v
+and print_as v = pp_print_as (DLS.get std_formatter_key) v
+and print_string v = pp_print_string (DLS.get std_formatter_key) v
+and print_int v = pp_print_int (DLS.get std_formatter_key) v
+and print_float v = pp_print_float (DLS.get std_formatter_key) v
+and print_char v = pp_print_char (DLS.get std_formatter_key) v
+and print_bool v = pp_print_bool (DLS.get std_formatter_key) v
+and print_break v = pp_print_break (DLS.get std_formatter_key) v
```
jberdine: I suspect that functions like `print_break`, `set_geometry`, etc. (all the functions that take more than one arg) should be eta-expanded to avoid eager `DLS.get` calls similar to the issue `printf` avoids.

stdlib/format.ml
```diff
@@ -1102,82 +1154,84 @@ let formatter_of_symbolic_output_buffer sob =
...
+and print_tab v = pp_print_tab (DLS.get std_formatter_key) v
+
+and set_margin v = pp_set_margin (DLS.get std_formatter_key) v
+and get_margin v = pp_get_margin (DLS.get std_formatter_key) v
+
+and set_max_indent v = pp_set_max_indent (DLS.get std_formatter_key) v
+and get_max_indent v = pp_get_max_indent (DLS.get std_formatter_key) v
+
+and set_geometry ~max_indent =
+  pp_set_geometry (DLS.get std_formatter_key) ~max_indent
```
jberdine: eta

stdlib/format.ml
```diff
@@ -1102,82 +1154,84 @@ let formatter_of_symbolic_output_buffer sob =
...
+and safe_set_geometry ~max_indent =
+  pp_safe_set_geometry (DLS.get std_formatter_key) ~max_indent
```
jberdine: eta

stdlib/format.ml
```diff
@@ -1102,82 +1154,84 @@ let formatter_of_symbolic_output_buffer sob =
...
+and print_cut v = pp_print_cut (DLS.get std_formatter_key) v
+and print_space v = pp_print_space (DLS.get std_formatter_key) v
+and force_newline v = pp_force_newline (DLS.get std_formatter_key) v
+and print_flush v = pp_print_flush (DLS.get std_formatter_key) v
+and print_newline v = pp_print_newline (DLS.get std_formatter_key) v
+and print_if_newline v = pp_print_if_newline (DLS.get std_formatter_key) v
+
+and open_tbox v = pp_open_tbox (DLS.get std_formatter_key) v
+and close_tbox v = pp_close_tbox (DLS.get std_formatter_key) v
+and print_tbreak v = pp_print_tbreak (DLS.get std_formatter_key) v
```
jberdine: eta

stdlib/format.mli
```diff
@@ -30,14 +30,22 @@
    - {!std_formatter} outputs to {{!Stdlib.stdout}stdout}
    - {!err_formatter} outputs to {{!Stdlib.stderr}stderr}
 
-   Most functions in the {!Format} module come in two variants:
-   a short version that operates on {!std_formatter} and the
-   generic version prefixed by [pp_] that takes a formatter
-   as its first argument.
+   Most functions in the {!Format} module come in two variants: a short version
+   that operates on the current domain's {!std_formatter} and the generic
```
jberdine: It seems to me that "the current domain's {!std_formatter}" is somewhat unclear (since `std_formatter` refers to the initial, not current, domain), and it would be better to write something like "the current domain's standard formatter as obtained by {!get_std_formatter}". I also wonder if it would be a good idea to clarify that the `get_std_formatter` call is internally performed after the last argument is received.

stdlib/format.ml
```diff
@@ -1026,18 +1026,45 @@ module DLS = Domain.DLS
...
+
+let buffered_out_flush oc key () =
+  let buf = Domain.DLS.get key in
+  let len = Buffer.length buf in
+  let str = Buffer.contents buf in
+  output_substring oc str 0 len ;
+  Buffer.clear buf
+
+let std_buf_key = Domain.DLS.new_key (fun () -> Buffer.create 80)
+let err_buf_key = Domain.DLS.new_key (fun () -> Buffer.create 80)
```
jberdine: I now see that above `pp_buffer_size = 512` is used for the size of `str_formatter`. Maybe it would be better to use the same buffer size here.

stdlib/format.ml
```diff
@@ -1026,18 +1026,45 @@ module DLS = Domain.DLS
...
 let std_formatter_key = DLS.new_key (fun () ->
-  formatter_of_out_channel Stdlib.stdout)
+  let ppf =
+    pp_make_formatter (buffered_out_string std_buf_key)
+      (buffered_out_flush Stdlib.stdout std_buf_key) ignore ignore ignore
+  in
+  ppf.pp_out_newline <- display_newline ppf;
+  ppf.pp_out_spaces <- display_blanks ppf;
+  ppf.pp_out_indent <- display_indent ppf;
+  ppf)
```
jberdine, line 1054: My reading of this is that it accumulates output in `std_buf` until the formatter's `pp_out_flush` function is called. This happens when {`pp_`}`print_flush` is called, and hence also when {`pp_`}`print_newline` is called. But it is not called for `printf "@\n"` (or for `printf "\n"`). So I think that the docs for the current implementation should state that output is interleaved at points where the formatter is flushed. If interleaving at `printf "@\n"` is desired (to have a much smaller expected memory consumption effect), then the `pp_out_newline` function should also flush the buffer. That implementation would also probably be better documented by something along the lines of "output is interleaved at points where the formatter splits a line, such as with `pp_print_cut`, but not necessarily at literal newline characters `\n`".

stdlib/format.mli
```diff
@@ -41,8 +41,7 @@
    If multiple domains write to the same output channel such as
    {{!Stdlib.stdout}stdout} and {{!Stdlib.stderr}stderr} using distinct
    formatters on each domain, the output from the domains will be interleaved
-   with each other. It is expected that the user of this library will implement
-   the necessary synchronisation.
+   with each other at new lines.
```
jberdine: Since this only applies to the predefined formatters on stdout and stderr, I think that keeping the comment about client-side synch would be a good idea. Perhaps:

   Warning: Since {{!section:formatter}formatters} contain mutable
   state, it is not thread-safe to use the same formatter on multiple
   domains in parallel without synchronization.

   If multiple domains write write to the same output channel using
   the predefined formatters (as obtained by {!get_std_formatter} or
   {!get_err_formatter}), the output from the domains will be
   interleaved with each other at points where the formatters are
   flushed, such as with {!print_flush}. This synchronization is not
   performed by formatters obtained from {!formatter_of_out_channel}
   (on the standard out channels or others).

jberdine (commented):
Thanks, I think that the addition of the at first spawn functionality is a significant improvement. I tried to simplify the client code to make the formatters synchronized, but it is still necessary to remember to do something after each domain is spawned and before the formatters are used.

I have a question about the buffering implementation, and have made a few comments on the documentation.

It is not necessary, but I can imagine that users might want the buffering implementation to be exposed so that it could easily be used e.g. for a formatter on a channel to a file.

The code I was playing with before the run at first spawn functionality is:
```ocaml
let synchronize_formatter get_ppf =
  Domain.DLS.new_key (fun () ->
      let ppf = get_ppf () in
      let out_funs = Format.pp_get_formatter_out_functions ppf () in
      let buf = Buffer.create 80 in
      let out_string str ofs len = Buffer.add_substring buf str ofs len in
      let out_flush () =
        out_funs.out_string (Buffer.contents buf) 0 (Buffer.length buf) ;
        Buffer.clear buf ;
        out_funs.out_flush ()
      in
      Format.pp_set_formatter_out_functions ppf
        {out_funs with out_string; out_flush} )

let () =
  let domains = Array.init 7 (fun i ->
    Domain.spawn (fun () ->
          Domain.DLS.get (synchronize_formatter Format.get_std_formatter);
          for j = 1 to 10000000 do () done;
          for j = 1 to 100 do
            Format.printf "The quick brown fox@\n";
            Format.printf "jumped over the lazy dog@.";
          done
      )
    ) in
  Array.iter Domain.join domains

let make_synchronized_formatter out flush =
  Domain.DLS.new_key (fun () ->
      let buf = Buffer.create 80 in
      let out_string str ofs len = Buffer.add_substring buf str ofs len in
      let out_flush () =
        out (Buffer.contents buf) 0 (Buffer.length buf) ;
        Buffer.clear buf ;
        flush ()
      in
      Format.make_formatter out_string out_flush )

let std_formatter_sync =
  make_synchronized_formatter
    (Stdlib.output_substring Stdlib.stdout)
    (fun () -> Stdlib.flush Stdlib.stdout )

let () =
  let domains = Array.init 7 (fun i ->
    Domain.spawn (fun () ->
          let ppf = Domain.DLS.get std_formatter_sync in
          for j = 1 to 10000000 do () done;
          for j = 1 to 100 do
            Format.fprintf ppf "The quick brown fox@\n";
            Format.fprintf ppf "jumped over the lazy dog@.";
          done
      )
    ) in
  Array.iter Domain.join domains
```

stdlib/domain.ml
talex5: I never use `at_exit`, so I don't really mind, though I'm generally against anything that silently drops exceptions.

stdlib/format.mli
kayceesrk: > I also wonder if it would be a good idea to clarify that the get_std_formatter call is internally performed after the last argument is received.

Good point. I've added

```
For the version that operates on the current domain's standard formatter, the 
call to {!get_std_formatter} is delayed until the last argument is received.
```

Sudha247 (commented):
This would increase Multicore's compatibility with other libraries! The DLS instances look correct to me, had a question on the new tests.

stdlib/format.ml
kayceesrk: I shall update the documentation to say that 

```
the output is interleaved at points where the formatter is flushed
```

I like the current formatter for the reason that we can do multi-line printing from a domain and not have it interleaved with output from other domains. I don't know whether this is a reasonable usecase, but I may wish to the produce log from multiple domains as a well-formed markdown file:

```md
* Domain 0 did Event e
  + Variable a was xxx
  + Variable b was yyy
```

Assuming that a flush is done after the whole event is emitted, the current version will produce unsurprising behaviours.  The same would be true for emitting well-formatted json, where each event is an item in an array.

testsuite/tests/lib-format/mc_pr586_par2.ml
kayceesrk: Good point. I'll remove `mc_pr586_par.ml`.

stdlib/format.mli
```diff
@@ -31,22 +31,26 @@
...
-   If multiple domains write to the same output channel such as
-   {{!Stdlib.stdout}stdout} and {{!Stdlib.stderr}stderr} using distinct
-   formatters on each domain, the output from the domains will be interleaved
-   with each other at new lines.
-
    Warning: Since {{!section:formatter}formatters} contain mutable state, it is
    not thread-safe to use the same formatter on multiple domains in parallel
    without synchronisation.
 
+   If multiple domains write write to the same output channel using the
```
jberdine:
```suggestion
   If multiple domains write to the same output channel using the
```

testsuite/tests/lib-format/mc_pr586_par.ml
```diff
@@ -5,8 +5,7 @@ let () =
     Domain.spawn (fun () ->
           for j = 1 to 10000000 do () done;
           for j = 1 to 100 do
-            Format.printf "The quick brown fox jumped over the lazy dog\n";
-            flush stdout
+            Format.printf "21. +. 21. is %f\n" (21. +. 21.);
```
jberdine: Note that `\n` does not flush the formatter, and the output does not reveal what the interleaving between domains was. I suspect that this test currently runs with completely serial access to the out channel.
```suggestion
            Format.printf "21. +. 21. is %f@." (21. +. 21.);
```

stdlib/format.ml
jberdine: I agree, and think that such use cases are legit. My only concern is that there is a potential for code that produces a lot of output before flushing to consume significantly more memory when run with such a buffered formatter. This could be memory consumption with asymptotic complexity that is linear in the number of execution steps. It can be avoided, but I'm not sure how it fits with the goal of how unchanged sequential code runs on multicore vs the sequential runtime. If there is a goal for existing code to not take asymptotically more memory on multicore, then the risk could be reduced by weakening to interleaving at newlines. I don't know the right trade-off.

jberdine (commented):
Thanks! I noticed one typo and I don't think the test is testing what you want.

kayceesrk:
Thanks Josh. I'll add `make_synchronized_formatter` and `synchronized_formatter_of_out_channel`. They seem useful to have.

kayceesrk:
> I'll add make_synchronized_formatter and synchronized_formatter_of_out_channel.

Added now. The interface is quite pleasant. 

I think I've addressed all the outstanding comments.

ctk21:
I make this PR ready to merge. It looks to me that everyones comments have been addressed.

kayceesrk:
Thanks @ctk21. I'm merging it now.

jberdine:
Nice, this is great, thanks @kayceesrk!

PR #589, opened 2021-06-16
Reintroduce adjust_gc_speed

abbysmal:
This PR reintroduce the `caml_adjust_gc_speed` function in the runtime.
The core logic of the function is mostly unchanged from trunk.

`Caml_state` now holds a specific variable (extra_heap_resources) to track additional work to be done by the garbage collector during a major collection.
`caml_adjust_gc_speed` will set this variable to a desired factor, that will be taken into account in `update_major_slice_work`.
It will be as well reset (to 0.0) at the end of work computation.

I've reintroduced the call to `adjust_gc_speed` in `alloc_custom_gen` as well. (in the default branch where custom values are allocated onto the shared heap, since the minor heap case is commented away.)

### DIFF
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -369,8 +369,9 @@ static void update_major_slice_work() {
   if( saved_terminated_words > 0 ) {
     while(!atomic_compare_exchange_strong(&terminated_domains_allocated_words, &saved_terminated_words, 0));
   }
-
   p = (double) (saved_terminated_words + dom_st->allocated_words) * 3.0 * (100 + caml_percent_free) / heap_words / caml_percent_free / 2.0;
+
+  if (p < dom_st->extra_heap_resources) p = dom_st->extra_heap_resources;
 
   if (p > 0.3) p = 0.3;
 
@@ -416,6 +417,7 @@ static void update_major_slice_work() {
 
   dom_st->stat_major_words += dom_st->allocated_words;
   dom_st->allocated_words = 0;
+  dom_st->extra_heap_resources = 0.0;
 }
 
 static intnat get_major_slice_work(intnat howmuch) {
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -344,7 +344,7 @@ static void update_major_slice_work() {
                     = dom_st->allocated_words * 3 * (100 + caml_percent_free)
                       / (2 * heap_words * caml_percent_free)
      Proportion of extra-heap resources consumed since the previous slice:
-                 PE = caml_extra_heap_resources
+                 PE = dom_st->extra_heap_resources
      Proportion of total work to do in this slice:
                  P  = max (PH, PE)
      Amount of marking work for the GC cycle:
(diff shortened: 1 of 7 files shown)


runtime/major_gc.c
```diff
@@ -369,9 +369,10 @@ static void update_major_slice_work() {
   if( saved_terminated_words > 0 ) {
     while(!atomic_compare_exchange_strong(&terminated_domains_allocated_words, &saved_terminated_words, 0));
   }
-
   p = (double) (saved_terminated_words + dom_st->allocated_words) * 3.0 * (100 + caml_percent_free) / heap_words / caml_percent_free / 2.0;
 
+  if (p < dom_st->extra_heap_resources) p = dom_st->extra_heap_resources;
+
```
ctk21, line 375: Can we add this debug message (as in upstream) after the 'raw work-to-do' just in case anyone wants it:
```
 caml_gc_message (0x40, "extra_heap_resources = %"
                             ARCH_INTNAT_PRINTF_FORMAT "uu\n",
                      (uintnat) (dom_st->extra_heap_resources * 1000000));
```

runtime/major_gc.c
```diff
@@ -369,9 +369,10 @@ static void update_major_slice_work() {
   if( saved_terminated_words > 0 ) {
     while(!atomic_compare_exchange_strong(&terminated_domains_allocated_words, &saved_terminated_words, 0));
   }
-
   p = (double) (saved_terminated_words + dom_st->allocated_words) * 3.0 * (100 + caml_percent_free) / heap_words / caml_percent_free / 2.0;
```
ctk21, line 372: Can we update `caml_extra_heap_resources` to `dom_st->extra_heap_resources` in the comment block above?

runtime/minor_gc.c
```diff
@@ -622,6 +624,7 @@ void caml_empty_minor_heap_promote (caml_domain_state* domain, int participating
     if (Is_block(*v) && Is_young(*v)) {
       if (get_header_val(*v) == 0) { /* value copied to major heap */
```
ctk21, line 624: I actually think we need to do `caml_adjust_gc_speed(elt->mem, elt->max);` here after `if (Is_block(*v) && Is_young(*v)) {`.

In its current location, we will only apply the adjustment if the custom block has been promoted by another domain or root. We are promoting all the custom blocks so should do the adjustment for every element that is a block.

runtime/memory.c
```diff
@@ -689,3 +689,14 @@ CAMLexport wchar_t* caml_stat_wcsconcat(int n, ...)
 }
 
 #endif
+
+CAMLexport void caml_adjust_gc_speed (mlsize_t res, mlsize_t max)
```
ctk21: Please can we import the upstream ocaml comment that goes with this block:
```C
/* Use this function to tell the major GC to speed up when you use
   finalized blocks to automatically deallocate resources (other
...
```

runtime/minor_gc.c
```diff
@@ -512,6 +512,8 @@ void caml_empty_minor_heap_domain_clear (caml_domain_state* domain, void* unused
   clear_table ((struct generic_table *)&minor_tables->ephe_ref);
   clear_table ((struct generic_table *)&minor_tables->custom);
 
+  Caml_state->extra_heap_resources_minor = 0.0;
```
ctk21: Please can we uses the `domain` variable over `Caml_state`, we're trying to keep the uses of `Caml_state` to only those places where necessary.

ctk21 (commented):
This PR is great, it unlocks a chunk of code that uses `caml_adjust_gc_speed`.

Couple of cleanups and I think the speed adjustment is not quite in the right place in `caml_empty_minor_heap_promote`.

runtime/minor_gc.c
abbysmal: Right, that does make sense, thanks for spotting this.

ctk21:
Thanks, I think this one looks good to merge.

PR #591, opened 2021-06-16
Add test implementing Format output synchronization via double buffering

jberdine:
This is just one commit on top of https://github.com/ocaml-multicore/ocaml-multicore/pull/586, is there a way to get github to understand this and display it better?


kayceesrk:
Thanks for the prototype @jberdine. For future reference, the buffering implementation proposed in this PR has been made the default in Format when multiple domains are in use. See https://github.com/ocaml-multicore/ocaml-multicore/pull/586#issuecomment-865101345.

PR #593, opened 2021-06-18
Fix two issues with ephemerons

### LINKED ISSUE #545: ephetest6 fails with more number of domains
The test `ephetest6.ml` can fail when more number of domains are spawned like in #544.

To reproduce: 
```USE_RUNTIME=d OCAMLRUNPARAM=v=0,V=1 taskset --cpu-list 2 make -C testsuite one DIR=tests/weak-ephe-final```

Results in:
```
### begin stdout ###
 test1: ok
 [00] file major_gc.c; line 771 ### Assertion failed: Tag_val(v) == Abstract_tag
### end stdout ###
```

### PR DESCRIPTION
**Pull request by kayceesrk**
This PR simplifies ephemeron handover during termination. 

Primarily, we no longer hand over ephemerons that are yet to be processed in the current cycle (in the `todo` list). Instead, we force the ephemerons yet to be processed and their data to be live in the current cycle such that there is no work to be done for those ephemerons in the current cycle. This moves all the ephemerons to the `live` list. This simplifies invariants quite a bit. The downside is that ephemerons and their data on a domain that is currently terminating will be GCed in the next cycle. This cost is small compared to the simplification benefits. 

Fixes #545 and #588.

### DIFF
--- a/runtime/caml/major_gc.h
+++ b/runtime/caml/major_gc.h
@@ -62,7 +62,6 @@ void caml_orphan_allocated_words();
 void caml_add_to_orphaned_ephe_list(struct caml_ephe_info* ephe_info);
 void caml_add_orphaned_finalisers (struct caml_final_info*);
 void caml_final_domain_terminate (caml_domain_state *domain_state);
-void caml_adopt_orphaned_work (void);
 
 struct heap_stats {
   intnat pool_words;
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1119,11 +1119,8 @@ static void handover_ephemerons(caml_domain_state* domain_state)
     return;
 
   caml_add_to_orphaned_ephe_list(domain_state->ephe_info);
-  if (domain_state->ephe_info->todo != 0) {
-    caml_ephe_todo_list_emptied();
-  }
-  domain_state->ephe_info->live = 0;
-  domain_state->ephe_info->todo = 0;
+  Assert (domain_state->ephe_info->live == 0);
+  Assert (domain_state->ephe_info->todo == 0);
 }
 
 static void handover_finalisers(caml_domain_state* domain_state)
--- a/runtime/major_gc.c
+++ b/runtime/major_gc.c
@@ -98,10 +98,12 @@ struct ephe_cycle_info_t {
  * the lock. */
 static caml_plat_mutex ephe_lock = CAML_PLAT_MUTEX_INITIALIZER;
 
-static void update_ephe_info_for_marking_done ()
+static void ephe_next_cycle ()
 {
   caml_plat_lock(&ephe_lock);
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by kayceesrk** [comment]
ugh. The CI is failing with the same bug as this one is purportedly fixing. I've marked this PR as draft. I'll continue working on it next week.

**Comment by kayceesrk** [comment]
I've modified the original description to reflect the current state of the PR.

File: runtime/caml/major_gc.h [hunk 1]
```diff
@@ -62,7 +62,6 @@ void caml_orphan_allocated_words();
 void caml_add_to_orphaned_ephe_list(struct caml_ephe_info* ephe_info);
```
ctk21, line 62: Because it is no longer used elsewhere, can we remove `caml_ephe_todo_list_emptied` from the interface exposed by `major_gc.h`?

ctk21 (commented):
This code is fiddly. However I think it does achieve the intent of forcing all ephemerons on the terminating domain to be alive for this GC cycle and then placing them in the orphaned list for another domain to deal with. 

I have a niggling feeling that we have another bug (although not currently exercised by the test code):
we setup `ephe_cycle_info.num_domains_todo` in `cycle_all_domains` at the end of a major GC cycle to `num_domains_in_stw`. However I couldn't see anywhere how we handle the case of a domain that is spawned in the middle of a major GC cycle and how that might then start altering `ephe_cycle_info.num_domains_done`

Put another way, we never increment `ephe_cycle_info.num_domains_todo` however on every domain termination, we will decrement it by 1 (via `caml_add_to_orphaned_ephe_list` and `caml_ephe_todo_list_emptied`). Doesn't this mean that we can have `num_domains_todo` reaching arbitrary negative numbers and then hanging the GC process (`is_complete_phase_sweep_and_mark_main` will never pass)? 

Not for this PR, but we should consider renaming the ephemeron state variables to match up with the [ICFP Retrofitting Parallelism](https://arxiv.org/pdf/2004.11663.pdf) Appendix A pseudo-code as it may make things easier to get into.

kayceesrk:
Regarding the question, a newly spawned domain will only ever have ephemerons in the live list. This is because new objects are allocated marked. Since num_domains_todo is only ever decremented by domains that had emptied their todo list. By definition, newly spawned domains will neither decrement num_domains_todo nor increment num_domains_done.

ctk21:
> Regarding the question, a newly spawned domain will only ever have ephemerons in the live list. This is because new objects are allocated marked. Since num_domains_todo is only ever decremented by domains that had emptied their todo list. By definition, newly spawned domains will neither decrement num_domains_todo nor increment num_domains_done.

Thanks, armed with your comment, I now understand what's going on. The only way to end up on the todo list is by being a participant of the last major cycle STW completion.

kayceesrk:
Thanks for the review @ctk21!

PR #596, opened 2021-06-22
systhreads: do st_thread_id after initializing the thread descriptor

abbysmal:
This PR address the issue found at #528 (`linux-O0` build.)

The mistake is pretty easy to spot, the thread id was set before initializing the related thread descriptor.

I'm not quite sure why this is only showing up in the `O0` build, I would guess some checks or accesses are elided during a following optimization pass.


ctk21:
This LGTM. The `-O2` optimizations are noticing that `st_thread_set_id(Ident(th->descr))` can be removed because `st_thread_set_id` inlines to being null.

PR #598, opened 2021-06-26
Don't deliver signals to threads that have blocked them

sadiqj:
At present if a thread blocks a signal it might still receive it. This PR incorporates some of the changes from https://github.com/ocaml/ocaml/pull/2211 to fix that.

It's a draft because I _think_ we still need to handle the without-systhreads case. I couldn't find the right initialisation hook to switch out the sigprocmask wrapper (as the ocaml PR does).

cc @Engil @ctk21

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1099,9 +1099,39 @@ CAMLexport void (*caml_enter_blocking_section_hook)(void) =
 CAMLexport void (*caml_leave_blocking_section_hook)(void) =
    caml_leave_blocking_section_default;
 
+static int check_for_pending_signals(void)
+{
+  int i;
+  for (i = 0; i < NSIG; i++) {
+    if (caml_pending_signals[i]) return 1;
+  }
+  return 0;
+}
+
 CAMLexport void caml_leave_blocking_section() {
-  caml_leave_blocking_section_hook();
+  int saved_errno;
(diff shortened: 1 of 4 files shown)


runtime/domain.c
```diff
@@ -1099,9 +1099,39 @@ CAMLexport void (*caml_enter_blocking_section_hook)(void) =
...
+     pending signals (masked in the other thread). To handle this
+     case, we force re-examination of all signals by setting it back
+     to 1.
+     Another case where this is necessary (even in a single threaded
+     setting) is when the blocking section unmasks a pending signal:
+     If the signal is pending and masked but has already been
+     examined by [caml_process_pending_signals_exn], then
+     [signals_are_pending] is 0 but the signal needs to be
+     handled at this point. */
+  if (check_for_pending_signals()) {
```
ctk21, line 1128: I wish there were a better way than scanning an array of size `NSIG = 64` on every `caml_leave_blocking_section` call.

abbysmal: Wouldn't a bitfield do the trick to avoid going through the array all the time, and act as a cost-free-ish barrier to these computations?

sadiqj:
Closing this PR as https://github.com/ocaml-multicore/ocaml-multicore/pull/630 contains a fix for this issue too.

Issue #448, opened 2020-11-26
reintroduce caml_stat_ accessors in the C API


abbysmal:
We were checking with @kit-ty-kate some recent build failures on the Opam CI for Multicore
core kernel currently fails on such errors:
```
- gc_stubs.c: In function 'minor_words':
- gc_stubs.c:7:20: error: 'caml_stat_minor_words' undeclared (first use in this function)
-     7 |   return (intnat) (caml_stat_minor_words +
-       |                    ^~~~~~~~~~~~~~~~~~~~~
- gc_stubs.c:7:20: note: each undeclared identifier is reported only once for each function it appears in
- gc_stubs.c:8:23: error: 'caml_young_end' undeclared (first use in this function)
-     8 |             (double) (caml_young_end - caml_young_ptr));
-       |                       ^~~~~~~~~~~~~~
- gc_stubs.c:8:40: error: 'caml_young_ptr' undeclared (first use in this function)
-     8 |             (double) (caml_young_end - caml_young_ptr));
-       |                                        ^~~~~~~~~~~~~~
- gc_stubs.c: In function 'promoted_words':
- gc_stubs.c:13:20: error: 'caml_stat_promoted_words' undeclared (first use in this function)
-    13 |   return ((intnat) caml_stat_promoted_words);
-       |                    ^~~~~~~~~~~~~~~~~~~~~~~~
- gc_stubs.c: In function 'major_words':
- gc_stubs.c:23:20: error: 'caml_stat_major_words' undeclared (first use in this function)
-    23 |   return (intnat) (caml_stat_major_words + (double) caml_allocated_words);
-       |                    ^~~~~~~~~~~~~~~~~~~~~
- gc_stubs.c:23:53: error: 'caml_allocated_words' undeclared (first use in this function); did you mean 'caml_orphan_allocated_words'?
-    23 |   return (intnat) (caml_stat_major_words + (double) caml_allocated_words);
-       |                                                     ^~~~~~~~~~~~~~~~~~~~
-       |                                                     caml_orphan_allocated_words
- In file included from /home/opam/.opam/4.10.0+multicore+no-effect-syntax/lib/ocaml/caml/gc.h:20,
-                  from /home/opam/.opam/4.10.0+multicore+no-effect-syntax/lib/ocaml/caml/memory.h:26,
-                  from gc_stubs.c:2:
- gc_stubs.c: In function 'core_kernel_gc_minor_collections':
[...]

PR #600, opened 2021-06-30
Expose a few more GC variables in headers

abbysmal:
This PR exposes a few more internals that seems to be used by the outside world, from what I gather from the Opam CI:
http://check.ocamllabs.io:8082/log/1624377054-df20c61009ce6b16d04ffcfe43ef26c629c2599c/4.12+domains+gc_stubs/bad/grenier.0.12
See #448 as well.


ctk21:
LGTM, merging

PR #603, opened 2021-06-30
Systhreads tick thread

abbysmal:
This is a first attempt at porting the tick thread to Multicore.
Since we are addressing the signal situation in the runtime (#584, #598) I felt that it was useful to try to address this piece of the puzzle as well and see how it fits with the overall signal story.

This needs to be refined a fair bit, there's a few details I'm not fond of:

- I don't think setting up the domain through `caml_init_domain_self` in the main tick thread function is required: I think it could be possible to record a signal to a domain directly rather than having the thread do it and interrupt itself.
- sticking to the current code structure makes the usage of another `Max_domain` size array (containing `tick_thread_stop`) unfortunate. I didn't think this through yet, there is maybe a more elegant solution.

I re-enabled `testpreempt` as a result of this PR, which does not pass on bytecode. I'm investigating this issue.


abbysmal:
I just pushed major changes to this PR.

With our signal semantics a bit clearer now, it turns out the old way of implementing the tick thread is not reliable:
Indeed we cannot make sure a signal can be processed by a specific domain, which means we cannot reliably force systhreads preemption across many domains if we keep using signal delivery as the mechanism to handle preemption.

@sadiqj and I agreed that it was worth trying to instead interrupt each domains through a special interrupt flag, and then add a hook to allow external code to be called when that interrupt is triggered.

The last few commits implements this idea through the addition of two things in the domain API:
- Caml_state->requested_external_interrupt
- caml_domain_external_interrupt_hook

A tick thread will now:
- spawn when another systhread is spawned on a domain.
- it will trigger an interrupt on the domain and flip `requested_external_interrupt` to 1.
- Once the interrupt is processed by a mutator thread, it will then end up eventually calling in the hook registered by systhreads to force the mutator to `yield` to another thread.

This approach seems to work, though discussion about whether or not it is sensible, and how this API can be improved, are much welcome.

To note, I did not move to a single tick thread approach (versus the current one, where one tick thread is to spawn per domain to handle preemption duty.).
The reason is that I could not find a way to interrupt all domains from outside the runtime's private APIs, and I did not want to infringe further down the path of expanding domain's API.

otherlibs/systhreads/st_posix.h
```diff
@@ -298,6 +316,27 @@ static int st_event_wait(st_event e)
...
+/* The tick thread: posts a SIGPREEMPTION signal periodically */
+
+static void * caml_thread_tick(void * arg)
+{
+  uintnat *domain_id = (uintnat *) arg;
+  struct timeval timeout;
+
+  caml_init_domain_self(*domain_id);
+
+  while(! Tick_thread_stop) {
```
ctk21: Shouldn't this be `atomic_load_acq` given that we are `atomic_load_store` on the variable in `caml_thread_cleanup`

ctk21 (commented):
I think this gives us a pre-emption implementation that works for systhreads. 

minor nit on an unpaird atomic.

otherlibs/systhreads/st_posix.h
abbysmal: Right, good catch, just pushed a commit to address this, thank you!

Issue #574, opened 2021-06-09
mirage-crypto-rng-async doesn't build on 4.12+domains


Sudha247:
The error is:

```
In file included from /home/opam/.opam/4.12.0+domains/lib/ocaml/caml/misc.h:32,
-                  from /home/opam/.opam/4.12.0+domains/lib/ocaml/caml/mlvalues.h:23,
-                  from native/mirage_crypto.h:8,
-                  from native/chacha_generic.c:3:
- /home/opam/.opam/4.12.0+domains/lib/ocaml/caml/camlatomic.h:30:9: warning: ISO C99 does not support the '_Atomic' qualifier [-Wpedantic]
-    30 | typedef _Atomic uintnat atomic_uintnat;
-       |         ^~~~~~~
```

Full log: http://check.ocamllabs.io:8082/log/1623071597-161f03f0a11110be99f478b0ee1506b820e05f82/4.12+domains/bad/mirage-crypto-rng-async.0.10.1

abbysmal:
After looking into it with @kit-ty-kate I submitted #610 to address the first part of this issue.
We discussed it a bit with @Sudha247 that encountered similar issues with Tezos.

The core idea is that many packages are not well behaved and hardcode explicitly the cflags to build code against the runtime.
OCaml is currently using any C99 compliant compiler (`-std=c99`) and a few packages compiles their code under this assumption:
See `mirage-crypto`: https://github.com/mirage/mirage-crypto/blob/bdbc9e45346b7f04e985ecac90a548ae225b77b8/config/cfg.ml#L1
`numerics` in Tezos: https://gitlab.com/tezos/tezos/-/blob/master/vendors/numerics/lib/dune

These problems should be addressed by making sure these packages retrieves the related cflags the correct way. (by parsing `ocamlc -config`).

I think we can either convert this issue in a bullet list, or open individual issues to track each packages we find with the offending warnings.

abbysmal:
Taking note that `mirage-crypto` is maybe a bit trickier as there is also some pkg-config things going on with `ocaml-freestanding`, and I'm not quite sure what are the full implications, see:
https://github.com/mirage/mirage-crypto/blob/main/src-freestanding/Makefile

abbysmal:
Also noting that the build failure in this issue is caused by `caml_channel_mutex_unlock` missing rather than what we just discussed.
I'm currently working on the `caml_channel_*` hooks to fix this specific issue.

hannesm:
[...]

PR #610, opened 2021-07-05
Add std gnu11 common cflags

abbysmal:
While looking at https://github.com/ocaml-multicore/ocaml-multicore/issues/574 with @Kate and @Sudha247 we noticed that it seems there's a discrepancy in what kind of flags OCaml packages are using to build C stubs against the runtime.

This does not address further issues in the ecosystem (I'll update #574 with the problematic packages.), but at least it allows people using `ocamlc -config` to retrieve the correct C flags to select the right C variant.

For now I added it to Clang and GCC, and I'm not quite sure this is entirely required, however I feel like that being explicit on it is better than not.

### DIFF
--- a/configure
+++ b/configure
@@ -12686,7 +12686,7 @@ case $host in #(
     internal_cflags="-Wno-unused $cc_warnings \
 -fexcess-precision=standard"
         # TODO: see whether the code can be fixed to avoid -Wno-unused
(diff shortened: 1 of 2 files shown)


damiendoligez:
`gnu11` is C11 with gnu extensions, right? That looks reasonable to me.

Issue #571, opened 2021-06-08
Make extern/intern multi-thread safe and inline with upstream ocaml


ctk21:
The `extern.c` / `intern.c` files have diverged from their upstream ocaml counterparts. This divergence should no longer be necessary, except that needed to make the functions multi-thread safe. 

Note that the marshalling for closures in upstream changed recently [ocaml#9649](https://github.com/ocaml/ocaml/pull/9649)

ctk21:
Related #75

kayceesrk:
Fixed #612.

Issue #75, opened 2016-11-23
Marshalling and unmarshalling


stedolan:
Marshalling has various problems in multicore (the current implementation has just enough hacks to get the compiler to bootstrap). Specifically:

 - Marshalling of closures is broken
Code pointers are treated specially by marshalling. In stock OCaml, code pointers are detected by using the page table. In multicore, there is no page table. Some changes to the closure representation are needed to allow the marshalling code to reliably detect code pointers.

 - Marshalling of Custom_tag objects is strange
The API for marshalling Custom_tag objects is a bit odd: the format does not store the length and custom unmarshallers are passed a large memory region of unknown size in which to unmarshal. This works in stock OCaml because the unmarshaller knows about the heap format, but it doesn't work in multicore. Multicore requires either API changes (done, but not all clients updated), or a marshal format that stores length (probably a better option, since the old API can be kept)

 - Marshalling of concurrently-modified objects is unsafe
There's no particular care taken at the moment to correctly marshal objects being mutated on a different domain. I don't expect the marshaller to give sensible output in this case, but it shouldn't segfault (the current marshaller can crash, in principle, if other domains modify objects being marshalled and a full GC cycle occurs during marshalling, but this is hard to trigger).

Some of the changes to marshalling might be worth upstreaming to stock OCaml independently of multicore.

dhil:
I'd be interested in fixing at least marshalling of closures. In order to give me some idea could you elaborate on what changes to the closure representation are necessary?

stedolan:
The issue with marshalling closures is exactly the one being fixed by ocaml/ocaml#1156

dhil:
Great! I suppose we can back port it once it has been merged.

kayceesrk:
PR https://github.com/ocaml/ocaml/pull/9353 implements marshaling using hash tables instead of modifying GC colour bits in the header.

kayceesrk:
PR#9553 has been merged. The only item left to do is make marshaling safe when the objects are being modified concurrently.

kayceesrk:
Everything except

> Marshalling of concurrently-modified objects is unsafe

is fixed in #612. I'll carve this into a new issue.

PR #612, opened 2021-07-07
Make intern and extern work with multicore

kayceesrk:
This PR brings the upstream changes made to intern and extern to accommodate Multicore OCaml runtime changes. In particular, `Closure_tag` and `Custom_tag` marshalling now works. All the tests disabled due to lack of support for `Closure_tag` and`Custom_tag` marshalling have been re-enabled and they pass the default run of the testsuite. 

Another difference from stock is that Multicore's interning has been made re-entrant. The reason for this is how memory is allocated for the objects in the intern module. On stock, the memory required is allocated in one page and the interning process carves up this memory for the objects. This new page of memory is then attached to the free-list. This doesn't work on Multicore since Multicore uses a size-segmented slab allocator. The allocation has to be made by calling `caml_alloc` which can, in turn, call a GC. This GC call may trigger finalisers to run, which may perform interning (See the test [intern_final.ml](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/testsuite/tests/lib-marshal/intern_final.ml)). Hence, we have to make the interning re-entrant. This change itself is straightforward.

The PR deliberately avoids making marshalling thread-safe with respect to concurrent modifications by another domain. It is unclear how to implement this, let alone efficiently, it is not possible to prevent concurrent modifications short of obtaining a global lock on the runtime, which would stop all of the other domains from executing. Concurrent modification and marshalling of the same object may lead to a crash.


kayceesrk:
Fixes #571 and #75.

runtime/io.c
```diff
@@ -516,6 +526,7 @@ CAMLprim value caml_ml_out_channels_list (value unit)
       Field (res, 0) = chan;
```
ctk21: I got worried about `caml_alloc_channel` and `caml_alloc_small` (which occur inside the `channel->max==NULL` block above) being called while the `caml_all_opened_channels_mutex` is held. 
My worry is:
- one of the allocs causes a garbage collection
- the garbage collection results in a finaliser (or signal handler) running which itself then somehow calls `caml_ml_out_channels_list` which will result in a deadlock. 

Is this series of events impossible?

runtime/extern.c
```diff
@@ -93,61 +82,121 @@ struct position_table {
...
+  struct output_block * next;
+  char * end;
+  char data[SIZE_EXTERN_OUTPUT_BLOCK];
+};
+
+struct caml_extern_state {
+
+  int extern_flags;        /* logical or of some of the flags */
+
+	uintnat obj_counter;  	/* Number of objects emitted so far */
```
ctk21: The indentation is a bit haywire in this struct with a mix of tabs and spaces.

runtime/intern.c
```diff
@@ -326,16 +397,26 @@ static void stack_push_items(struct intern_stack* s, value dest, int n) {
...
 
-static void intern_cleanup(struct intern_stack* s)
+static void intern_cleanup(struct caml_intern_state* s)
 {
-  if (intern_input != NULL) {
-    caml_stat_free (intern_input);
-    intern_input = NULL;
+  s->intern_src = NULL;
+  if (s->intern_input != NULL) {
+    caml_stat_free (s->intern_input);
```
ctk21, line 404: I'm a bit worried this isn't always going to work as in stock OCaml. 

Take the following calls:
 - `caml_input_value_from_malloc` with a buffer passed in with `data` 
 - this will call `intern_cleanup(s)` in `input_val_from_block`
 - which will result in the passed `char * data` having `caml_stat_free` called on it
 
 Have I missed something or is that possible?
 I'm wondering if we need a more explicit way of saying: call this free function on the buffer when you clean it up.

ctk21 (commented):
I think this fixes the known problems we had in multicore with intern and extern. 

I had a couple of queries where some things might not be 100% safe.

runtime/intern.c
kayceesrk: I think freeing the input data is precisely what we want for `caml_input_value_from_malloc`. See https://github.com/ocaml/ocaml/blob/trunk/runtime/caml/intext.h#L135-L139. In particular, if `s->input_data` is not NULL, then it is supposed to be freed. See https://github.com/ocaml/ocaml/blob/trunk/runtime/intern.c#L44-L46. I believe we match the stock semantics. Am I missing something?

runtime/io.c
kayceesrk: Good point. I'll move the allocations outside the loop. It may be easiest to maintain a separate count of the open channels. I'll have a go at this next week.

runtime/intern.c
ctk21: Apologies, looks like I got the wrong end of the stick here.

runtime/io.c
kayceesrk: https://github.com/ocaml-multicore/ocaml-multicore/pull/612/commits/30622d770732fe69a2698198f1ed025bb0f76d64 moves the allocations out of the critical section.

runtime/io.c
```diff
@@ -401,7 +420,12 @@ void caml_finalize_channel(value vchan)
...
   if ((chan->flags & CHANNEL_FLAG_MANAGED_BY_GC) == 0) return;
-  if (--chan->refcount > 0) return;
+  unlink_channel(chan);
+  if (atomic_fetch_add (&chan->refcount, -1) > 1) {
+    /* [caml_ml_out_channels_list] may have a reference to this channel. */
+    link_channel (chan);
+    return;
+  }
   caml_plat_mutex_free(&chan->mutex);
   /* TODO KC: See commented out section */
```
ctk21, line 430: May or may not be for this PR, but do we want to clean out that commented out section while the code is 'in cache'?

ctk21:
I think the changes in `io.c` make sense and so make things multi-thread safe. 

I did wonder if we had any explicit test coverage for the IO channels being used from multiple domains that will exercise these code paths.

kayceesrk:
> I did wonder if we had any explicit test coverage for the IO channels being used from multiple domains that will exercise these code paths.

The new test [intext_par.ml](https://github.com/ocaml-multicore/ocaml-multicore/blob/43501299f0c997d4ab1fc557c1745d2f2b9055bb/testsuite/tests/lib-marshal/intext_par.ml) does some of it, which is how we've ended up fixing bugs in `io.c`. 

Rather than fix this issues in this PR, I shall make two separate PRs, one for reviewing `io.c` for thread-safety and another for adding test cases to cover parallel uses of `io.c`.

kayceesrk:
I've made a single issue that covers both review and adding tests: https://github.com/ocaml-multicore/ocaml-multicore/issues/618.

ctk21:
LGTM, merging

PR #613, opened 2021-07-07
Thread-safe skiplist insert and remove methods for codefrag

### LINKED ISSUE #570: Make codefrag multi-thread safe
The codefrag structures used in multicore currently do not handle multiple threads concurrently altering them. 

What is needed is to make these structures multi-thread safe.

Sudha247:
This is an attempt at making codefrag thread-safe (fixes #570) . This patch introduces two new methods in the skiplist structure for thread-safe insertion and deletion operations, codefrag uses these methods, so registering and removing code fragments from multiple threads should be safe to do. Searching remains unaltered and is lockfree, thereby can be used along with signal handlers.

### DIFF
--- a/runtime/caml/skiplist.h
+++ b/runtime/caml/skiplist.h
@@ -74,10 +74,15 @@ extern int caml_skiplist_find_below(struct skiplist * sk, uintnat k,
 extern int caml_skiplist_insert(struct skiplist * sk,
                                 uintnat key, uintnat data);
 
+extern int caml_skiplist_insert_ts(struct skiplist* sk,
+                            uintnat key, uintnat data);
+
 /* Deletion in a skip list.
    If [key] was there, remove it and return 1.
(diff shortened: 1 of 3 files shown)


ctk21:
I can see that this design uses locking for the levels of the skiplist (rather than say a lockfree skiplist). However are you able to give the argument for why the locking is correct?

Further to why the locking is correct, is it possible to have test code which exercises the new code paths?

I don't think this completely fixes #570. For example: `code_fragments_counter` in `codefrag.c` is still a non-atomic raw global. 

I'm also not sure how this PR fixes the multi-domain memory problem for the `code_fragment` elements of the data structure:
 - if one domain calls `caml_remove_code_fragment` while another domain is using that same code fragment returned by `caml_find_code_fragment_*` then the `caml_stat_free` could cause the other domain using that code fragment to segfault.

Sudha247:
Closing this one as this doesn't quite solve what we're trying to do to fix #570.

PR #616, opened 2021-07-10
Expose functions to program with effects

kayceesrk:
The aim is allow programmers to write programs that use effects without the effect syntax. Yet to port all of the `effects` tests. Here is an example of how a program using effects looks: https://github.com/ocaml-multicore/ocaml-multicore/pull/616/files#diff-876f70156c99bcbb20282b91eda75d866ebb682130a7debfc6e4bb066cc23a3eL1


kayceesrk:
Closing in favour of #648.

Issue #620, opened 2021-07-16
tcgetattr writes every value to field 0


talex5:
notty-unix's `Term.create` fails on multicore. I simplified the problem to:

```ocaml
#load "unix.cma";;
open Unix;;
let () = Printf.printf "ibaud = %d\n" (tcgetattr stdout).c_ibaud
```

On 4.12.0, this gives `ibaud = 38400`.
On 4.12.0+domains+effects, it gives `ibaud = 0`.

It looks to me like this is because the `field++` was removed in https://github.com/ocaml-multicore/ocaml-multicore/commit/30c9d5c67392685ad530182baf4fd3436cc04bf9, so everything is written to the first field.

abbysmal:
Thank you for reporting and figuring out the issue.
#621 will solve the problem.
```ocaml
utop # open Unix;;
utop # let () = Printf.printf "ibaud = %d\n" (tcgetattr stdout).c_ibaud;;
ibaud = 38400

```

ctk21:
Closing as #621 is merged and fixes the issue.

PR #621, opened 2021-07-16
otherlibs: encode_terminal_status does not set all fields

abbysmal:
This addresses issue #620

This was a small error when we moved from `caml_initialize_field` to `caml_initialize`.

I re-read the commit to check for similar errors, I could not spot any.


ctk21:
The fix looks good. Merging!

PR #623, opened 2021-07-20
Exposing caml_channel_mutex_* hooks

abbysmal:
This PR attempt implementing some missing pieces of the runtime's interface.

The idea is that, on trunk, there exists a set of four hooks that are used to handle locking channel when required.
- `caml_channel_mutex_lock`: will lock a channel's mutex (`chan->mutex`)
- `caml_channel_mutex_unlock`: will unlock the mutex
- `caml_channel_mutex_free`: will be called then the channel is being freed.
- `caml_channel_mutex_unlock_exn`: I'm not a 100% sure, but it looks like to me this one is meant to unlock the last locked channel when an exception is being raised. (see [trunk's fail_nat.c](https://github.com/ocaml/ocaml/blob/trunk/runtime/fail_nat.c#L64) and the hook set by [systhreads](https://github.com/ocaml/ocaml/blob/trunk/otherlibs/systhreads/st_stubs.c#L296))

On trunk, these hooks are by default doing nothing.
However, when systhreads is linked in, [it will setup these hooks to handle potential concurrent accesses to channels](https://github.com/ocaml/ocaml/blob/trunk/otherlibs/systhreads/st_stubs.c#L254)

On Multicore, we do not have these because channel are already being locked by default. (because many domains may try to access the same channels.)
However, we need to expose these hooks still, because some libraries may take the freedom to handle locking on channels themselves:
See for instance [core's bigstring implementation](https://github.com/janestreet/core/blob/master/bigstring_unix/src/bigstring_unix_stubs.c#L307) where IO and processing on the channel are done with custom code and as such, the hooks are called to ensure that concurrent accesses are kept in check.

This PR aims to devise what we actually need to expose for the ecosystem to be happy, and I started by implementing `mutex_lock` and `mutex_unlock`: these are pretty straightforward and I think this implementation is valid in the context of the bigstring example: we just set the hooks to `lock` or `unlock` the channel's mutex, by default.

Three pieces are missing still and may need discussion and investigation:
- `caml_channel_mutex_free`: is it used outside of the runtime code (systhreads) included, in any meaningful way besides freeing the mutex (which Multicore's runtime already does)?
- `caml_channel_mutex_unlock_exn`: My understanding is that this is used in trunk as a way to unlock the last locked channel when an exception is being raised. (see previously in this message). Is it required in our context? We do not seem to perform any channel unlocks when raising an exception.
- Linking the two previous points: do we need to restore systhreads's specifics hooks in Multicore? Most of it revolves around `last_channel_locked_key` in systhreads which is used to unlock a channel when an exception has been thrown.

### DIFF
--- a/runtime/caml/io.h
+++ b/runtime/caml/io.h
@@ -106,6 +106,12 @@ CAMLextern intnat caml_really_getblock (struct channel *, char *, intnat);
 
 #define Channel(v) (*((struct channel **) (Data_custom_val(v))))
 
+/* The locking machinery */
+
+CAMLextern void (*caml_channel_mutex_lock) (struct channel *);
+CAMLextern void (*caml_channel_mutex_unlock) (struct channel *);
+CAMLextern void (*caml_channel_mutex_unlock_exn) (void);
+
 CAMLextern struct channel * caml_all_opened_channels;
 
 /* Conversion between file_offset and int64_t */
--- a/runtime/io.c
+++ b/runtime/io.c
@@ -54,6 +54,20 @@
 #define lseek _lseeki64
 #endif
 
+static void channel_lock (struct channel *chan) {
+  caml_plat_lock(&chan->mutex);
+  return ;
+}
+
+static void channel_unlock (struct channel *chan) {
+  caml_plat_unlock(&chan->mutex);
+  return ;
+}
+
+/* Hooks for locking channels */
+CAMLexport void (*caml_channel_mutex_lock) (struct channel *) = channel_lock;
+CAMLexport void (*caml_channel_mutex_unlock) (struct channel *) = channel_unlock;
+
 /* List of opened channels and its mutex */
 CAMLexport caml_plat_mutex
   caml_all_opened_channels_mutex = CAML_PLAT_MUTEX_INITIALIZER;
--- a/runtime/io.c
+++ b/runtime/io.c
@@ -54,6 +54,11 @@
 #define lseek _lseeki64
 #endif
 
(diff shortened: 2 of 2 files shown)


kayceesrk:
@Engil Have you made any progress on the outstanding questions? Perhaps @stedolan may have something to say here as he touched this code most recently :-)

gadmm:
I reviewed this code in details at the time. Yes, you need to unlock the channel when an exception arises. The current code remembering the last channel locked might look brittle, but you need something like this since the proper way would be to return the encoded exception through the call chain, which is doable but requires a bit of work.

runtime/caml/io.h
```diff
@@ -106,6 +106,12 @@ CAMLextern intnat caml_really_getblock (struct channel *, char *, intnat);
 
 #define Channel(v) (*((struct channel **) (Data_custom_val(v))))
 
+/* The locking machinery */
+
+CAMLextern void (*caml_channel_mutex_lock) (struct channel *);
+CAMLextern void (*caml_channel_mutex_unlock) (struct channel *);
+CAMLextern void (*caml_channel_mutex_unlock_exn) (void);
```
ctk21, line 113: minor nit, but if we define this here then we should actually implement it in the `io.c` file.

ctk21 (commented):
Is the plan to merge this PR as is to help with wider eco-system issues and unlock some more testing then do a follow up PR on the remaining points?

Or should we wait for this PR to implement all the features listed before merging?

abbysmal:
This PR can be closed as all of these concern are being addressed as a part of #649

PR #626, opened 2021-07-23
Add Obj.drop_continuation

stedolan:
Programs using multi-shot handlers via `Obj.clone_continuation` can often end up with leftover continuations. Ignoring these causes a leak, and `discontinue` risks running an undesired exception handler. So this PR adds a primitive to silently drop a continuation, by freeing its stack.

(In the presence of resources, this primitive is just as unsafe as Obj.clone_continuation, which is why it lives in Obj)


kayceesrk (approved):
This is a necessary feature that is missing currently.

kayceesrk:
I'll add unit tests for these in a later PR.

PR #628, opened 2021-07-26
in link_channel, channel->prev should be set to NULL

### LINKED ISSUE #627: Strange memory leak when using `flush_all`
Using `flush_all` with `ocamlc` on a fresh version of multicore produces a strange memory leak.

The problematic code

```
let printer num 
    =
  print_int num;
  flush_all ()

let lst = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10; 11; 12; 13; 14; 15; 16; 17; 18; 19; 20]

let _ = List.iter printer (lst @ lst @ lst @ lst)
```

On my computer (macos), this produces just `12345678910111213` and then crashes (the memory consumption increases rapidly just before the crash at approx 0.3 Gb per second), running the same inside docker gets it killed due to lack of memory and produces the same output. 

Removing `flush_all` makes the problem go away. 

Strangely, this only happens with `ocamlc` eg running`ocamlc a.ml && ./a.out`, running `ocaml a.ml` produces desired output.

Multicore ocaml was installed through opam `opam switch create 4.12.0+domains --repositories=multicore=git+https://github.com/ocaml-multicore/multicore-opam.git,default`.

Reported version:
```
The OCaml runtime, version 4.12.0+domains+effects
Built with git hash '6c8abc7ab4' on branch 'master' with tag '<tag unavailable>'
```

abbysmal:
This PR attempt to fix #627.

The problem we are seeing on that issue is related to how `caml_close_channel` can re-link a channel using `link_channel`.

In any circumstances, when linking a channel, `channel->prev` should be NULL, as `link_channel` and `unlink_channels` are the one doing the accounting work on the list.

The interactions between `unlink_channel` and `link_channel` and where they are called may benefit from further proofreading.

### DIFF
--- a/runtime/io.c
+++ b/runtime/io.c
@@ -80,6 +80,7 @@ static void link_channel (struct channel* channel)
 {
   caml_plat_lock (&caml_all_opened_channels_mutex);
   channel->next = caml_all_opened_channels;
+  channel->prev = NULL;
   if (caml_all_opened_channels != NULL)
     caml_all_opened_channels->prev = channel;
   caml_all_opened_channels = channel;
--- a/runtime/io.c
+++ b/runtime/io.c
@@ -80,7 +80,7 @@ static void link_channel (struct channel* channel)
 {
   caml_plat_lock (&caml_all_opened_channels_mutex);
   channel->next = caml_all_opened_channels;
-  channel->prev = NULL;
+  CAMLassert(channel->prev == NULL);
   if (caml_all_opened_channels != NULL)
     caml_all_opened_channels->prev = channel;
   caml_all_opened_channels = channel;
@@ -99,6 +99,8 @@ static void unlink_channel(struct channel *channel)
     channel->prev->next = channel->next;
     if (channel->next != NULL) channel->next->prev = channel->prev;
   }
+  channel->next = NULL;
+  channel->prev = NULL;
   caml_plat_unlock (&caml_all_opened_channels_mutex);
 }
 


ctk21:
I think I've convinced myself that the fix is correct; otherwise we end up with `channel->prev` pointing somewhere odd, when it should be the first element of the list after `link_channel`.

kayceesrk:
Looks like I'd introduced this bug when I changed the logic around closing channels here: https://github.com/ocaml-multicore/ocaml-multicore/pull/628/files#diff-5455bfdb354e5d8ad156cf2975ce29c57a622dffed9ad3b7e541f81bb715d3e5R141-R144. In order to handle concurrent access, the new logic first unlinks the channel and then links it back again if necessary. Old code only unlinks the channel when necessary. 

Can I recommend a different way of fixing this which is a little bit cleaner? I'd recommend:

1. Setting the `prev` and `next` pointer of the channel to `NULL` when the channel is unlinked here: https://github.com/ocaml-multicore/ocaml-multicore/pull/628/files#diff-5455bfdb354e5d8ad156cf2975ce29c57a622dffed9ad3b7e541f81bb715d3e5R102
2. Replacing the newly inserted line (https://github.com/ocaml-multicore/ocaml-multicore/pull/628/files#diff-5455bfdb354e5d8ad156cf2975ce29c57a622dffed9ad3b7e541f81bb715d3e5R83) to an assertion `CAMLassert (channel->prev == NULL)`.

abbysmal:
Thanks for your review @ctk21, @kayceesrk 
I just pushed some changes, accordingly to your suggestions.

kayceesrk (approved):

kayceesrk:
LGTM.

Issue #624, opened 2021-07-21
core v0.14: test triggers a segfault in the GC


emillon:
Hi!

Hi had a go at running the test suite for core v0.14 under `4.12.0+domains`.
This is a bit involved because the tests are stripped from the public releases. I'm documenting that primarily so that I don't forget how to do it...

I first assembled core (from the tarball) into a monorepo with its dependencies. The general workflow is:

- untar the latest core release
- create a vanilla 4.12.0 switch
- `opam monorepo lock` to create a lock file
- `opam monorepo pull` to assemble a duniverse directory
- to fix tests, the main cycle is
  + `dune runtest`
  + edit `core.opam` to add dependencies `{with-test}`
  + re-lock
  + re-pull

I had to extract `quickcheck_deprecated` from an older `core_extended` release and move it by hand to a subdirectory.
Next step is to enable inline tests: libraries that contain `let%expect_test` bits (alternatively, whose `dune` file refer to `ppx_expect` or `ppx_jane`) need to have an extra `(inline_tests)` field. I applied the above loop to pull new dependencies (like async). Some warnings appear (the new 4.12 ones), they can be silenced with an `(env)` stanza. Some tests needed to be stubbed out but generally `dune runtest` works. I had to manually clone some dependencies because these depend on `core` itself. I believe this is just a `opam-monorepo` limitation. These dependencies are: `async`,  `async_kernel`, `async_rpc_kernel`, `async_unix`, `expect_test_helpers_async`.

At that point, dune runtest works. The switch can be replaced by a `+domains switch. The main issues are:

- there is a new deprecation warning because core reexports `Thread.wait_read` and `Thread.wait_write` - it can be marked as nonfatal in the dune file
- some stubs refer to `caml_channel_mutex_lock`: https://github.com/janestreet/core/blob/v0.14/bigstring_unix/src/bigstring_unix_stubs.c#L269 - related to #623 (patch makes it pass)
- a test segfaults:

```
% dune runtest
inline_test_runner_core_test alias test/src/runtest (got signal SEGV)
(cd _build/default/test/src && .core_test.inline-tests/inline_test_runner_core_test.exe inline-test-runner core_test -source-tree-root ../.. -diff-cmd -)
# under gdb
(gdb) bt
#0  caml_darken (v=0, ignored=<optimized out>, state=<optimized out>) at major_gc.c:760
[...]

PR #629, opened 2021-07-28
Backtrace last exn is val unit

abbysmal:
This PR attempts to correct an issue found when running `core's` testsuite, as reported by @emillon in #624.

#### Excerpt from the issue:

What happens is that core links `systhreads` in, so the runtime will go through the process of registering `backtrace_last_exn` as a generational global root.
https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/runtime/backtrace.c#L44

This should be fine for any valid values for `backtrace_last_exn`, except there's a codepath that sets it to `NULL` here:
https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/runtime/backtrace.c#L44

Which means that later on, when the GC kicks in and this global root was registered with `NULL`, `caml_darken` will crash the runtime because it is an invalid value.

#### Proposed solution

After discussing with @ctk21, it seems that the second branch in `caml_record_backtrace` is a bit odd, as the generational global root may still be accessed elsewhere in the codebase after removal. (as demonstrated by this issue.)

A simple fix would be to remove this branch and just clear `backtrace_last_exn` to `Val_unit` instead everytime the flag is changed, and not remove the generational global root like it is done right now.

### DIFF
--- a/runtime/backtrace.c
+++ b/runtime/backtrace.c
@@ -41,7 +41,7 @@ CAMLprim value caml_record_backtrace(value vflag)
       caml_modify_generational_global_root(&Caml_state->backtrace_last_exn, Val_unit);
     } else {
       caml_remove_generational_global_root(&Caml_state->backtrace_last_exn);
-      Caml_state->backtrace_last_exn = (value) NULL;
+      Caml_state->backtrace_last_exn = (value) Val_unit;
     }
   }
   return Val_unit;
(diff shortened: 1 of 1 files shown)


kayceesrk:
I assume that the fact that we're leaving the generational root is not going to be a problem? Otherwise, this looks good to me.

kayceesrk (approved):

abbysmal:
@kayceesrk Looking at the code, I think it won't cause any harm, it's actually much saner to not touch it too much if not required.
Only reason I see to not let it be would be to save one extra generational global root (if backtraces are disabled explicitly through that codepath), which I don't think if worth the accounting that should be added.

kayceesrk:
Ok. LGTM. Merging.

PR #631, opened 2021-07-29
Don't raise asynchronous exceptions from signals in caml_alloc C functions

sadiqj:
This PR avoids polling for pending signals from `caml_alloc_*` calls that originate from C. This prevents asynchronous exceptions that are raised from signal handlers.

This mirrors the work on trunk in https://github.com/ocaml/ocaml/pull/8691 though some of our allocation code has diverged and we don't have statmemprof currently - so we couldn't simply apply those changes.

It would be super useful if in review people could check that I've not missed any calls from C that could trigger a GC which is not done in a way that prevents asynchronous exceptions.

It's worth pointing out that this doesn't handle the asynchronous exceptions that are raised from finalisers.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -1079,7 +1079,7 @@ void caml_handle_gc_interrupt()
 {
   handle_gc_interrupt();
 
-  caml_raise_if_exception(caml_process_pending_signals_exn());
+  caml_process_pending_signals();
 }
 
 CAMLexport int caml_bt_is_in_blocking_section(void)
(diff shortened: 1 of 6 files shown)


runtime/domain.c
```diff
@@ -991,6 +991,11 @@ int caml_try_run_on_all_domains(void (*handler)(caml_domain_state*, void*, int,
   return caml_try_run_on_all_domains_with_spin_work(handler, data, leader_setup, 0, 0);
 }
 
+int caml_check_pending_actions() {
+  return (((uintnat)Caml_state->young_ptr - Bhsize_wosize(Max_young_wosize) <
```
kayceesrk, line 995: I don't understand this particular check. Can you explain how this check works?

sadiqj: This function is actually left over from the EINTR combined PR. I've removed it for now.

(It's checking specifically whether the minor heap is full-ish)

kayceesrk:
Looks good to me. Merging.

PR #635, opened 2021-09-01
Make lib-str domain safe

abbysmal:
This PR moves the usage of global variables in `str` to thread local storage.

The one weird part revolves around freeing the resources on domain shutdown. It make use of the previously introduced `domain_stop_hook` (for systhreads).

There's also the alternative route of allocating str's internal data structures as custom blocks and letting the GC take care of these, though I'm not quite sure it is a cleaner solution.

I think adding a testcase that does some `str` computations in parallel would be useful, I will add some in the coming days.


abbysmal:
I also noticed there's one case I'm not handling properly, will mark it as draft until I fixed it.

Sudha247 (commented):
I went through this PR and had a discussion with @Engil offline, the gist of it is below:

Thread-local values are stored in pthread specific variable heres. This means the thread-local values will be unique not just for domains, but also for systhreads. 

To be noted: systhreads access the same global state on trunk, if we'd like to replicate this behaviour I think it might be useful to have a C API for domain-local storage. It appears `Str` module is currently not thread-safe in the context of systhreads and this PR would actually fixes it. We think it's all right to move forward with this approach, unless there are any alternate ideas.

otherlibs/str/strstubs.c
```diff
@@ -89,10 +94,81 @@ struct re_group {
...
+  }
+
+  reg = get_re_register();
+
+  if (reg != NULL) {
+    caml_stat_free(reg);
+    pthread_setspecific(re_register_key, NULL);
+  }
+
+  prev_domain_stop_hook();
```
ctk21, line 157: I see `prev_domain_stop_hook` getting called, but I don't see it being assigned to. 
Wouldn't we need to initialize this in `caml_str_initialize`

otherlibs/str/strstubs.c
```diff
@@ -89,10 +94,81 @@ struct re_group {
...
+}
+
+static unsigned char **get_re_register()
+{
+  unsigned char **re_register = pthread_getspecific(re_register_key);
 
-/* The initial backtracking stack */
-static struct backtrack_stack initial_stack = { NULL, };
+  if (re_register == NULL) {
+    re_register =  caml_stat_alloc(sizeof(unsigned char *) * NUM_REGISTERS);
```
ctk21, line 123: Should we add error handling if the allocation fails?

ctk21 (commented):
I think this works and solves the problem of multiple domains using str. 

I had a nit about error handling should an alloc fail. Also I'm not convinced we store `prev_domain_stop_hook` which is a bug.

otherlibs/str/strstubs.c
ctk21: it's been pointed out that the error handling is at least how it is for the rest of the file. 
we can leave it and merge this PR as is.

xavierleroy:
I wonder if we could avoid global state in strstubs.c altogether, using C local variables for the initial RE engine stack and for the register array.  I'll give it a try in the sequential version of OCaml.

xavierleroy:
Are you OK to use https://github.com/ocaml/ocaml/pull/10670 instead?  Much less code.

ctk21:
> Are you OK to use [ocaml/ocaml#10670](https://github.com/ocaml/ocaml/pull/10670) instead? Much less code.

The approach taken in [ocaml/ocaml#10670](https://github.com/ocaml/ocaml/pull/10670) looks good to me and is less code. 

When that gets merged upstream, we will pick it up naturally as we rebase.

abbysmal:
It is much cleaner than the current approach, and removes the need to add more cleanup hooks in the domain's lifecycle, thank you for this. :)

xavierleroy:
Thanks for the confirmation.  In the future I'll try to make suggestions for alternate implementations earlier!

Issue #637, opened 2021-09-02
caml_page_table_lookup is not available in ocaml-multicore


kit-ty-kate:
`ancien` uses the `Is_in_heap_or_young` macro which uses `caml_page_table_lookup` and is defined but not implemented in ocaml-multicore.
```
#=== ERROR while compiling ancient.0.9.1 ======================================#
# context              2.1.0 | linux/x86_64 | ocaml-variants.4.12.0+domains+channel_mutex | file:///home/opam/opam-repository
# path                 ~/.opam/4.12.0+domains+channel_mutex/.opam-switch/build/ancient.0.9.1
# command              ~/.opam/opam-init/hooks/sandbox.sh build make
# exit-code            2
# env-file             ~/.opam/log/ancient-19-db1b31.env
# output-file          ~/.opam/log/ancient-19-db1b31.out
### output ###
# ocamlfind ocamlopt   -package unix -c test_ancient_dict_write.ml
# LIBRARY_PATH=.:$LIBRARY_PATH \
# ocamlfind ocamlopt  -package unix -linkpkg -o test_ancient_dict_write.opt ancient.cmxa test_ancient_dict_write.cmx
# /usr/bin/ld: ./libancient.a(ancient_c.o): in function `_mark':
# /home/opam/.opam/4.12.0+domains+channel_mutex/.opam-switch/build/ancient.0.9.1/ancient_c.c:206: undefined reference to `caml_page_table_lookup'
# /usr/bin/ld: /home/opam/.opam/4.12.0+domains+channel_mutex/.opam-switch/build/ancient.0.9.1/ancient_c.c:238: undefined reference to `caml_page_table_lookup'
# /usr/bin/ld: ./libancient.a(ancient_c.o): in function `ancient_delete':
# /home/opam/.opam/4.12.0+domains+channel_mutex/.opam-switch/build/ancient.0.9.1/ancient_c.c:418: undefined reference to `caml_page_table_lookup'
# /usr/bin/ld: ./libancient.a(ancient_c.o): in function `ancient_is_ancient':
# /home/opam/.opam/4.12.0+domains+channel_mutex/.opam-switch/build/ancient.0.9.1/ancient_c.c:434: undefined reference to `caml_page_table_lookup'
# collect2: error: ld returned 1 exit status
# File "caml_startup", line 1:
# Error: Error during linking (exit code 1)
# make: *** [Makefile:35: test_ancient_dict_write.opt] Error 2
```
[...]

PR #642, opened 2021-09-06
Remove the remanents of page table functionality.

kayceesrk:
Multicore OCaml does not use page tables, and the macro and function definitions are no longer used. Related to https://github.com/ocaml-multicore/ocaml-multicore/issues/637.

### DIFF
--- a/runtime/caml/address_class.h
+++ b/runtime/caml/address_class.h
@@ -15,44 +15,25 @@
 
 /* Classification of addresses for GC and runtime purposes. */
 
-/* The current runtime supports two different configurations that
-   correspond to two different value models, depending on whether
-   "naked pointers", that do not point to a well-formed OCaml block,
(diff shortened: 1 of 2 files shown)


ctk21:
I think if we are cleaning up `address_class.h` we should go further and tidy up the comment:
```C
/* The current runtime supports two different configurations that
   correspond to two different value models, depending on whether
   "naked pointers", that do not point to a well-formed OCaml block,
   are allowed (considered valid values).
   
   ...
```

We should state that in multicore we are always in stock OCaml's "no naked pointers" mode. We then just need the macros to work as expected for compatibility.

kayceesrk:
I've rewritten the comments and only retained definitions that are relevant in Multicore OCaml.

ctk21:
Thanks, LGTM, merging!

PR #649, opened 2021-09-15
Integrate all of trunk's EINTR fixes

sadiqj:
This PR incorporates nearly all of the changes from @stedolan 's https://github.com/ocaml/ocaml/pull/9722 . It should significantly reduce our diff with trunk on files such as `io.c`. It also restores the mutex lock/unlock hooks that have caused some compatibility issues (https://github.com/ocaml-multicore/ocaml-multicore/pull/623). 

After https://github.com/ocaml-multicore/ocaml-multicore/pull/630 and https://github.com/ocaml-multicore/ocaml-multicore/pull/631 this should be the last signals PR (though we still need to deal with asynchronous exceptions from finalisers).

One design question for @Engil to consider (for potentially a follow-up PR) - do we still need the systhreads implementations of the hooks given how similar they are to the default ones in `io.c` now?

### DIFF
--- a/runtime/io.c
+++ b/runtime/io.c
@@ -57,6 +57,48 @@
 /* List of opened channels and its mutex */
 CAMLexport caml_plat_mutex
   caml_all_opened_channels_mutex = CAML_PLAT_MUTEX_INITIALIZER;
+
+/* Hooks for locking channels */
+
+static __thread struct channel* last_channel_locked = NULL;
+
+static void channel_mutex_free_default(struct channel *chan)
+{
+  caml_plat_mutex_free(&chan->mutex);
+}
+
+static void channel_mutex_lock_default(struct channel *chan)
+{
+  if( caml_plat_try_lock(&chan->mutex) ) {
+    last_channel_locked = chan;
+    return;
+  }
+
+  /* If unsuccessful, block on mutex */
+  caml_enter_blocking_section();
+  caml_plat_lock(&chan->mutex);
+  last_channel_locked = chan;
+  caml_leave_blocking_section();
+}
+
+static void channel_mutex_unlock_default(struct channel *chan)
+{
+  caml_plat_unlock(&chan->mutex);
+  last_channel_locked = NULL;
+}
+
+static void channel_mutex_unlock_exn_default(void)
+{
+  struct channel * chan = last_channel_locked;
+  if (chan != NULL && caml_channel_mutex_unlock != NULL) caml_channel_mutex_unlock(chan);
+}
+
+CAMLexport void (*caml_channel_mutex_free) (struct channel *) = channel_mutex_free_default;
+CAMLexport void (*caml_channel_mutex_lock) (struct channel *) = channel_mutex_lock_default;
+CAMLexport void (*caml_channel_mutex_unlock) (struct channel *) = channel_mutex_unlock_default;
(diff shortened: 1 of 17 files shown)


runtime/io.c
```diff
@@ -281,22 +339,33 @@ CAMLexport file_offset caml_pos_out(struct channel *channel)
...
-  if (n == 0) caml_raise_end_of_file();
+  if (n == Io_interrupted) goto again;
+  else if (n == 0) caml_raise_end_of_file();
   channel->offset += n;
   channel->max = channel->buff + n;
   channel->curr = channel->buff + 1;
   return (unsigned char)(channel->buff[0]);
 }
 
+#define Getch(channel)                                                      \
```
abbysmal, line 364: Isn't this functionally equivalent to the currently defined `caml_getch` in `io.h`?

runtime/io.c
```diff
@@ -213,6 +271,11 @@ CAMLexport void caml_flush(struct channel *channel)
 
 /* Output data */
 
+#define Putch(channel, ch) do{                                            \
```
abbysmal, line 274: Same as for `Getch`, isn't it the same as `caml_putch` defined in `io.h`?

runtime/io.c
```diff
@@ -429,36 +502,7 @@ void caml_finalize_channel(value vchan)
     link_channel (chan);
     return;
   }
-  caml_plat_mutex_free(&chan->mutex);
-  /* TODO KC: See commented out section */
-#if 0
   if (caml_channel_mutex_free != NULL) (*caml_channel_mutex_free)(chan);
-
```
abbysmal, line 436: Is there a reason for the removal of this block?

runtime/signals.c
```diff
@@ -139,6 +142,7 @@ CAMLexport void (*caml_enter_blocking_section_hook)(void) =
 CAMLexport void (*caml_leave_blocking_section_hook)(void) =
    caml_leave_blocking_section_default;
 
+CAMLno_tsan /* The read of [caml_something_to_do] is not synchronized. */
```
abbysmal: I'm not sure about this line?
At least the comment is referring to trunk rather than Multicore.

abbysmal (commented):
I reviewed the changes, posted a few comments (nothing major), and took a look at trunk in the process.
Everything make sense and I think the changes are good.

For systhreads I think we can indeed get rid of the systhreads specific hooks: they operate in the same way, the `last_exception_key` is stored in a thread-local variable which means that each executor will behave pretty much the same.
This cleanup can be left for a follow up PR I think. (I can take care of it myself as well.)

runtime/io.c
sadiqj: Yes. It was the same on trunk (https://github.com/ocaml/ocaml/blob/4.12.0/runtime/caml/io.h#L84 and https://github.com/ocaml/ocaml/blob/trunk/runtime/io.c#L309) and will get fixed on the rebase: https://github.com/ocaml/ocaml/commit/f7e32a3b4852d65772dc75af72956ee379a2784e#diff-cc528a0b067be28c2081c889ec38bcc70ccc0c42d2e70d3de7b2b4967c0ddd10L82

runtime/signals.c
sadiqj: Yes, this is left over from trunk and doesn't make sense here. I'll remove.

runtime/io.c
sadiqj: Actually, I think it could be kept. Let me see.

PR #650, opened 2021-09-17
Add primitives necessary for exposing effect handlers as functions

kayceesrk:
The idea is to upstream these primitives so that our [current procedure](https://github.com/ocaml-multicore/ocaml-multicore/wiki/Instructions-for-rebasing-to-remove-the-effects-syntax) for updating `4.12+domains` with latest changes from `4.12+domains+effects` can continue to work. 

The PR#651 exposes effect handlers as functions.


stdlib/obj.mli
```diff
@@ -199,3 +199,12 @@ module Ephemeron: sig
   (** Maximum length of an ephemeron, ie the maximum number of keys an
       ephemeron could contain *)
 end
+
+module Effect_handlers : sig
```
ctk21: Why do we need the `Effect_handlers` module in this PR rather than in #651 ?

kayceesrk: I thought it was needed, but it turns out it is not. Fixed.

PR #651, opened 2021-09-17
Expose deep and shallow handlers as functions

kayceesrk:
This function exposes deep and shallow handlers as function from the obj module. In order to make it easier to keep the task of updating 4.12+domains with bug fixes, the primitives that have been added have also been proposed to be added to 4.12+domains+effects in #650. When #650 and this PR is merged, we can continue to use the [current instructions](https://github.com/ocaml-multicore/ocaml-multicore/wiki/Instructions-for-rebasing-to-remove-the-effects-syntax) to keep 4.12+domains updated with any changes to 4.12+domains+effects. 

This PR also removes the ability to clone continuations. Cloning continuation breaks compiler optimisations and is hard to use correctly in the presence of resources. See 

1. A separation logic for effect handlers, https://dl.acm.org/doi/10.1145/3434314.  
2. An example that shows continuation cloning breaking compiler optimisation, https://gist.github.com/kayceesrk/c796b826a481619c0a13e90062bedfdd


stdlib/printexc.mli
```diff
@@ -190,6 +190,22 @@ external get_callstack: int -> raw_backtrace = "caml_get_current_callstack"
     @since 4.01.0
 *)
 
+external get_kcallstack_deep :
```
ctk21, line 193: Are we sure we want these functions in `Printexc` or could they go into `Effect_handlers` in obj?
Just wondering if everything should be in obj while the API has a 'experimental' flavour. 

I might be missing a reason why this one needs to be here.

runtime/fiber.c
```diff
@@ -515,6 +478,23 @@ CAMLprim value caml_continuation_use (value cont)
   return v;
 }
 
+CAMLprim value caml_clone_continuation (value cont)
+{
+  caml_fatal_error ("caml_clone_continuation is unsupported");
+  return Val_unit;
+}
+
+CAMLprim value caml_update_handler (value stack, value hval, value hexn,
```
ctk21: This function worried me from a multi-thread safety angle; if two domains tried to update the handler at the same time then things would go bad.
 
It is actually in this PR because `caml_update_handler` is only ever called like this:
```ocaml
      let stack = take_cont_noexc k in
      let stack' = update_handler stack handler.retc handler.exnc effc in
```
Might it be better to have a `take_cont_and_update_handler` primitive that ensures that it wraps up the continuation take as well?
Alternatively maybe a comment to suggest that you should never call `caml_update_handler` without first ensuring that you own the stack.

ctk21 (commented):
I think this one works.

stdlib/printexc.mli
kayceesrk: The function returns a `raw_backtrace` which is defined in `Printexc`, but `Obj` comes before `Printexc` in the dependency chain (see `stdlib/StdlibModules`). Hence I've left the functions in `Printexc`.

runtime/fiber.c
kayceesrk: Combined them into a single primitive.

stdlib/printexc.mli
ctk21: ah, I see. Fair enough

stdlib/obj.ml
```diff
@@ -196,3 +196,128 @@ module Ephemeron = struct
...
+    let try_with comp arg handler =
+      let effc' eff k last_fiber =
+        match handler.effc eff with
+        | Some f -> f k
+        | None -> reperform eff k last_fiber
+      in
+      let s = alloc_stack (fun x -> x) (fun e -> raise e) effc' in
+      runstack s comp arg
+
+    external drop_continuation : ('a,'b) continuation -> unit =
```
dhil, line 258: What is the rationale for having `drop_continuation` in addition to `discontinue`? In case there is a good reason to keep it, I have a follow-up question: is it necessary for `drop_continuation` to be a primitive? I'd assume it is is definable in terms of `discontinue`:

```ocaml
let drop_continuation k = 
  let exception Drop in
  try discontinue k Drop with Drop -> ()
```

That is, unless `drop_continuation` doesn't perform any clean up of resources in the stack represented by the continuation.

stdlib/obj.ml
```diff
@@ -196,3 +196,128 @@ module Ephemeron = struct
...
+      ('a,'b) continuation ->
+      ('b -> 'c) ->
+      (exn -> 'c) ->
+      ('d eff -> ('d,'b) continuation -> last_fiber -> 'c) ->
+      ('a,'c) stack = "caml_continuation_use_and_update_handler_noexc" [@@noalloc]
+
+    external reperform :
+      'a eff -> ('a, 'b) continuation -> last_fiber -> 'c = "%reperform"
+
+    let continue_with k v handler =
```
dhil, line 302: For what it is worth, this isn't exactly shallow handlers as they are presented in the literature. As by this interface you are obliged to install a handler upon resume, which has the nice property that when a subsequent effect operation is invoked there will be a handler in place. However, genuine shallow continuations need to not be run under a new handler, rather they can run in the ambient environment as is. Obviously, one can "simulate" this behaviour by installing the identity handler, though, such a simulation is not space-efficient.

dhil: (This interface is the same as the interface we have in the "typed continuations" proposal for Wasm. I don't think it has a proper name in the literature yet.)

stdlib/obj.ml
kayceesrk: I've responded in #653.

stdlib/obj.ml
kayceesrk: I agree that this is not the standard presentation in literature but matches the typed continuations proposal for Wasm. I think the reasoning towards going for this interface is exactly what you have described, which is that there is always a handler guaranteed to be in scope.

PR #652, opened 2021-09-20
Make young_start/end/ptr pointers to value

### LINKED ISSUE #643: Core_kernel minor words report are off
@emillon found out that the core testsuite still fails on Multicore because the report of allocated words are off.

A simple program showcase this behavior:
```ocaml
open Core_kernel

let () =
  let (), report =
    Gc.For_testing.measure_allocation (fun () -> ignore (List.init 1000 ~f:Fn.id))
  in
  (* Vanilla: 3005 0 *)
  (* Multicore: 24040 0 *)
  Printf.printf "%d %d\n" report.minor_words_allocated report.major_words_allocated
```
The reason for this behavior (allocations are multiplied by 8) is that young_ptr and young_end are char * in [Multicore](https://github.com/ocaml-multicore/ocaml-multicore/blob/4.12%2Bdomains%2Beffects/runtime/caml/domain_state.tbl#L22)
In trunk, it's [value *](https://github.com/ocaml/ocaml/blob/trunk/runtime/caml/domain_state.tbl#L18)
Leading to the skewed computation here: https://github.com/janestreet/core_kernel/blob/master/core/src/gc_stubs.c#L10

The fix would be to change `young_ptr` and `young_end` to align to trunk's, and to make sure every subsequent arithmetic with these do line up properly in the runtime code.

Thanks again @emillon for finding this out.

abbysmal:
This PR attempts to fix #643 

In Multicore, `young_end`, `young_start` and `young_ptr` are `char *`, rather than `value *` as defined in `trunk`.

This can cause issues with external code relying on that fact to do any kind of pointer arithmetic.

This PR aligns OCaml Multicore to trunk.

### DIFF
--- a/runtime/caml/memory.h
+++ b/runtime/caml/memory.h
@@ -213,11 +213,11 @@ extern uintnat caml_use_huge_pages;
                                           CAMLassert ((tag_t) (tag) < 256); \
                                  CAMLassert ((wosize) <= Max_young_wosize); \
   dom_st = Caml_state;                                                      \
-  Move_young_ptr(dom_st, -Bhsize_wosize(wosize));			    \
-  while (Caml_check_gc_interrupt(dom_st)) {				    \
-    Move_young_ptr(dom_st, +Bhsize_wosize(wosize));  			    \
+  Move_young_ptr(dom_st, -Bhsize_wosize(wosize));                           \
+  while (Caml_check_gc_interrupt(dom_st)) {                                 \
+    Move_young_ptr(dom_st, +Bhsize_wosize(wosize));                         \
     { GC }                                                                  \
-    Move_young_ptr(dom_st, -Bhsize_wosize(wosize));  			    \
+    Move_young_ptr(dom_st, -Bhsize_wosize(wosize));                         \
   }                                                                         \
   Hd_hp (dom_st->young_ptr) =                                               \
     Make_header_with_profinfo ((wosize), (tag), 0, profinfo);               \
--- a/runtime/caml/memory.h
+++ b/runtime/caml/memory.h
@@ -205,17 +205,19 @@ extern uintnat caml_use_huge_pages;
 #define DEBUG_clear(result, wosize)
 #endif
 
+#define Move_young_ptr(d, to) d->young_ptr = (value*)((uintnat)d->young_ptr to);
+
 #define Alloc_small_with_profinfo(result, wosize, tag, GC, profinfo) do{    \
   caml_domain_state* dom_st;                                                \
                                                 CAMLassert ((wosize) >= 1); \
                                           CAMLassert ((tag_t) (tag) < 256); \
                                  CAMLassert ((wosize) <= Max_young_wosize); \
   dom_st = Caml_state;                                                      \
-  dom_st->young_ptr -= Bhsize_wosize (wosize);                              \
-  while (Caml_check_gc_interrupt(dom_st)) {                                 \
-    dom_st->young_ptr += Bhsize_wosize (wosize);                            \
+  Move_young_ptr(dom_st, -Bhsize_wosize(wosize));			    \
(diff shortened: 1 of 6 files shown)


abbysmal:
It seems the testsuite is not quite passing yet, I'll mark it as draft for now.

abbysmal:
I confirm that this fixes the stat issue with `core_kernel` at #643 with the testcase provided by @emillon.

I think the PR is now ready for review.

runtime/caml/memory.h
```diff
@@ -205,17 +205,19 @@ extern uintnat caml_use_huge_pages;
 #define DEBUG_clear(result, wosize)
 #endif
 
+#define Move_young_ptr(d, to) d->young_ptr = (value*)((uintnat)d->young_ptr to);
```
ctk21: I'm not a fan of this macro as it adds an indirection to understand what is going on. I think we should just do the arithmetic in words (as in upstream).
See suggestions below.

runtime/caml/memory.h
```diff
@@ -205,17 +205,19 @@ extern uintnat caml_use_huge_pages;
...
+
 #define Alloc_small_with_profinfo(result, wosize, tag, GC, profinfo) do{    \
   caml_domain_state* dom_st;                                                \
                                                 CAMLassert ((wosize) >= 1); \
                                           CAMLassert ((tag_t) (tag) < 256); \
                                  CAMLassert ((wosize) <= Max_young_wosize); \
   dom_st = Caml_state;                                                      \
-  dom_st->young_ptr -= Bhsize_wosize (wosize);                              \
+  Move_young_ptr(dom_st, -Bhsize_wosize(wosize));                           \
```
ctk21: Isn't this just 
```suggestion
  dom_st->young_ptr -= Whsize_wosize(wosize);                             \
```

runtime/caml/memory.h
```diff
@@ -205,17 +205,19 @@ extern uintnat caml_use_huge_pages;
...
   while (Caml_check_gc_interrupt(dom_st)) {                                 \
-    dom_st->young_ptr += Bhsize_wosize (wosize);                            \
+    Move_young_ptr(dom_st, +Bhsize_wosize(wosize));                         \
```
ctk21:
```suggestion
    dom_st->young_ptr += Whsize_wosize (wosize);                         \
```

runtime/caml/memory.h
```diff
@@ -205,17 +205,19 @@ extern uintnat caml_use_huge_pages;
...
     { GC }                                                                  \
-    dom_st->young_ptr -= Bhsize_wosize (wosize);                            \
+    Move_young_ptr(dom_st, -Bhsize_wosize(wosize));                         \
```
ctk21:
```suggestion
   dom_st->young_ptr -= Whsize_wosize (wosize);                         \
```

runtime/signals_nat.c
```diff
@@ -108,7 +108,8 @@ void caml_garbage_collection()
     alloc_bsize = whsize * sizeof(value);
 
     /* Put the young pointer back to what is was before our tiggering allocation */
-    Caml_state->young_ptr += alloc_bsize;
+    Caml_state->young_ptr = (value *)((uintnat)Caml_state->young_ptr
```
ctk21: I'm wondering if in this segment here, we do away with `alloc_bsize`
Then just do pointer arithmetic with `whsize` on the raw `value` pointers in `young_ptr` and `young_limit`. 
That should cut the amount of casting things around.

ctk21 (commented):
I've checked this and I believe it correct. 
However there are a couple of places where I think we might be able to improve the clarity and cut the amount of casting.

runtime/signals_nat.c
abbysmal: Right, good point.
I pushed the change just now, thank you for the suggestion!

PR #653, opened 2021-09-27
Drop `drop_continuation`

dhil:
This PR is largely a response to my own question on PR#651 (https://github.com/ocaml-multicore/ocaml-multicore/pull/651) regarding whether `drop_continuation` is redundant since `clone_continuation` has been removed. Since the removal of `clone_continuation` we can no longer have a "dangling" extra continuation, which needs to be discarded. The primitives `clone_continuation` and `drop_continuation` go hand-in-hand and it makes little sense to have one without the other. Hence, this PR removes `drop_continuation`.


kayceesrk:
Isn't `drop_continuation` still useful if you know that the code does not close over resources, and you'd just want to free the stack? This would be faster using `discontinue` to unwind the stack. But this is a very minor argument in favour, and I'd prefer to remove `drop_continuation`. 

Moreover, the current implementation of `drop_continuation` is not correct. It leaks fiber memory if the current fiber points to a parent. Only the top fiber's memory is freed; it's parent fiber memory is not.

kayceesrk (approved):

dhil:
> Isn't `drop_continuation` still useful if you know that the code does not close over resources, and you'd just want to free the stack? This would be faster using `discontinue` to unwind the stack. But this is a very minor argument in favour, and I'd prefer to remove `drop_continuation`.

Sure, I accept that argument. I think this use-case is rare in a single one-shot setting, whereas in a multi-shot setting with explicit copying it arises more frequently, e.g. when you do backtracking.

dhil:
I suppose this patch is superseded by #689.

kayceesrk:
Yep. Let's close this one.

PR #654, opened 2021-09-30
Enable effects tests

kayceesrk:
Looks like the effect handler tests remained disabled in after the syntax support was added. This PR removes the disabled effects tests.


ctk21:
I took the liberty of implementing this by fixing up the 'expose deep and shallow handlers as functions' commit on `4.12+domains`.

PR #656, opened 2021-09-30
Core testsuite workflow

abbysmal:
This PR aims to implement a workflow to run Core's testsuite once a day.

The long run ideal would be to include in a matrix other pieces of Core's ecosystem (`async_unix`, `core_kernel`, to name a few) and to keep tab on those from within our CI environment, since running their testsuite is rather hard.

This follows @emillon's approach in #624 (thanks!), and the monorepo is a fork of his, hosted at `ocaml-multicore/core`.

This PR and monorepo will need updates once our rebase to 4.13 is done, and I will try to dig further into integrated more libraries in this run, when possible.

As such I will mark it as a draft for now.

Note that in Core's case, I included a fix for the `channel` layout assumption in bigstring, since this will be a required upstream change in the long run.

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -1,0 +1,32 @@
+name: core
+
+on:
+  push:
+  schedule:
+    - cron: '1 0 * * *'
+
+jobs:
+
+  build:
+    name: 'linux'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          repository: 'ocaml-multicore/core'
+          ref: 'monorepo-0.14'
+
+      - name: OCaml
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: 'ocaml-variants.4.12.0+domains' 
+          opam-depext: false
+          opam-repositories: |
+            default: https://github.com/ocaml/opam-repository.git
+
+      - run: opam monorepo lock
+
+      - run: opam monorepo pull
+
+      - run: opam exec -- dune runtest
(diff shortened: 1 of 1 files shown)


kayceesrk:
Which is the branch used in `ocaml-multicore/core`? Is is `monorepo-0.14`? How does the workflow file pick this branch? Should we make `monorepo-0.14` the default branch on `ocaml-multicore/core` because that's the relevant one.

abbysmal:
I think we can make the default branch monorepo-0.14 indeed, though the workflow files picks it up as a part of the checkout actions, like this:
```
        with:
          repository: 'ocaml-multicore/core'
          ref: 'monorepo-0.14'
```

.github/workflows/core.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: OCaml
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: 'ocaml-variants.4.12.0+domains' 
+          opam-depext: false
+          opam-repositories: |
+            default: https://github.com/ocaml/opam-repository.git
+
+      - run: opam monorepo lock
```
emillon: If the repository contains `core.opam.locked`, it's better to omit this line, that way you'll get reproducible results. And you can update the dependencies by hand by pushing a new lockfile (obtained by locally running `opam monorepo lock`).

abbysmal: Right, anyway I noticed it fails nowadays and cannot recreate a lock file (because of some `ocaml-base-compiler` conflict at some point, I'm not sure I understand it). So it leaves the previous lockfile in place, apparently.
I will remove that line, thanks.

PR #657, opened 2021-10-01
libstr: use a domain local value to store last_search_result_key

abbysmal:
This PR fixes our recent CI failure.
It seems I omitted a reference in the ML part of the code that is used to store the last results from a matching session.

This stores this value into a domain local storage instead.

I'll let the libstr testsuite run locally with many more domains than our CI and see if anything else was forgotten, though I think it's only really this thing that was missing.


kayceesrk:
LGTM!

PR #658, opened 2021-10-01
Enable last dynlink test

kayceesrk:
The test seems to work fine now on my dev machine. Pushing to Github to see whether something breaks on CI.

PR #659, opened 2021-10-01
Reimport the threadsigmask test and remove systhread-todo test directory

abbysmal:
#658 reminded me of this testcase which is perfectly fine now.

I ran it for an hour on my machine with no issues, I think systhreads fixes + signals handling fixed this a while ago, I don't see why it wouldn't behave properly now.


ctk21:
Looks good, merging

PR #661, opened 2021-10-07
testsuite: re-enable pr9971 test

abbysmal:
I don't see any reasons why this would fail with the current state of systhreads, I had it run on my machine with debug runtime and couldn't get it to fail.
Let's just re-enable it.


ctk21:
Looks good, merging

PR #662, opened 2021-10-07
Disable changes check on 5.0

abbysmal:


ctk21:
merging, lets avoid the noise on the change entries for now

PR #669, opened 2021-10-07
Set thread names for domains

abbysmal:
This PR implements thread naming for Multicore OCaml various units.

It exposes as well an interface to allow one user to name Domains and Threads differently.

While this is working, I am unsure the macros logic is robust enough to work on anything beyond common Unixes and such. I tested it successfully on Linux and MacOS. The Windows implementation is left undone.

There's also a specific bit I solved by setting a 16 char restriction on the size of the name of a thread.
(following [glibc's](https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html) restrictions.)
This may be a bit too restrictive on other platforms where such a restriction may not exist?

One last thing, maybe this function should not sit in `Domain` (since it can be used for both systhreads and domains), but I think having a dedicated module for this does not sound great.
(or maybe it could fit into the future `ThreadLocal` module)

(extra check-typo pass on unix.c and win32.c)

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -281,7 +281,8 @@ static void caml_wait_interrupt_serviced(struct interruptor* target)
 void caml_domain_set_name(char_os *name)
 {
   char thread_name[MAX_DOMAIN_NAME_LENGTH];
-  snprintf_os(thread_name, MAX_DOMAIN_NAME_LENGTH, T("%s%d"), name, Caml_state->id);
+  snprintf_os(thread_name, MAX_DOMAIN_NAME_LENGTH,
+              T("%s%d"), name, Caml_state->id);
   caml_thread_setname_os(thread_name);
 }
 
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -277,6 +277,14 @@ static void caml_wait_interrupt_serviced(struct interruptor* target)
   }
 }
 
+#define MAX_DOMAIN_NAME_LENGTH 16
+void caml_domain_set_name(char_os *name)
+{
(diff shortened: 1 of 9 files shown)


abbysmal:
@sadiqj I rebased this branch and made check-typo adjustments.

runtime/domain.c
```diff
@@ -1320,3 +1333,16 @@ CAMLprim value caml_domain_dls_get(value unused)
...
+
+CAMLprim value caml_ml_domain_set_name(value name)
+{
+  CAMLparam1(name);
+  char_os* name_os;
+
+  if (caml_string_length(name) >= MAX_DOMAIN_NAME_LENGTH)
+    caml_invalid_argument("caml_ml_domain_set_name");
+  name_os = caml_stat_strdup_to_os(String_val(name));
+  caml_thread_setname_os(name_os);
```
abbysmal, line 1330: check return value

sadiqj:
@Engil this looks good to go. Can you rebase and address your review comment?

abbysmal:
@sadiqj the comment was more about double-checking than an actual todo, I think looking at the `pthread_setname_np` manpage checking is not required since the line above checks for the already very conservative name size limit.
I don't think we can get `ERANGE` returned after the previous check.
To me this PR is good to go.

sadiqj:
> To me this PR is good to go.

It looks like there's a conflict to resolve?

abbysmal:
> > To me this PR is good to go.
> 
> It looks like there's a conflict to resolve?

rebased and pushed, let's wait for CI.

sadiqj:
merged. Thanks @Engil

PR #672, opened 2021-10-08
Codefrag thread safety

sadiqj:
This PR intends to make codefrag thread safe. It achieves this by introducing a lock-free skiplist: https://github.com/sadiqj/ocaml-multicore/blob/codefrag/runtime/lf_skiplist.c

A few changes to codefrag itself are needed. We can no long free code fragments as soon as we remove them, they are instead added a list of removed fragments that are cleaned up during a later stop-the-world pause.


sadiqj:
(I've touched a few big files, so there's lots of hygiene to do)

xavierleroy:
Very impressive!  I didn't know about lock-free skip lists, and we do need a lock-free data structure for code fragments since the data structure is accessed  (read-only) from signal handlers (function `segv_handler` in `runtime/signals_nat.c`).  I'm sure we'll find other uses for lock-free skip lists in the future...  At any rate, this is an elegant solution to an otherwise thorny problem.

I'm not familiar enough with lock-free data structures to review this part of the PR.  Maybe @maranget could lend a hand?

There's a huge diff on `major_gc.c` -- so big that Github doesn't display it by default -- which is mostly reformatting of existing code.  Could you please perform the reformatting separately, either right now, pushing it to the multicore repo, or after this PR has been reviewed?

sadiqj:
> Very impressive!

Thanks. The requirement of access for a signal handler was the main motivation for the design (lock-free and lack of allocation/free). Hopefully we can find other opportunities in the runtime.
 
> There's a huge diff on `major_gc.c` -- so big that Github doesn't display it by default -- which is mostly reformatting of existing code. Could you please perform the reformatting separately, either right now, pushing it to the multicore repo, or after this PR has been reviewed?

Yes, this is currently what is blocking me on the PR. I started the work just prior to the multicore rebase from 4.12 to trunk and somewhere in that process check-typo got re-enabled in the github workflow. This means touching essentially any file in a PR requires fixing all the typo issues.

We were intending to fix things incrementally on a PR by PR basis but it might be worth taking the pain in one big PR.

xavierleroy:
> We were intending to fix things incrementally on a PR by PR basis but it might be worth taking the pain in one big PR.

Quite possibly.  Or just deactivating check-typo until just before the final merge.  I find this check-typo script to be more of an annoyance than of a help.

xavierleroy:
I also mentioned this code today to @jhjourdan @fpottier @damiendoligez .

maranget:
I think I found a space leak: not all cells get transferred into the `garbage_head` field of the skiplist structure when snipped out from the list.

The root cause is  in the compare exchange that attempt to set value 1 into the `garbage_next` field of he cell: https://github.com/ocaml-multicore/ocaml-multicore/blob/463d446e2c34d9677124068eb8505d0475ac7a30/runtime/lf_skiplist.c#L187
If the update above fails the cell is not transferred to the `garbage_head` field and the cell may not be freed later. Patch  [testsuite.txt](https://github.com/ocaml-multicore/ocaml-multicore/files/7335602/testsuite.txt) tests the leak.

A potential fix may be to initialise the `garbage_next` field to `NULL` at cell allocation time. See [fix.txt](https://github.com/ocaml-multicore/ocaml-multicore/files/7335609/fix.txt)

That said, I am not so sure about the fix. The above mentioned transition from NULL to 1 serves the purpose of not transferring a marked cell more than once to the `garbage_head` field.  Should a cell be reachable by several paths, we may now face some ABA problem: an "old" pointer points to a recycled cell, invalidating the skiplist property that all paths follow increasing key order. I am not so sure that this situation can occur.

sadiqj:
Thanks so much for taking a detailed look at this @maranget.

Failing to initialise the `garbage_next` field to `NULL` was a mistake on my part, it's basically just luck whether that field ends up zeroed or not as to whether we leaked the cell.

Testing the lengths of the list and garbage lists is a good idea, we should keep that. With the initialisation fix the test no longer fails with a space leak but I get an assertion failure in `hammer_skiplist`'s insert. Will try to figure out what that is in a bit.

On the issue in your final paragraph - I'm not sure I fully understand the circumstances in which that could happen. Cells are never recycled within a cleanup-to-cleanup period. We may be handed back the same address by the allocator in a later period but we will have had to go through a cleanup. A cleanup can only happen when there is no concurrent access, so all previous insert/removes must have returned and there can be no marked pointers in the structure (since the insert/remove `find`s must have returned).

maranget:
> On the issue in your final paragraph - I'm not sure I fully understand the circumstances in which that could happen. Cells are never recycled within a cleanup-to-cleanup period. We may be handed back the same address by the allocator in a later period but we will have had to go through a cleanup. A cleanup can only happen when there is no concurrent access, so all previous insert/removes must have returned and there can be no marked pointers in the structure (since the insert/remove `find`s must have returned).

It is likely that `caml_stat_alloc` recycles nodes, _i.e._ return pointers to nodes freed by `caml_stat_free_`. However, If no  node is present (_i.e_ reachable) in skiplist that is also present in the `garbage_head` list, nothing bad can occur.  I am not so sure that this is the case when cleanup begins.

sadiqj:
I figured out what the test failure was and I've created #690 which should clear up the `major_gc.c` diff when it's merged.

maranget:
> However, If no node is present (_i.e_ reachable) in skiplist that is also present in the `garbage_head` list, nothing bad can occur. I am not so sure that this is the case when cleanup begins.

I am now more confident that the `garbage_head` list and the skiplist are disjoint at cleanup. Hence, it is likely that no "ABA" problem can occur.

sadiqj:
> I am now more confident that the `garbage_head` list and the skiplist are disjoint at cleanup. Hence, it is likely that no "ABA" problem can occur.

I've been going through the implementation again to satisfy myself that this is the case. My justification is the `skiplist_find` in `caml_lf_skiplist_remove` will only return if there are no marked cells with `key` in the skip list. Any compare-and-swap contention with other threads results in retrying until completion. Since there can be no calls to `caml_lf_skiplist_remove` still in progress when cleanup is happening, there can be no marked cells in the skiplist.

maranget:
Hi, in case you find it useful, here are two concurrent tests. Those are concurrent versions of the cardinality test already included in the test suite. Here is [the patch](https://github.com/ocaml-multicore/ocaml-multicore/files/7373904/more.txt)

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+  sk->tail = caml_stat_alloc(SIZEOF_LF_SKIPCELL +
+                             NUM_LEVELS * sizeof(struct lf_skipcell *));
+  sk->tail->key = -1; // Would be better to have a UINTNAT_MAX
+  sk->tail->data = 0;
+  sk->tail->garbage_next = NULL;
+  sk->tail->top_level = NUM_LEVELS - 1;
+
+  /* We do this so we can disambiguate between garbage_next being NULL and being
+     set during cleanup */
+  sk->garbage_head = sk->head;
```
maranget, line 123: Why not initialise `sk->garbage_head` to the more conventional value `NULL`?

Looking at references to the `gabage_head` field in this file, I do not see how the comment applies. Moreover,  the field `garbage_head` does not seem to be referenced from other files.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+#define CAML_INTERNALS
+
+/* A concurrent dictionary data structure implemented as skip lists. This
+    implementation is based on the sequential skip list implemetation in
+    the runtime by Xavier Leroy but extends it to be safe under concurrent
+    modification. It has the property that insert/remove are lock-free and
+    contains is further wait-free. It is literally a textbook implementation
+    and can be found in Herlihy et al's "The Art of Multiprocessor
+    Programming" 2nd Edition, section 14.4. It only differs from the
+    textbook implementation to fix errors in the pseudocode, to add a racy
```
maranget: Minor remark, programs in my version of "The Art of Multiprocessor Programming" are in Java.
It may be interesting to have a list of the error you have found.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+  sk->head = caml_stat_alloc(SIZEOF_LF_SKIPCELL +
+                             NUM_LEVELS * sizeof(struct lf_skipcell *));
+  sk->head->key = 0;
+  sk->head->data = 0;
+  sk->head->garbage_next = NULL;
+  sk->head->top_level = NUM_LEVELS - 1;
+
+  sk->tail = caml_stat_alloc(SIZEOF_LF_SKIPCELL +
+                             NUM_LEVELS * sizeof(struct lf_skipcell *));
+  sk->tail->key = -1; // Would be better to have a UINTNAT_MAX
```
maranget: UINTNAT_MAX can be defined locally.

runtime/caml/lf_skiplist.h
```diff
@@ -0,0 +1,128 @@
...
+  in the header for FOREACH_LF_SKIPLIST_ELEMENT to work */
+#define LF_SK_IS_MARKED(p) ((p)&1)
+#define LF_SK_MARKED(p) ((struct lf_skipcell *)(((uintptr_t)(p)) | 1))
+#define LF_SK_UNMARK(p) ((struct lf_skipcell *)(((uintptr_t)(p)) & ~1))
+#define LF_SK_EXTRACT(from, mark_to, ptr_to)                                   \
+  {                                                                            \
+    uintptr_t tmp =                                                            \
+        (uintptr_t)atomic_load_explicit(&from, memory_order_acquire);          \
+    mark_to = LF_SK_IS_MARKED(tmp);                                            \
+    ptr_to = (struct lf_skipcell *)LF_SK_UNMARK(tmp);                          \
```
maranget: The cast looks redundant given that the definition of `LK_SK_UNMARK` already includes the same cast.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+  sk->tail->garbage_next = NULL;
+  sk->tail->top_level = NUM_LEVELS - 1;
+
+  /* We do this so we can disambiguate between garbage_next being NULL and being
+     set during cleanup */
+  sk->garbage_head = sk->head;
+
+  /* each level in the skip list starts of being just head pointing to tail */
+  for (int j = 0; j < NUM_LEVELS; j++) {
+    sk->head->forward[j] = sk->tail;
```
maranget: The cells of the `forward` arrays are of atomic type. I do not know what the compiler does when those are accessed in the standard way. It would probably be cleaner to use the `atomic_store_explicit(..,..,...)` primitive -- memory_order_relaxed probably being sufficient and `memory_order_release` safer, depending on the calling context of function`caml_lf_skiplist_init`.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+            while (1) {
+              current_garbage_head =
+                  atomic_load_explicit(&sk->garbage_head, memory_order_acquire);
+
+              atomic_store_explicit(&curr->garbage_next, current_garbage_head,
+                                    memory_order_release);
+
+              if (atomic_compare_exchange_strong(
+                      &sk->garbage_head,
+                      (struct lf_skipcell **)&current_garbage_head, curr)) {
```
maranget, line 218: The cast is redundant

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+        } else {
+          break;
+        }
+      }
+
+      preds[level] = pred;
+      succs[level] = curr;
+    }
+
+    if (curr->key == key) {
```
maranget: Why not `return curr->key == key ; ` (like in the book)?

sadiqj:
> Hi, in case you find it useful, here are two concurrent tests. Those are concurrent versions of the cardinality test already included in the test suite. Here is [the patch](https://github.com/ocaml-multicore/ocaml-multicore/files/7373904/more.txt)

Thank you! I've just added this.

runtime/lf_skiplist.c
sadiqj: Agreed, the comment is misleading.

The reason we initialise to `sk->head` is that we need to disambiguate on line 187 whether a cell's `garbage_next` is uninitialised and we can race to take ownership of it. If NULL was a valid value (because it was the last cell in the garbage list) then this would not work.

https://github.com/ocaml-multicore/ocaml-multicore/blob/e81c8e3b323b088fd0497f1bef68687a17ef5276/runtime/lf_skiplist.c#L187

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+     set during cleanup */
+  sk->garbage_head = sk->head;
+
+  /* each level in the skip list starts of being just head pointing to tail */
+  for (int j = 0; j < NUM_LEVELS; j++) {
+    sk->head->forward[j] = sk->tail;
+    sk->tail->forward[j] = NULL;
+  }
+}
+
```
maranget, line 134: The `skiplist_find` functions is quite close to the `find` function from the book TAMP.  Besides memory management, which looks ok,  one significant  difference is in the memory model. The book recommends declaring all fields to be `volatile` (Java style). This more or less corresponds to using`memory_order_seq_cst` everywhere as the memory model argument to atomic primitives (or to using the abbreviated forms, which is done here for `atomic_exchange_strong`). However, I have no reason to doubt the `memory_order_acquire` and `memory_order_release` used here to be wrong. In fact, I think that the load from `sk->garbage_head` and the store from  to `cur->garbage_next` (lines 195 and 197) could be `memory_order_relaxed`. Not being totally sure about this, I do not suggest the change.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+          is called when no thread can be concurrently modifying the skiplist.
+          */
+          if (atomic_compare_exchange_strong(&curr->garbage_next, &null_cell,
+                                             (struct lf_skipcell *)1)) {
+            /* Despite now having exclusivity of the current node's
+               garbage_next, having won the CAS, we might be racing another
+               thread to add a different node to the skiplist's garbage_head.
+               This is why we need to a retry loop and yet another CAS. */
+            while (1) {
+              current_garbage_head =
```
maranget: One can declare the local variable current_garbage_head here instead of line 164 above.

runtime/lf_skiplist.c
maranget: Now I see, quite smart. The comment may refer to line 187 in some stable manner.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+    }
+  }
+}
+
+/* [lf_skiplist_lookup] will return a skipcell or node that is greater than or
+   equal to the key provided, along with the node that directly proceeds it. It
+   is a much simplified version of [lf_skiplist_find] as it simply ignores
+   marked nodes and does not snip them out. As a consequence, it is wait-free.
+ */
+static struct lf_skipcell *lf_skiplist_lookup(struct lf_skiplist *sk,
```
maranget, line 257: This function is similar to method `contains` in TAMP. As far as I see there are some errors in the book that are corrected here: in the `while (marked)` loop `pred` is replaced by curr.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,500 @@
...
+     out marked nodes. If we did that for long enough we might leak memory. */
+  for (int level =
+           atomic_load_explicit(&sk->search_level, memory_order_relaxed);
+       level >= 0; level--) {
+    curr = LF_SK_UNMARK(
+        atomic_load_explicit(&pred->forward[level], memory_order_acquire));
+    while (1) {
+      LF_SK_EXTRACT(curr->forward[level], marked, succ);
+      while (marked) {
+        curr = LF_SK_UNMARK(
```
maranget: Cannot this affectation of `curr` be replaced by the more simple `curr = succ`?

runtime/lf_skiplist.c
sadiqj: Thanks. I've expanded the comment above 109 to explain in a bit more depth why we have to initialise as so.

runtime/lf_skiplist.c
sadiqj: Yes, I had intended that more as a reminder for us to maybe define UINTNAX_MAX in `config.h`. I know there's a few other places I've encountered -1 elsewhere in the codebase and a macro would probably be clearer.

runtime/caml/lf_skiplist.h
sadiqj: Agreed. Fix incoming.

runtime/lf_skiplist.c
sadiqj: Removing.

runtime/lf_skiplist.c
sadiqj: That is odd. I've changed it to your suggestion.

runtime/lf_skiplist.c
sadiqj: Yep. Done.

runtime/lf_skiplist.c
sadiqj: https://github.com/ocaml-multicore/ocaml-multicore/pull/672#discussion_r732600687 were the two errors I fixed in the Java version. I could explain them here or maybe in `contains`.

runtime/lf_skiplist.c
sadiqj: Yes, these were the errors that needed fixing.

There's also another key difference from the book that I thought I'd highlight. The book's `contains` starts at MAX_LEVEL whereas we include a `search_level` in the skiplist that we start from. This level never decreases but may increase as we add nodes. It may also be incorrect if there are races in setting it and when a `contains` races an `insert`. My reasoning is that this is fine though, since all searches must finish at the lowest level of the skiplist an incorrect `search_level` merely makes our searching less efficient (if it's too high we just examine sentinel nodes unnecessarily and if it's too low then we traverse too many nodes in the list). It should always be better than not having the optimisation though.

runtime/lf_skiplist.c
maranget: A reference to `contains/lf_skiplist_lookup` here may be sufficient.

sadiqj: I've explained it in lookup and referenced it in the earlier comment.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,504 @@
...
+      /* The bottom layer is what ultimately determines whether the node is
+       * present in the skiplist or not. We try to remove it and if we succeed
+       * then indicate so to the caller. If not then another thread raced us an
+         won. */
+      LF_SK_EXTRACT(to_remove->forward[0], marked, succ);
+      while (1) {
+        int mark_success = atomic_compare_exchange_strong(
+            &to_remove->forward[0], &succ, LF_SK_MARKED(succ));
+
+        LF_SK_EXTRACT(succs[0]->forward[0], marked, succ);
```
maranget: Isn't `succs[0]` always equal to `to_remove`? If so, why not simplify the previous line:
```ocaml
LF_SK_EXTRACT(to_remove->forward[0], marked, succ);
```

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,504 @@
...
+
+  while (curr != NULL) {
+    struct lf_skipcell *next = curr->garbage_next;
+
+    if (curr != sk->head) {
+      caml_stat_free(curr);
+    }
+
+    curr = next;
+  }
```
maranget: Maybe the loop can be simplified, stopping at sentinel? (current code is slightly more robust, protecting against replacing sentinel by NULL).
```suggestion
  struct lf_skipcell *head = sk->head;
  while (curr != head) {
    struct lf_skipcell *next = atomic_load_explicit(&curr->garbage_next,memory_order_relaxed);
    // acquire not useful, if executed in STW
    caml_stat_free(curr);
    curr = next;
  }
```

maranget:
I have reviewed file `lf_skiplist.c` (and interface `lf_skiplist.h`). I cannot really review the other files, by lack of knowledge of the runtime and GC.

runtime/lf_skiplist.c
sadiqj: Having removed this, I remember why this was there in the first place. Clang complains that `current_garbage_head` is an `Atomic lf_skipcell**` and so we need the cast to satisfy it.

runtime/lf_skiplist.c
sadiqj: Yes, thanks. This is clearer as NULL cannot be present in the garbage list and the existing implementation might confuse a reader.

runtime/lf_skiplist.c
sadiqj: Yes, I think this is correct. I've made the change.

runtime/lf_skiplist.c
sadiqj: I've added a UINTNAT_MAX to config.h to resolve this.

runtime/lf_skiplist.c
sadiqj: Yes, this is the correct thing to do here. I've replaced them with `atomic_store_explicit` and `memory_order_release`. `memory_order_relaxed` is sufficient for codefrag but hopefully we can find some other uses in the runtime for this - which may not have a thread creation between initialisation and use in a different thread.

runtime/caml/config.h
```diff
@@ -153,6 +153,8 @@ typedef uint64_t uintnat;
 #error "No integer type available to represent pointers"
 #endif
 
+#define UINTNAX_MAX -1
```
ctk21: shouldn't this be `UINTNAT_MAX`?
Although I can't see this used anywhere in the code (aside documentation). I'd fix and use, or remove.

runtime/lf_skiplist.c
```diff
@@ -0,0 +1,505 @@
...
+#endif
+
+/* Generate a random level for a new node: 0 with probability 3/4,
+   1 with probability 3/16, 2 with probability 3/64, etc.
+   We use a simple linear congruential PRNG (see Knuth vol 2) instead
+   of random(), because we need exactly 32 bits of pseudo-random data
+   (i.e. 2 * (NUM_LEVELS - 1)).  Moreover, the congruential PRNG
+   is faster and guaranteed to be deterministic (to reproduce bugs). */
+
+static uint32_t random_seed = 0;
```
ctk21: It 'might not matter', but concurrent calls to `random_level()` will each attempt to set this state in parallel. 
Might it be better if it were thread or domain local?

runtime/codefrag.c
```diff
@@ -17,25 +17,37 @@
...
-static struct skiplist code_fragments_by_num = SKIPLIST_STATIC_INITIALIZER;
+struct code_fragment_garbage *_Atomic garbage_head = NULL;
 
-static int code_fragments_counter = 0;
+static struct lf_skiplist code_fragments_by_pc;
 
-int caml_register_code_fragment(char * start, char * end,
+static struct lf_skiplist code_fragments_by_num;
+
+static int code_fragments_counter = 1;
```
ctk21: This variable is not atomic, but isn't it possible it can be altered inside `caml_register_code_fragment` from multiple domains?

ctk21 (commented):
My review has focused on how this is integrated with the GC, rather than the nitty gritty of the concurrent skiplist.

I believe it integrates with the runtime in a sensible way. Some minor things need cleaning up.

runtime/caml/config.h
sadiqj: Hah, I totally missed that I typoed this and then used it. It's in the lf_skiplist. Fixing..

runtime/lf_skiplist.c
sadiqj: This is a good point and I'm not sure what the best answer is here. If we make the seed thread local then every thread will produce the same sequence, which probably isn't what we want. On the other hand I don't think what we're doing at the moment is correct.

I could change this so that we use a CAS to update the random seed - which I think would give us a single sequence shared amongst all threads.

runtime/codefrag.c
sadiqj: Good catch, this needs fixing.

runtime/lf_skiplist.c
sadiqj: I've gone with the latter approach.

runtime/codefrag.c
sadiqj: Done.

runtime/lf_skiplist.c
sadiqj: I wasn't sure why this was necessary either but the pseudocode in TAMP does this for this case (but not for the other levels). I was wondering if there was a subtle issue I was missing but I think this might just be an oversight. I will change to using `to_remove`.

sadiqj:
> I have reviewed file `lf_skiplist.c` (and interface `lf_skiplist.h`). I cannot really review the other files, by lack of knowledge of the runtime and GC.

@maranget thanks so much for taking a really detailed look at the data structure.

runtime/lf_skiplist.c
xavierleroy: Yes, the CAS-based approach looks very good to me.  A thread-local seed could work too but then we'd have to initialize it properly, perhaps from the C library PRNG, each time a domain is created.

sadiqj:
I think I've gone through all the reviews and addressed the comments. Let me know if anyone else is planning to take a look otherwise we plan to merge early next week.

(We can also revisit in the working group - @Engil 's made sure he's happy with the changes)

maranget:
Hi, here is [one last addition](https://github.com/ocaml-multicore/ocaml-multicore/files/7400347/again.txt) to the parallel test: concurrent insertions and removes, with extra check of the absence of marked nodes after `Domain.join`'s.

maranget:
An impressive implementation: a real-life lock-free example, with style.

runtime/caml/lf_skiplist.h
```diff
@@ -0,0 +1,128 @@
...
+/*   the GNU Lesser General Public License version 2.1, with the          */
+/*   special exception on linking described in the file LICENSE.          */
+/*                                                                        */
+/**************************************************************************/
+
+/* A concurrent dictionary data structure implemented as skip lists. See
+   implementation for much more detail. */
+
+/* Keys and associated data are natural-width integers (type [uintnat]).
+   Key values 0 and uintnat_max are reserved for internal use. Do not use.
```
avsm:
```suggestion
   Key values 0 and uintnat_max are reserved for internal use; do not use.
```

just a minor tweak to connect the "do not use" to the key values sentence.

sadiqj:
> Hi, here is [one last addition](https://github.com/ocaml-multicore/ocaml-multicore/files/7400347/again.txt) to the parallel test: concurrent insertions and removes, with extra check of the absence of marked nodes after `Domain.join`'s.

Thank you, I've committed that.

ctk21:
Good stuff, merging. Many thanks for all the careful review here.

PR #682, opened 2021-10-11
Move effect handlers to its own module in Stdlib

kayceesrk:
This PR introduces `EffectHandlers` module under Stdlib moving it from its previous location in `Obj`.


stdlib/effectHandlers.ml
```diff
@@ -0,0 +1,136 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                                 OCaml                                  *)
+(*                                                                        *)
+(*      KC Sivaramakrishnan, Indian Institute of Technology, Madras       *)
+(*                                                                        *)
+(*   Copyright 1996 Institut National de Recherche en Informatique et     *)
```
ctk21: Really? I suspect the date at least is wrong!

stdlib/effectHandlers.mli
```diff
@@ -0,0 +1,108 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                                 OCaml                                  *)
+(*                                                                        *)
+(*      KC Sivaramakrishnan, Indian Institute of Technology, Madras       *)
+(*                                                                        *)
+(*   Copyright 1996 Institut National de Recherche en Informatique et     *)
```
ctk21: At least the date can't be right!

stdlib/stdlib.ml
```diff
@@ -581,64 +581,65 @@ external naked_pointers_checked : unit -> bool
 let () = if naked_pointers_checked () then at_exit major
 
 (*MODULE_ALIASES*)
-module Arg          = Arg
```
ctk21, line 584: By changing the indent here, we will introduce a large whitespace diff for merging upstream, which is fine if that's what we want to do.

ctk21 (commented):
I think this one works. Some very minor nits.

One question, I heard mention of making `EffectHandlers` something in `otherlibs` is that going to happen later or we've dropped that idea?

stdlib/stdlib.ml
kayceesrk: This was intentional. Not changing indent will make the new addition for `EffectHandler` not in line with the rest of the alias declarations. Github has nice support for ignoring whitespace change: https://github.com/ocaml-multicore/ocaml-multicore/pull/682/files?diff=split&w=1#diff-c092384db01339e4abcc95c97c7dd23e234677db093d01232c304d965db71eb9L596.

kayceesrk:
> One question, I heard mention of making EffectHandlers something in otherlibs is that going to happen later or we've dropped that idea?

We've dropped that idea.

Issue #683, opened 2021-10-11
`threadsigmask` CI failures


ctk21:
The test `lib-systhread/threadsigmask` got re-enabled in the CI with #659.

Sadly, the test still seems to fail sporadically. We have seen Linux Debug failures ([here](https://github.com/ctk21/ocaml-multicore/runs/3836762041?check_suite_focus=true#step:5:4861) and [here](https://github.com/ocaml-multicore/ocaml-multicore/runs/3829295924#step:5:4861)).

I took a brief look to try and recreate locally with MacOS and Linux Debug, but did not manage to recreate.

ctk21:
Also looks like it can fail sometimes on MacOS in CI (see [here](https://github.com/ctk21/ocaml-multicore/runs/3858293206?check_suite_focus=true#step:5:4038)).

ctk21:
Fixed in #688

PR #686, opened 2021-10-11
caml_threadstatus_wait should use caml_enter_blocking_section_no_pending

abbysmal:
Addresses #683.


abbysmal:
I do not think this is actually the right solution, closing.

PR #688, opened 2021-10-12
Better signal handling in systhreads

abbysmal:
This PR lands improvements in the signal handling in systhreads.

These two are also to be found in trunk, and were responsible for the `threadsigmask` testcase failure on our CI. (see #683)

This made me notice that outside of the various style clashes and rebase cruft, and Multicore specific bits, there are some more bits missing from trunk's systhreads.

Some appears to be fixes, some improvements, and I will take care of importing anything missing in a subsequent PR to our systhreads port. (that will also address diff and typos issues in the process)


ctk21:
Fixes #683 

LGTM, we block signals on the tick thread and also check signals on every systhread yield.

PR #689, opened 2021-10-13
Add EffectHandlers module

kayceesrk:
This PR adds the effect handlers functions to `4.12.0+domains+effects`. This allows `domainslib` using effect handler functions to work with `4.12.0+domains+effects` switch. See https://github.com/ocaml-multicore/domainslib/pull/51.


ctk21:
LGTM

PR #690, opened 2021-10-13
check-typo fixes for major_gc, so the changes in #672 don't get clobbered

sadiqj:


abbysmal:
Looks good to me, merging.

PR #691, opened 2021-10-14
Add ability to discontinue with backtrace

kayceesrk:
This feature will come in handy for getting the backtrace when modelling async/await. In the case where the awaited task raises an exception, we can get a backtrace that includes frames from both the awaited and awaiting task.

### DIFF
--- a/stdlib/effectHandlers.mli
+++ b/stdlib/effectHandlers.mli
@@ -40,6 +40,15 @@ module Deep : sig
       @raise Invalid_argument if the continuation has already been
       resumed. *)
 
+  val discontinue_with_backtrace:
+    ('a, 'b) continuation -> exn -> Printexc.raw_backtrace -> 'b
+  (** [discontinue_with_backtrace k e bt] resumes the continuation [k] by
+      raising the exception [e] in [k] using [bt] as the origin for the
+      exception.
+
+      @raise Invalid_argument if the continuation has already been
+      resumed. *)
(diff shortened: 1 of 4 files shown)


stdlib/effectHandlers.mli
```diff
@@ -99,6 +108,17 @@ module Shallow : sig
...
    *)
 
+  val discontinue_with_backtrace :
+    ('a,'b) continuation -> exn -> Printexc.raw_backtrace ->
+    ('b,'c) handler -> 'c
+  (** [discontinue_with k e bt h] resumes the continuation [k] by raising the
+      exception [e] with the handler [h] using the raw backtrace [bt] as the
+      origin of the exception.
+
+      @raise Invalid_argument if the continuation has already been resumed.
```
ctk21: Are we sure, I thought the exception is `Continuation_already_taken`

ctk21 (commented):
I think this is good.

Minor query on the documentation; I thought we had a dedicated exception `Continuation_already_taken` when we try to resume a continuation and it has already been resumed.

stdlib/effectHandlers.mli
kayceesrk: You are right. Fixed in https://github.com/ocaml-multicore/ocaml-multicore/pull/691/commits/e417c470e12b41a07e489a161c78b3fd7f8c62b1.

PR #694, opened 2021-10-14
First pass to improve the diff to startup code

abbysmal:
This PR attempts to improve the diff in the startup code on trunk and Multicore.

This actually does not quite improve the diff heavily, but it make sure to be more consistent with the changes introduced by Multicore when it comes to `runparams` and command line arguments (given to `ocamlrun`).

I also made sure to check that we did not drop any newer options from the list.

Side notes:
- As mentioned, Multicore handles such parameters within a dedicated struct (`caml_params`) that is read-only outside of `startup_aux`. This means that after startup, these parameters cannot be tempered with. In the Multicore setting, no other runtime code ever modify these fields after startup. This implies some amount of code motion versus trunk.
- In Multicore, there is still a few `runparams` disabled:
  - `a` (setting the allocation policy, ie `best_fit`, `first_fit` or `next_fit`.).
  - `H` (allocating pages using mmap/mmunmap)
  - `w` (the initial size of the major heap window)
- In Multicore, the atom table is handled differently, so the diffing here cannot really be worked around.

The last commit included in this PR also fixes what seems like an omission to me: if backtraced are enabled, it seems like we do not currently enable it for any other domains than the main one.
I moved the backtrace enabling behavior to `create_domain`, relying on `caml_params` to check if it needs to be enabled.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -23,6 +23,7 @@
 #include <pthread.h>
 #include <string.h>
 #include "caml/alloc.h"
+#include "caml/backtrace.h"
 #include "caml/callback.h"
 #include "caml/domain.h"
 #include "caml/domain_state.h"
@@ -441,6 +442,11 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
     domain_state->backtrace_buffer = NULL;
     domain_state->backtrace_last_exn = Val_unit;
     caml_register_generational_global_root(&domain_state->backtrace_last_exn);
+
+    if (caml_params->backtrace_enabled) {
+      caml_record_backtraces(1);
+    }
+
 #ifndef NATIVE_CODE
     domain_state->external_raise = NULL;
     domain_state->trap_sp_off = 1;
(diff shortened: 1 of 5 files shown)


abbysmal:
(also introduced in this PR, dubious decision on the coding style of domain.c to make check-typo happy, bikeshedding is allowed.)

runtime/domain.c
```diff
@@ -426,6 +436,11 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
     domain_state->backtrace_buffer = NULL;
     domain_state->backtrace_last_exn = Val_unit;
     caml_register_generational_global_root(&domain_state->backtrace_last_exn);
+
+    if (caml_params->backtrace_enabled) {
```
ctk21, line 446: This change is a good spot. In the old code we were only initializing the backtraces on the startup domain.

ctk21 (commented):
I think this PR does what it intends and makes us a bit closer to all the options that ocaml/ocaml supports on startup.

ctk21:
I rebased this as there were conflicts on `domain.c`.
Will wait for CI to green then merge

PR #695, opened 2021-10-14
Improve systhreads's diff with trunk

abbysmal:
This PR aims to improve Systhread's diff with trunk.

There is also at least one problematic behavior that was fixed in `caml_thread_stop` by making sure `Current_thread` indeed points to the correct data. (has it may not be the case once the thread reaches `caml_thread_stop`).

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -30,11 +30,6 @@
 #include "caml/printexc.h"
 #include "caml/roots.h"
 #include "caml/signals.h"
-#ifdef NATIVE_CODE
-#include "caml/stack.h"
-#else
-#include "caml/stacks.h"
-#endif
 #include "caml/sys.h"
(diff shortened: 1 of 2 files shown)


otherlibs/systhreads/st_stubs.c
```diff
@@ -199,9 +218,11 @@ void caml_thread_restore_runtime_state(void)
 
 static void caml_thread_enter_blocking_section(void)
 {
+  /* save the current runtime state in the thread descriptor
+     of the current thread */
   Current_thread = st_tls_get(Thread_key);
```
ctk21, line 223: not for this PR, but is it the case that we can avoid setting `Current_thread` here?
We will have come from OCaml, so it is either a fresh systhread or we will have come from a paired `caml_thread_leave_blocking_section`

ctk21:
LGTM, merging. 
Small comment on an optimisation that might be possible in future.

PR #696, opened 2021-10-14
stdlib: fix typos in `effectHandlers.mli`

craigfe:
(Let me know if this should target a different branch.)

### DIFF
--- a/stdlib/effectHandlers.mli
+++ b/stdlib/effectHandlers.mli
@@ -44,12 +44,12 @@ module Deep : sig
     { retc: 'a -> 'b;
       exnc: exn -> 'b;
       effc: 'c.'c eff -> (('c,'b) continuation -> 'b) option }
-  (** [('a,'b) handler] is a handler record with three field -- [retc]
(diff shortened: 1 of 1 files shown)


stdlib/effectHandlers.mli
```diff
@@ -71,7 +71,7 @@ module Shallow : sig
   (* Shallow handlers *)
 
   type ('a,'b) continuation
-  (** [('a,'b) continuation] is a delimited continuation that expects a ['a]
+  (** [('a,'b) continuation] is a delimited continuation that expects an ['a]
```
kayceesrk: I read `a ['a] value` as "a quote a value". Perhaps that's not how it should be read :-)

craigfe: Interesting :-) I read it as "an alpha value" (and sometimes as "an A value"), but I'm almost certainly in a minority here.

I've dropped this diff :+1:

kayceesrk: Thanks. :-)

Issue #452, opened 2020-12-06
no-effect-system not building with --disable-cfi


EduardoRFS:
When trying to build the `no-effect-system` branch with `--disable-cfi` the following error happens

```
amd64.S:450: Error: CFI instruction used without previous .cfi_startproc
```

PR #699, opened 2021-10-18
Cleanup fiber implementation and add documentation

kayceesrk:
When we perform external calls now, we no longer push 24 words at the top of the stack. Hence, this commit removes the extra space.

The commit also simplifies amd64.S by removing unused code, fixes formatting, etc.

### DIFF
--- a/runtime/amd64.S
+++ b/runtime/amd64.S
@@ -155,6 +155,16 @@
 /******************************************************************************/
 
 /* Switch from OCaml to C stack. Clobbers %r10, %r11. */
+#ifdef ASM_CFI_SUPPORTED
+#define SWITCH_OCAML_TO_C_CFI                                   \
+        CFI_REMEMBER_STATE;                                     \
+          /* %rsp points to the c_stack_link. */                \
+        .cfi_escape DW_CFA_def_cfa_expression, 5,               \
+           DW_OP_breg + DW_REG_rsp, Cstack_sp, DW_OP_deref,     \
+           DW_OP_plus_uconst, 8 /* retaddr */
+#else
+#define SWITCH_OCAML_TO_C_CFI
+#endif
 #define SWITCH_OCAML_TO_C                                  \
     /* Fill in Caml_state->current_stack->sp */            \
         movq    Caml_state(current_stack), %r10;           \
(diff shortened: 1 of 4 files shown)


runtime/amd64.S
```diff
@@ -176,38 +165,43 @@
         movq    %r10, Cstack_stack(%r11);                  \
     /* Switch to C stack */                                \
         movq    %r11, %rsp;                                \
-        SWITCH_OCAML_TO_C_NO_CTXT_CFI(PUSHED)
+        CFI_REMEMBER_STATE;                                \
+        .cfi_escape DW_CFA_def_cfa_expression, 5,          \
```
ctk21: This change looks to me that it will break the fix in #454 to make `--disable-cfi` work (reported in #452).
One way is to wrap the CFI magic into a block that respects `ASM_CFI_SUPPORTED`.

ctk21 (commented):
I believe this PR is fine. I agree that we do not need the addition of 24bytes for an external call, since we no longer have any additional information pushed to the stack when we transition from OCaml to C. 

There is one nit we should fix, let's make sure the `--disable-cfi` configure option still works.

runtime/caml/fiber.h
```diff
@@ -67,8 +87,18 @@ CAML_STATIC_ASSERT(sizeof(struct stack_info) == Stack_ctx_words * sizeof(value))
  * +------------------------+
  */
 
+
+/* This structure is used only for DWARF backtraces. This strcture is allocated
```
ctk21: I think this structure is about more than the DWARF backtraces. I would go with:
```
This structure is used for storing the OCaml return pointer when transitioning from 
a C stack to an OCaml stack. Care is needed to walk this linked list to update the 
OCaml stack pointers when an OCaml stack is reallocated. It is also used for 
DWARF backtraces. 
```

runtime/caml/fiber.h
```diff
@@ -67,8 +87,18 @@ CAML_STATIC_ASSERT(sizeof(struct stack_info) == Stack_ctx_words * sizeof(value))
...
+ * on the C stack at [caml_start_program]. Hence, there will be one of these on
+ * the stack for the main program entry and each callback. These fields are
+ * updated at [caml_c_call*] functions. */
 struct c_stack_link {
+  /* The reference to the fiber stack */
   struct stack_info* stack;
+  /* Since OCaml callbacks are executed on the last active fiber, there may be
+   * multiple logical OCaml stacks on the same physical stack. The [sp] field
+   * stores the stack pointer at [caml_c_call*] so that DWARF can correctly
```
ctk21: Isn't the primary use of the `sp` field to allow switching from a C stack back to an OCaml stack?
(for example `movq    Cstack_sp(%rsp), %rsp;` in `SWITCH_C_TO_OCAML_NO_CTXT` in `amd64.S`)
It is _also_ used by DWARF to correctly unwind the stack.

ctk21 (commented):
I think we want to change the documentation emphasis around the `c_stack_link` being a DWARF only concern.

runtime/caml/fiber.h
kayceesrk: Agreed. Good catch.

runtime/caml/fiber.h
kayceesrk: Agreed. Updated. I've gone with the text

```
/* This structure is used for storing the OCaml return pointer when
 * transitioning from an OCaml stack to a C stack at a C call. When an OCaml
 * stack is reallocated, this linked list is walked to update the OCaml stack
 * pointers. It is also used for DWARF backtraces. */
```

ctk21:
LGTM, merging!

Issue #666, opened 2021-10-07
systhreads: audit we don't have the issue that ocaml#10478 fixes


ctk21:
During the rebase to ocaml/trunk it wasn't immediately obvious if the ocaml-multicore implementation suffers from the bug fixed by [ocaml#10478](https://github.com/ocaml/ocaml/pull/10478).

Are we able to confirm that ocaml-multicore does not suffer from this bug?

ctk21:
Fixed by #702

PR #702, opened 2021-10-19
otherlibs: add PR10478 fix back to systhreads

abbysmal:
This PR addresses #666 and re-instate the fix from ocaml#10478 in systhreads.


ctk21:
Fixes #666 

Merging!

PR #705, opened 2021-10-20
otherlibs: remove caml_channel_mutex_io hooks from systhreads

abbysmal:
This PR removes the `caml_channel_mutex_io` hooks in systhreads.
The rationale is that this behavior is now the default behavior in Multicore-OCaml as of #649, and the relevant thread local values are kept thread local in our implementation (rather than domain-local), which means it behaves the same as systhreads.


ctk21:
LGTM, merging

PR #707, opened 2021-10-20
Move Domain.DLS to a ThreadLocal module and make it work under systhreads.

abbysmal:
As discussed in the mailing list, this draft implements the move of `Domain.DLS` to the `ThreadLocal` module.
It also enables the usage of thread-local storage to systhreads by adding `dls_root` as a field in each thread descriptors that needs to be a part of the usual systhreads context-switch.

I am pretty ambivalent about this implementation, because I could not find a correct interface for `ThreadLocal` because of how the internal `Domain.DLS` state is currently working.

The idea is that, currently:
- `Domain.DLS` will use the `create_tls` function to initialize the array stored in `dls_root` to store the various thread local values the user wants to store.
- This function is called at the toplevel of the module (so the main domain goes through it), and after each subsequent `Domain.spawn`, within the child domains.
- `create_tls` is hidden from external user, as it is not exposed in `domain.mli`. This is not a problem since only `Domain.DLS` and `Domain.spawn` really needs it.

This creates a bit of an issue if we were to allow systhreads to use the same `dls` facility, since `systhreads` can not access `create_dls`, which really is an implementation details users should not care about in the first place. (as is the current API designed).

So, I took the wrong option here (for the sake of rolling out a draft of this feature), and exposed `create_dls` in the `ThreadLocal` module so that systhreads can correctly initialize its state for every new thread being spawned. This function should not be called by users.

Keeping it in `Domain` does not help, as systhreads needs to know about the initial state setup when spawning threads.
Having an explicit setup function that the user needs to call manually. (thus making it ok to expose it) sounds cumbersome.
I am thinking that perhaps the right approach to dls is that its implementation should sit in the runtime, rather than in `stdlib` (as in, the high level OCaml code employed in the `ThreadLocal` interface maybe is better found in the runtime, sadly as C code, but at least in a footing that makes it easier to use accross the board for our usecases.)


abbysmal:
I am closing this PR, as the implementation is not quite functioning, the approach is not the right one, and the feature is not desired for the time being.

PR #709, opened 2021-10-22
Update VERSION string

kayceesrk:
Updates the VERSION string to `5.00.0+dev0`


ctk21:
I'm assuming we need this to play nice with opam and sandmark. 
If we make this change, can we also keep [ocaml-variants.opam](https://github.com/ocaml-multicore/ocaml-multicore/blob/5.00/ocaml-variants.opam) in sync?

kayceesrk:
Since `5.00` branch tracks ocaml trunk, we should keep the version as whatever trunk does `+domains`. Hence, closing this issue.

PR #712, opened 2021-10-22
otherlibs: Unix.kill should check for pending signals

abbysmal:
This PR re-enable the `unix_kill` testcase, and fix it by making sure `Unix.kill` does check for pending signals on return.

(which is an upstream fix, see ocaml/ocaml#9802)


sadiqj:
LGTM, merging.

ctk21:
This looks good to me and fixes up a discrepancy with upstream

PR #713, opened 2021-10-22
Clarify documentation of Lazy wrt. RacyLazy and Undefined exceptions

gadmm:
@kayceesrk Following discussion at #520. From the commit log:

---
- Remove ambiguous thread-safety claim
- Clarify raising behaviour of `try_force`
- Clarify unsafety before 5.0
---

With this clarification done, the asymmetry of `try_force` regarding race conditions appears clearly. I can also look into returning a value in every case and never raise if that's obviously better. (But as mentioned in the discussion above, it will not help the use-case I had in mind.)

### DIFF
--- a/stdlib/lazy.mli
+++ b/stdlib/lazy.mli
@@ -40,14 +40,16 @@ type 'a t = 'a CamlinternalLazy.t
    [lazy] keyword.  You should not use it directly.  Always use [Lazy.t]
    instead.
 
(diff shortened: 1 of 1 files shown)


kayceesrk:
LGTM. Thanks!

Separately, it may be useful to provide force_spin_wait which always returns a value, with "spin_wait" hopefully making the semantics clear?

Issue #714, opened 2021-10-25
Marshal segfault


ygrek:
```
$ echo 'Marshal.(to_string () [No_sharing])' | MALLOC_PERTURB_=33 ocaml -stdin
free(): invalid pointer
Aborted (core dumped)
```
```
$ gdb ~/.opam/4.12.0+domains/bin/ocamlrun core -ex bt
[...]
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007e065ad76535 in __GI_abort () at abort.c:79
#2  0x00007e065adcd508 in __libc_message (action=action@entry=do_abort, 
    fmt=fmt@entry=0x7e065aed828d "%s\n") at ../sysdeps/posix/libc_fatal.c:181
--Type <RET> for more, q to quit, c to continue without paging--
#3  0x00007e065add3c1a in malloc_printerr (
    str=str@entry=0x7e065aed643b "free(): invalid pointer") at malloc.c:5341
#4  0x00007e065add7b3e in free_check (mem=<optimized out>, 
    caller=<optimized out>) at hooks.c:254
#5  0x000059d8654593c4 in extern_free_position_table (s=s@entry=0x59d867387170)
    at extern.c:236
#6  0x000059d865459da8 in extern_rec (v=<optimized out>, s=0x59d867387170)
    at extern.c:871
#7  extern_value (s=s@entry=0x59d867387170, v=v@entry=1, 
    flags=flags@entry=138263587400648, 
    header=header@entry=0x7fff20d19a80 "\b\232\321 \377\177", 
    header_len=header_len@entry=0x7fff20d19a7c) at extern.c:894
#8  0x000059d86545add5 in caml_output_value_to_bytes (v=1, 
    flags=138263587400648) at extern.c:973
#9  0x000059d86546f141 in caml_interprete (prog=<optimized out>, 
    prog_size=<optimized out>) at interp.c:1029
#10 0x000059d865470b72 in caml_main (argv=0x7fff20d19ef8) at startup_byt.c:380
#11 0x000059d86544986c in main (argc=<optimized out>, argv=<optimized out>)
    at main.c:41
(gdb) frame 6 
[...]

PR #716, opened 2021-10-25
runtime: extern_free_position_table should return on extern_flags & NO_SHARING

abbysmal:
This addresses issue #714.

Should we also add a testcase for this, I wonder?


kayceesrk (approved):

kayceesrk:
LGTM. I don't think this needs a test case. Merging.

PR #718, opened 2021-10-26
Deprecate Sync and timer_ticks from Domain

bikallem:
Syncs 4.12.0+domains+effects with the mainline 5.00 branch, specifically the change from PR #704


ctk21:
I think this works and should provide a route for people to have the same code run on the `5.00` branch alongside the `4.12` multicore branches.

Sudha247 (commented):
Looks good! I think this sits well with both the old and new functions, too.

bikallem:
> I think this works and should provide a route for people to have the same code run on the `5.00` branch alongside the `4.12` multicore branches.

/cc @ctk21 should I take this as an approval?

ctk21:
LGTM, merging

bikallem:
Thanks.

PR #720, opened 2021-10-27
Improve ephemerons compatibility with testsuite

kayceesrk:
There are two improvements in this PR

1. Imports the upstream fixes for making ephemerons work with infix objects: https://github.com/ocaml/ocaml/pull/9742.
2. Clarifies why `weaktest.ml` fails on multicore and fixes the test by adding `Gc.full_major()` in the test. Here is the newly added comment copied from the `weaktest.ml` file: 

> Ephemeron / Weak array implementation in multicore OCaml differs significantly from stock OCaml. In particular, ephemerons keys and data in the minor heap are considered roots for the minor collection. Moreover, When blit'ing ephemerons, the source keys and data are marked as live to play nicely with the concurrent major GC. As a result, this test keeps the previous bunches of strings alive longer than on stock. Hence, the test fails. We add a [full_major] call that forces old bunches to be collected and to confirm that ephemeron implementation on multicore does work as intended.

### DIFF
--- a/runtime/caml/weak.h
+++ b/runtime/caml/weak.h
@@ -1,15 +1,17 @@
-/***********************************************************************/
-/*                                                                     */
-/*                                OCaml                                */
-/*                                                                     */
-/*            Damien Doligez, projet Para, INRIA Rocquencourt          */
-/*                                                                     */
-/*  Copyright 1997 Institut National de Recherche en Informatique et   */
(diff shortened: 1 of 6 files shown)


ctk21 (commented):
I think this PR implements what it intends.

kayceesrk:
Thanks for the review @ctk21.

PR #721, opened 2021-10-27
fix `make install`

bikallem:
This PR fixes an install bug which manifests as `/usr/bin/install: cannot stat 'caml/byte_domain_state.tbl': No such file or directory` when issuing `make install`.


ctk21:
Eak, fixed in #676 then broken in #672. 

Good catch, merging!

PR #722, opened 2021-10-27
testsuite: re-enable signals_alloc testcase

abbysmal:
This PR re-enable the `signals_alloc` testcase and attempts at making sure the bytecode interpreter polls for signals when it should.

I'm unsure if the fixes are not an overkill.

I do think that the `check_signals` macros are correct this way. (though I think `caml_check_for_pending_signals` is overkill vs trunk's `caml_something_to_do`, we may make it less expensive at some point.)

The problem lies in the `Enter_gc` macro I modified to explicitly check for signals:
I'm unsure it is fully correct and is the right polling point for signals. On `trunk`, `caml_small_alloc_dispatch` will do this duty just fine (and actually check for signals), versus our `Alloc_small` macro does not proceed in the same way.
Maybe the problem lies in how our `Alloc_small` code works rather than the interpreter. (though the native version of the testcase does not fail for us.).

For the record:
The issue we faced with `signals_alloc` was that, on bytecode, the testcase prints `01243` rather than `01234`.
The problem is that at `signals_alloc.ml:24`, this line triggers an allocation, that should make us poll for signals.
In the bytecode interpreter, it does not actually do that, and does not trigger the signal handler that would set the array in the expected order.

Careful review would be appreciated, and I think this change may be overblown and we could simplify the situation a lot.

### DIFF
--- a/runtime/caml/signals.h
+++ b/runtime/caml/signals.h
@@ -41,6 +41,7 @@ CAMLextern atomic_intnat caml_pending_signals[];
 #define caml_requested_major_slice (Caml_state_field(requested_major_slice))
 #define caml_requested_minor_gc (Caml_state_field(requested_minor_gc))
 
+int caml_check_for_pending_signals(void);
 void caml_update_young_limit(void);
 void caml_request_major_slice (void);
 void caml_request_minor_gc (void);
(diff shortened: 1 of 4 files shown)


sadiqj:
I think this fix is correct but that we should wait for https://github.com/ocaml-multicore/ocaml-multicore/pull/725 to merge - at which point we can simplify the test that happens to just checking `total_signals_pending`.

Also it's probably time we added `caml_process_pending_signals` and replaced all of the `caml_raise_if_exception(caml_process_pending_signals_exn)` we have all over the place. I will do that in another PR.

sadiqj:
I've added `caml_process_pending_signals` in https://github.com/ocaml-multicore/ocaml-multicore/pull/729

PR #724, opened 2021-10-27
runtime: remove unused fields from io.h

abbysmal:
`revealed` and `old_revealed` were removed from trunk a while ago.

See: https://github.com/ocaml/ocaml/pull/10136


ctk21:
LGTM, merging.

Issue #703, opened 2021-10-19
Possible loop in caml_enter_blocking_section when no domain can handle a blocked signal


abbysmal:
This can be triggered when a domain that blocked a specific set of signals ends up in a situation where no other domain can process the signal (for example: it is the last domain and a signal was recorded by another domain beforehand).

What can happen is that `caml_enter_blocking_section` will trigger a loop because processing all signals is made impossible:
This domain cannot possibly process the signal because the logic in `caml_process_pending_signals` makes it so that since it is masked, it will leave it unprocessed.

In normal circumstances, another domain that does not block it will eventually process the signals in question and this domain will eventually be freed, but when no other domain can do this, we end up in a loop.

PR #725, opened 2021-10-28
Blocked signal infinite loop fix

sadiqj:
This is a fix for https://github.com/ocaml-multicore/ocaml-multicore/issues/703

The problem stems from https://github.com/ocaml-multicore/ocaml-multicore/blob/5.00/runtime/signals.c#L152 where we need to check whether a signal has been delivered that races our entry in to a blocking section. If we have a waiting signal blocked then this loop will continue forever (or until another thread processes that waiting signal).

To fix this we introduce a monotonic `recorded_signals_counter`. We can replace the earlier conditional with a check that the `recorded_signals_counter` has not increased since we entered the blocking section. This also has the upside that it's cheaper than checking each of the pending counts for each signal we do at the moment.


kayceesrk:
Does this PR need a testcase?

sadiqj:
> Does this PR need a testcase?

We're working on one. It's tricky to build something that's not brittle.

sadiqj:
So there's now a test which hangs on 5.00 but will pass on this PR most of the time (thanks @Engil for some debugging help). The rest of the time it will segfault. We know what the cause of the segfault is, we'll create an issue for it.

sadiqj:
While https://github.com/ocaml-multicore/ocaml-multicore/issues/734#issuecomment-962033324 is an issue I've disabled the included test in this PR.

sadiqj:
I'm going to close this for now. The consensus at the working group was to move from counting signals to coalescing them, so this would probably need a rewrite regardless. The issue remains open.

PR #729, opened 2021-10-29
introduce caml_process_pending_signals which raises if exceptional

sadiqj:
This cleans up about a dozen: `caml_raise_if_exception(caml_process_pending_signals_exn())` calls we have littered around the codebase and also matches `caml_process_pending_actions` / `caml_process_pending_actions_exn` from trunk.


abbysmal:
LGTM, merging.

PR #731, opened 2021-11-03
AFL: segfault and lock resetting (fixes #497)

### LINKED ISSUE #497: Crash on AFL fuzzing
I am trying to AFL fuzz a simple program:
```ocaml
let _ = if ( read_int () mod 2) = 0 then () else failwith "false"
``` 
with `4.10.0+multicore` compiler. But it crashes with `PROGRAM ABORT : Fork server crashed with signal 11 Location : init_forkserver(), afl-fuzz.c:2230`. The backtrace from core dump is as below:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005597c3180639 in caml_raise (v=94110303735832) at fail_nat.c:63
63        if (Caml_state->c_stack == 0)
(gdb) bt
#0  0x00005597c3180639 in caml_raise (v=94110303735832) at fail_nat.c:63
#1  0x00005597c3180a30 in caml_raise_constant (tag=<optimized out>) at fail_nat.c:87
#2  caml_raise_out_of_memory () at fail_nat.c:146
#3  0x00005597c317e3c1 in caml_init_domains (minor_heap_wsz=262144) at domain.c:353
#4  0x00005597c3183ab0 in caml_init_gc () at gc_ctrl.c:301
#5  0x00005597c31803e6 in caml_startup_common (argv=0x7fffa027cdd8, pooling=pooling@entry=0) at startup_nat.c:117
#6  0x00005597c31805a8 in caml_startup_exn (argv=<optimized out>) at startup_nat.c:143
#7  caml_main (argv=<optimized out>) at startup_nat.c:148
#8  0x00005597c315f3ec in main (argc=<optimized out>, argv=<optimized out>) at main.c:44
```

Further on passing `-m none`, afl crashes with `PROGRAM ABORT : No instrumentation detected Location : perform_dry_run(), afl-fuzz.c:2908` and the corresponding backtrace is:
```
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f64f9d97801 in __GI_abort () at abort.c:79
#2  0x00005637a9c6747e in caml_fatal_error (msg=msg@entry=0x5637a9c8a954 "reading from afl-fuzz") at misc.c:138
#3  0x00005637a9c7d465 in afl_read () at afl.c:72
#4  0x00005637a9c7d5a6 in caml_setup_afl (unit=<optimized out>) at afl.c:133
#5  0x00005637a9c62afb in camlTest__entry ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```
Only time AFL runs is by disabling afl forkserver optimization with `AFL_NO_FORKSRV` and passing `-m none`.

jmid:
This is a fix to get AFL-instrumentation running on multicore again.

First off, the segfault reported in #497 is caused by a NULL-pointer dereference during error reporting.
The error is triggered because `afl-fuzz` runs the tested program under a memory budget with `setrlimit`:
  https://github.com/google/AFL/blob/61037103ae3722c8060ff7082994836a794f978e/afl-fuzz.c#L2034-L2040
The allocation of the minor heap space in `caml_init_domains` thus fails and calls `caml_raise_out_of_memory`.
It eventually ends up in `caml_raise` where it dereferences `Caml_state` which hasn't been initialized yet:
https://github.com/ocaml-multicore/ocaml-multicore/blob/13a6be2a538434b901404862dc37daadfc862850/runtime/fail_nat.c#L74
The PR changes `caml_init_domains` to use `caml_fatal_error` consistently.

Secondly, the AFL-instrumentation uses `fork` as an optimization to push more tests through.
For a simple setup, `afl-fuzz` can fork and execute a program repeatedly for each mutated input.
This is the behaviour you get with `AFL_NO_FORKSRV=1 afl-fuzz -i input -o output a.out` (also mentioned in #497).
To save time on redundant linking and initialization AFL also offers a "fork server" mode. This mode instead pauses the instrumented program at the point just before "proper execution" begins, and then cheaply forks test programs whenever `afl-fuzz` tells it to:
https://github.com/ocaml-multicore/ocaml-multicore/blob/13a6be2a538434b901404862dc37daadfc862850/runtime/afl.c#L116-L120
This setup is described in more detail in https://lcamtuf.blogspot.com/2014/10/fuzzing-binaries-without-execve.html
Note: Because we are still during initialization this `fork` occurrence will only execute with 1 domain running. 
The child however has problems with unlocking akin to #471. Since `afl-fuzz` silences the tested program
  https://github.com/google/AFL/blob/61037103ae3722c8060ff7082994836a794f978e/afl-fuzz.c#L2067-L2068
an `strace -f` helped reveal the problem:
```
...
[pid 97880] close(198)                  = 0
[pid 97880] close(199)                  = 0
[pid 97880] write(2, "Fatal error during unlock", 25) = 25
[pid 97880] write(2, ": Operation not permitted\n", 26) = 26
[pid 97880] exit_group(2)               = ?
[pid 97880] +++ exited with 2 +++
[pid 97879] <... wait4 resumed>[{WIFEXITED(s) && WEXITSTATUS(s) == 2}], WSTOPPED, NULL) = 97880
[pid 97879] --- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=97880, si_uid=1000, si_status=2, si_utime=0, si_stime=0} ---
[pid 97879] write(199, "\0\2\0\0", 4)   = 4
...
```
Because the child quickly fails silently during `afl-fuzz`s `perform_dry_run` it reports it as
`No instrumentation detected` which is a bit confusing https://github.com/google/AFL/blob/61037103ae3722c8060ff7082994836a794f978e/afl-fuzz.c#L2629-L2632

To fix the unlock issue the PR uses the `caml_atfork_hook` from 1a2c6120a12b414ef63e006806c732f4852f2420
With this fix the test case from #497 runs again using `-m none` (no memory budget).
I'll submit a separate PR regarding heap memory requirements.

### DIFF
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -1877,7 +1877,7 @@ let get_mod_field modname field =
 
 let code_force_lazy_block = get_mod_field "CamlinternalLazy" "force_lazy_block"
 
-let code_force_lazy = get_mod_field "CamlinternalLazy" "force"
+let code_force_lazy = get_mod_field "CamlinternalLazy" "force_gen"
 
 (* inline_lazy_force inlines the beginning of the code of Lazy.force. When
    the value argument is tagged as:
@@ -1986,7 +1986,7 @@ let inline_lazy_force arg loc =
       { ap_tailcall = Default_tailcall;
         ap_loc = loc;
         ap_func = Lazy.force code_force_lazy;
-        ap_args = [ arg ];
+        ap_args = [ Lconst (Const_base (Const_int 0)); arg ];
         ap_inlined = Default_inline;
         ap_specialised = Default_specialise
       }
--- a/runtime/afl.c
+++ b/runtime/afl.c
@@ -23,6 +23,7 @@ uintnat caml_afl_prev_loc;
 #if !defined(HAS_SYS_SHM_H) || !defined(HAS_SHMAT)
 
 #include "caml/mlvalues.h"
+#include "caml/domain.h"
 
 CAMLprim value caml_reset_afl_instrumentation(value full)
 {
@@ -117,6 +118,7 @@ CAMLexport value caml_setup_afl(value unit)
     int child_pid = fork();
     if (child_pid < 0) caml_fatal_error("afl-fuzz: could not fork");
     else if (child_pid == 0) {
+      caml_atfork_hook();
       /* Run the program */
       close(FORKSRV_FD_READ);
       close(FORKSRV_FD_WRITE);
--- a/runtime/afl.c
+++ b/runtime/afl.c
@@ -114,6 +114,10 @@ CAMLexport value caml_setup_afl(value unit)
   }
   afl_read();
 
+  /* ensure that another module has not already spawned a domain */
+  if (!caml_domain_alone())
+    caml_fatal_error("afl-fuzz: cannot fork with multiple domains running");
+
   while (1) {
     int child_pid = fork();
     if (child_pid < 0) caml_fatal_error("afl-fuzz: could not fork");
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -515,7 +515,7 @@ void caml_init_domains(uintnat minor_heap_wsz) {
   heaps_base = caml_mem_map(size, size, 1 /* reserve_only */);
   tls_base =
       caml_mem_map(tls_areas_size, tls_areas_size, 1 /* reserve_only */);
-  if (!heaps_base || !tls_base) caml_raise_out_of_memory();
+  if (!heaps_base || !tls_base) caml_fatal_error("Not enough heap memory to start up");
 
   caml_minor_heaps_base = (uintnat) heaps_base;
   caml_minor_heaps_end = (uintnat) heaps_base + size;
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -515,7 +515,8 @@ void caml_init_domains(uintnat minor_heap_wsz) {
   heaps_base = caml_mem_map(size, size, 1 /* reserve_only */);
   tls_base =
       caml_mem_map(tls_areas_size, tls_areas_size, 1 /* reserve_only */);
-  if (!heaps_base || !tls_base) caml_fatal_error("Not enough heap memory to start up");
+  if (!heaps_base || !tls_base)
+    caml_fatal_error("Not enough heap memory to start up");
 
   caml_minor_heaps_base = (uintnat) heaps_base;
   caml_minor_heaps_end = (uintnat) heaps_base + size;
--- a/stdlib/camlinternalLazy.ml
+++ b/stdlib/camlinternalLazy.ml
@@ -120,8 +120,8 @@ let try_force_gen_lazy_block ~only_val (blk : 'arg lazy_t) =
 
 (* [force_gen ~only_val:false] is not used, since [Lazy.force] is
    declared as a primitive whose code inlines the tag tests of its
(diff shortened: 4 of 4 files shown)


ctk21 (commented):
So using `caml_fatal_error` in `caml_init_domains` is a good thing.

I'm a bit worried about allowing fork in `caml_setup_afl` without checking that there are not multiple domains running, but I'm not an AFL expert. From what I could tell, `caml_setup_afl` is called from 'instrumented modules during intitialization'. Is it not possible for the following to happen:
- module A: uninstrumented and starts a domain
- module B: instrumented but initialized after module A at which point we have multiple domains

If this could happen, it could be fixed by inserting a `caml_domain_alone()` guard and erroring out if multiple domains are running?

kayceesrk:
> If this could happen, it could be fixed by inserting a caml_domain_alone() guard and erroring out if multiple domains are running?

This seems like a reasonable step to move forward.

jmid:
Following @ctk21's suggestion I've added a `caml_domain_alone` guard to ensure that the afl-instrumentation's "fork-server" does not invoke `fork` after other domains have already spawned.

jmid:
I noticed that `./configure -afl-instrument` and `make` fails with:
```
../boot/ocamlrun ../ocamlopt -strict-sequence -absname -w +a-4-9-41-42-44-45-48-70 -g -warn-error +A -bin-annot -nostdlib -principal -safe-string -strict-formats  -function-sections  -afl-inst-ratio 0 -c camlinternalLazy.ml
../boot/ocamlrun ../ocamlopt -strict-sequence -absname -w +a-4-9-41-42-44-45-48-70 -g -warn-error +A -bin-annot -nostdlib -principal -safe-string -strict-formats  -function-sections   \
           -o stdlib__Lazy.cmx -c lazy.ml
>> Fatal error: Primitive CamlinternalLazy.force not found.
Fatal error: exception Misc.Fatal_error
make[4]: *** [Makefile:237: stdlib__Lazy.cmx] Error 2
```

This is caused by c0ef11fa47a8ae0ff42f89177f0980f7903f028c AFAICS which unified `force` and `force_val` in one generic `force_gen` operation (CC @gadmm). For now I've gone with just calling that, but one could also roll back the unified interface to minimize interface changes to trunk. I've furthermore updated the comment in camlinternalLazy.ml to match the updated comment in trunk OCaml, so that it is clearer that the operation is actually used.

Thematically this is part of getting "AFL instrumentation working again" - but this can also be submitted as a separate PR if you prefer.

gadmm:
> This is caused by [c0ef11f](https://github.com/ocaml-multicore/ocaml-multicore/commit/c0ef11fa47a8ae0ff42f89177f0980f7903f028c) AFAICS which unified `force` and `force_val` in one generic `force_gen` operation (CC @gadmm). For now I've gone with just calling that, but one could also roll back the unified interface to minimize interface changes to trunk. I've furthermore updated the comment in camlinternalLazy.ml to match the updated comment in trunk OCaml, so that it is clearer that the operation is actually used.

Thanks, this seems correct to me, but I am not an expert of the affected code so it is best proof-read by someone else as well.

kayceesrk (approved):
The changes look good to me.

PR #732, opened 2021-11-04
Reduce minor heap reservation

jmid:
This PR will most likely turn out differently in the end, so I'm submitting it separately from #731.
As mentioned in #731 `afl-fuzz` runs a tested program with a `setrlimit`-inforced memory budget. By default there is not a lot to work with:
```
$ afl-fuzz | grep -- '-m'
  -m megs       - memory limit for child process (50 MB)
```
Incidentally this also explains why #497 reports of different behaviours when run under a 50MB budget (crash) compared to running without limitations (`-m none` - unlock-error/`No instrumentation`). Nevertheless, the low 50MB  default has been sufficient up to now, e.g., for fuzzng the one-line example from #497 with `ocaml 4.10.0`.

So how much does multicore need in comparison?
With the fork unlock fix from #731 multicore requires `-m 513G` (513GB) to run the example from #497 on my machine.
This bound can be confirmed with either `afl-fuzz` or the simpler `afl-showmap`: 
```
echo 0 | afl-showmap -m 513G -o out fuzz/test-5.00-forkfix
afl-fuzz -m 513G -i input -o output fuzz/test-5.00-forkfix
```
I understand that this is just a virtual memory reservation, nevertheless for the example to succeed `gdb` tells me that this `caml_mem_map` call passes a `size` of 274877906944 bytes
https://github.com/ocaml-multicore/ocaml-multicore/blob/13a6be2a538434b901404862dc37daadfc862850/runtime/domain.c#L515
which is then doubled and rounded up:
https://github.com/ocaml-multicore/ocaml-multicore/blob/13a6be2a538434b901404862dc37daadfc862850/runtime/platform.c#L134-L136

When AFL's requirements are lower it will instead fail...
In comparison 4.10.0+multicore needs `-m 8197M` on my machine. Since then there was a change to interpret `Minor_heap_max` as words rather than bytes: ade8a064faed15cedac429d45c07896e32aab6f9 I'm not sure it accounts for all of the difference though.

Bottomline: with `afl-fuzz` engineered as it is, this is a visible change from a user's perspective if they try to port a sequential program and fuzz it with 5.00. The PR proposes the simplest fix possible, by setting `Minor_heap_max` to 2097152 words.
This brings down the memory requirement for the example to `-m 4101`.
Perhaps the value should be even lower - or perhaps we should adjust `Max_domains`?

### DIFF
--- a/runtime/caml/config.h
+++ b/runtime/caml/config.h
@@ -223,7 +223,7 @@ typedef uint64_t uintnat;
 /* Maximum size of the minor zone (words).
    Must be greater than or equal to [Minor_heap_min].
 */
-#define Minor_heap_max (1 << 28)
+#define Minor_heap_max (1 << 21)
 
 /* Default size of the minor zone. (words)  */
 #define Minor_heap_def 262144


runtime/caml/config.h
```diff
@@ -223,7 +223,7 @@ typedef uint64_t uintnat;
 /* Maximum size of the minor zone (words).
    Must be greater than or equal to [Minor_heap_min].
 */
-#define Minor_heap_max (1 << 28)
+#define Minor_heap_max (1 << 21)
```
gadmm, line 226: This is very small (16MB); Coq uses 256MB by default. Instead, the reservation of contiguous address space for minor heaps is a remnant of the concurrent collector which should probably be removed.

ctk21:
A design assumption in multicore was that we could reserve address space and not commit that address space for use without it impacting users. This is an example where reserving large amounts of address space has an impact; note that most of the address space isn't committed.

As things currently stand, the parallel minor collector needs to have a region containing the minor heaps of all domains; this enables for a fast `Is_young` check based on the top and bottom of the region. Hence at startup multicore reserves the maximum required area and then commits it as needed by domains. 

Other designs for orchestrating the minor heaps of domains could be considered which provide fast `Is_young` checks but also have alternative performance for: domain spawn, changing the minor heap size via the `Gc.set` interface and virtual memory (even that reserved and uncommitted). I've opened #733 for that larger issue.

Within the context of this PR, squashing `Minor_heap_max ` is a blunt and harsh tool. An alternative approach might be to allow for a boot up param (e.g. `OCAMLRUNPARAM`) which overrides `Minor_heap_max` to a new value. I won't argue that is elegant, but it is a workaround.

runtime/caml/config.h
jmid: OK, thanks for the feedback.
After opening the PR I learned that the increase from ade8a06 is intentional.
I expect #733 will track this going forward (and that this will be closed).

jmid:
In case someone from the future stumbles over this PR:
For AFL-users the increased memory reservation (`mmap`) means that any invocation of `afl-fuzz` without an explicit `-m` option, e.g. `afl-fuzz -i input -o output ./readline` from [the OCaml manual](https://ocaml.org/manual/afl-fuzz.html) will fail with the following AFL error message despite the program being instrumented correctly. This is because of AFL's default 50MB virtual memory limit:
```
...
[*] Spinning up the fork server...
[+] All right - fork server is up.

[-] PROGRAM ABORT : No instrumentation detected
         Location : perform_dry_run(), afl-fuzz.c:2860
```

The current solution (after merging of #731) is to run `afl-fuzz` without a memory bound (or with a sufficiently large one):
```
 afl-fuzz -i input -o output -m none ./readline
 afl-fuzz -i input -o output -m 513G ./readline
```
A reduction below 50MB resulting from #733 is required to change this, AFAIU.

runtime/caml/config.h
kayceesrk: > the reservation of contiguous address space for minor heaps is a remnant of the concurrent collector which should probably be removed.

@gadmm for some additional context, we experimented with domain-local allocation buffers (DLABs), where a user-configurable minor heap area is allocated by a single domain and each domain carves a page for its own use. This implementation turned out to suffer from severe memory hierarchy impact (suspect both NUMA and cache effects). So performance-wise, it turned out to be sensible to retain the current implementation. See the PR https://github.com/ocaml-multicore/ocaml-multicore/pull/508 and the report: https://github.com/ocaml-multicore/ocaml-multicore/wiki/Domain-Local-Allocation-Buffers-Addendum. 

In terms of desiderata, any alternative design we come up will need to have:

1. A fast `Is_minor` check so that we can very cheaply check whether a value is in the minor heap. This is true in the current design since all of the minor heap arenas are in a reserved contiguous segment. `Is_minor` check needs only two comparisons. 
2. A minor heap allocation strategy where each domain allocates its own memory. This is true in the current design since the primary domain only reserves the contiguous space, but the individual domains themselves commit the memory. This turns out to be quite important as the DLAB work shows.

gadmm: @kayceesrk Thanks for the additional context. I am not sure to understand what was the motivation behind DLABs (the expected benefits) and I cannot tell what is going on from the description. The design I suggested at #733 is equivalent to the current one in terms of the domain allocating and keeping its own memory, so I expect that it avoids these issues. (But if the tail cost of spawning a domain is not an issue, then growing the contiguous space geometrically after a STW phase sounds easier in short term, of course.)

gadmm: @kayceesrk I now read more about DLABs and better understand how they reduced virtual memory consumption. If I understand correctly, #508 did not yet implement non-fixed minor heap sizes, so we know for sure that there is no impact of varying minor heap sizes on the rare observed speedup / some of the slowdown?

In addition to lower virtual memory usage, there is a benefit of allocating (not merely reserving) as contiguously as possible (i.e. avoiding big holes of unmapped memory) in terms of cache interference (according to the Streamflow paper) and MMU cache locality (in my experiment), which both DLABs and my design should have over the current one. I understand the wall you hit with DLABs. To sum up, implementing `Is_young` with a bibop is a way to open the design space by not requiring the minor heaps to occupy a fixed contiguous area. But I agree that it makes sense to try again with the DLABs design as suggested by @ctk21 in a way that respects your point 2. (smaller investment, higher cost of spawning a domain, solves a smaller problem).

Thanks for mentioning NUMA. The page allocator in my experiment should certainly be made NUMA-aware (but providing such control is what a page allocator is there for after all).

kayceesrk: The point of DLABs, which is not mentioned in the description, is to address a form of early promotion in the current design. In the current design, we trigger a minor collection when any one of the domains exhausts its minor heap arena. If we have `N` domains, each with `S` words of the minor heap space, it is likely that we will trigger a minor GC when the used space in the minor heap is much less than `N * S`. (We haven't empirically measured this, and we should if we are reviving the DLABs work). In the worst case, we will trigger a minor GC when one of the domains fills up its minor heap arena and others are empty -- worst-case utilised space is `S`. 

The idea of DLABs was to get close to `N * S` utilization by making the domains request pages of size `P` from the shared minor heap space. In the worst case, a minor GC will be triggered when one domain has filled up the last available page and all the others haven't started filling their last requested pages. So `N * S - (N-1) * P`. Given that `P << S`, this would be a win in terms of preventing early promotion. This was the reasoning behind going for DLABS. 

> Thanks for mentioning NUMA. The page allocator in my experiment should certainly be made NUMA-aware (but providing such control is what a page allocator is there for after all).

You will get NUMA friendliness on Linux by default if you allocate memory from the thread that will use it.

We should discuss the possibility of adapting bibop for multicore minor heap in https://github.com/ocaml-multicore/ocaml-multicore/issues/733 in order to avoid splintering the discussions across several threads. I had some questions here: https://github.com/ocaml-multicore/ocaml-multicore/issues/733#issuecomment-963807435.

gadmm: (About DLABs; I will discuss bibop elsewhere.)

I understood later. With the current multicore design it is hard to avoid false promotions, and it sounds better to collect when a fixed amount has been allocated (but this does not entirely avoid premature collections). This idea could be tried with a page allocator (minor heap extents could be pooled per core or per-NUMA-node).

> You will get NUMA friendliness on Linux by default if you allocate memory from the thread that will use it.

I expect it affects how I implement reclamation (`MADV_FREE` would probably not be enough).

Reading the DLABs discussions (esp. the commit_decommit variant) I also wanted to mention that you can map the whole mapping once in a thread, instead of mapping N times where N is the number of domains, since what counts for node attribution is who first touches the page.

jmid:
I'll close this and let the discussion continue in #733.

PR #735, opened 2021-11-05
add caml_young_alloc_start and caml_young_alloc_end in minor_gc.c

moyodiallo:
This PR aims to fix a retro-compatibility with the package [wall](https://github.com/let-def/wall/tree/master/lib).
The package use "caml_young_alloc_end"  in the file [wall/lib/wall__backend_stubs.c](https://github.com/let-def/wall/blob/master/lib/wall__backend_stubs.c)

`caml_young_alloc_start` and `caml_young_alloc_end` are not present in `ocaml-multicore`. 
I discussed with @Engil about them in `trunk` and we deduced, they should be the same as `young_start` and `young_end`.

### DIFF
--- a/runtime/caml/minor_gc.h
+++ b/runtime/caml/minor_gc.h
@@ -63,7 +63,9 @@ CAMLextern void caml_minor_collection (void);
 
 #ifdef CAML_INTERNALS
 extern void caml_set_minor_heap_size (asize_t); /* size in bytes */
-extern void caml_empty_minor_heap_no_major_slice_from_stw (caml_domain_state* domain, void* unused, int participating_count, caml_domain_state** participating); /* in STW */
+extern void caml_empty_minor_heap_no_major_slice_from_stw 
+  (caml_domain_state* domain, void* unused, int participating_count, 
+    caml_domain_state** participating); /* in STW */
 extern int caml_try_stw_empty_minor_heap_on_all_domains(); /* out STW */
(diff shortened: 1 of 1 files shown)


abbysmal:
I think this make sense as a compatibility macro.

One missing bit that we left out if `caml_young_alloc_mid`, because as it stands this does not seem to cause any breakage with OCaml packages in the wild.

The wider question is should we re-introduce these as macro (for compatibility's sake), or should we re-introduce these as proper fields in `Caml_state` and handle them properly within the minor gc code?

I remember digging for a while with @ctk21 and we found that these fields, more specifically `caml_yound_alloc_mid`, were introduced in https://github.com/ocaml/ocaml/pull/297
What we found is that `alloc_mid` was used as a marker to trigger marking work at the mid way point of the minor heap allocation buffer.
We do not have this feature in Multicore. Maybe we should re-introduce it (and so bring back those fields as they would become needed, `caml_young_alloc_mid` being a fast way to check for the mid point of the allocation window). ?

damiendoligez:
It does makes sense to have them as compatibility macros, and it does make sense to leave out `caml_young_alloc_mid`, if you're looking for a quick way to restore compatibility.

It would be cleaner to provide a function to do what `wall` needs: the total number of words allocated by the program so far, but of course that would need a change to the `wall` source code.

damiendoligez (approved):
LGTM

kayceesrk:
Thanks @damiendoligez for the review.

PR #739, opened 2021-11-11
Make ephemerons immutable.

kayceesrk:
This is a port of https://github.com/ocaml/ocaml/pull/10737 for multicore. Compared to https://github.com/ocaml/ocaml/pull/10737 it removes all the deprecated functions.


kayceesrk:
The consensus is that we keep the `Bucket` module in.

kayceesrk:
@ctk21 this PR should be good to go except for a quick review.

ctk21 (commented):
I took a (time limited) look. The PR does not alter the underlying runtime ephemeron GC mechanisms. However I think it does remove deprecated functions and makes the OCaml interface match up with post ocaml/ocaml#10737

I spotted a minor nit: the PR still references to the removed `iter`, `fold` functions in the documentation strings in `ephemeron.ml`.

I did not deeply review the testsuite alterations.

ctk21:
I have further taken a look at the alterations to `ephetest_par.ml`. I believe the alterations are the correct ones in this case; that is I didn't spot anything where we have changed the behaviour of the test with this PR.

kayceesrk:
Thanks @ctk21. Removed mentions of `iter` and `fold*` from documentation strings.

kayceesrk:
Let me know if anything else looks odd..

ctk21:
LGTM, merging!

testsuite/tests/weak-ephe-final/ephetest_par.ml
```diff
@@ -9,6 +9,9 @@ open Ephemeron
 
 let debug = false
 
+open Printf
+open Ephemeron
+
```
damiendoligez, line 14: Redundant opens.
```suggestion
```

damiendoligez (approved):
Looks good except for a small hiccup in `ephetest_par.ml`.

testsuite/tests/weak-ephe-final/ephetest_par.ml
kayceesrk: Fixed by #804.

PR #740, opened 2021-11-12
Systhread lifecycle

abbysmal:
This PR is divided in three commits and addresses different issues with systhread's lifecycle in Multicore.

### caml_thread_domain_stop_hook fixes

3cd2ce0d54b4b65a69633e631cb80497eb06ac7b implements a fix to the `caml_domain_stop_hook` implementation and make sure everything is covered properly.

Everything should be commented properly, and there's is an attached testcase.
The gist of it is: We make sure that we join remaining systhreads when a domain is exiting, so that we do not have leftover systhreads when the domain's main thread exits.
Letting other systhreads run when a domain's (but not `domain 0`) exits would involve a lot of hacking in the codebase to make sure everything work reliably.

It does not seem necessary to push for this to be supported.

Alternatively, we could also just push for this code to actually fail if we detect that there are remaining threads, and throw an exception.
This would be arguably better than the current situation where the code could eventually hang on non-terminating systhreads.

### Thread.exit fixups

127c3fcc912d7370a31fb5e5d1a67784672c58d6 implements Thread.exit properly under a single-domain setting.

It reimports some missing code and make sure it does behave the same as a traditional sequential program.
Supporting `Thread.exit` on multi-domains, would, however, prove to be a massively difficult task because of the extra book keeping required.
Domain 0 (the actual program's main thread) exiting and Domain 1 (a child thread) exiting would behave differently when using `pthread_exit`, for instance.
Adding this on top of the different implementation for `Thread.exit` on both the native and bytecode backend, and the existence of the backup thread in Multicore, makes `Thread.exit` even harder to get right in a multicore setting.

https://github.com/ocaml/ocaml/issues/9071 also seems to suggest that those observations sits in an overall bigger picture questioning the merits of this API, so I think it is fair for us to only support it in a retro-compatible fashion in sequential program and have it fail under Multicore.

### Fix in caml_c_thread_unregister
3a3fb8644dd89d1f46554fc1ccc8b47e6c58cfe0 implements a fix into the unregister code, where we restore the runtime state when there's possibly the thread left in the chaining.


abbysmal:
I believe this PR is ready for review and addresses a few issues mentioned during WG3. (as well as other issues that were outstanding.)
This PR as started before WG3, so some discussions happened since then, specifically related to `Thread.exit`.

Would you have time to take a look at these changes @xavierleroy ?

Thank you very much,.

otherlibs/systhreads/st_stubs.c
```diff
@@ -355,15 +355,50 @@ static void caml_thread_domain_start_hook(void) {
...
+    th = Current_thread->next;
+
+    while (th != Current_thread) {
+      next = th->next;
+      caml_thread_join(th->descr);
+      th = next;
+    }
+
+    /* We should have joined on all threads */
+    Assert(Current_thread->next = NULL);
```
xavierleroy: I was worried this could block the domain for a long time *and* prevent it from participating in GC, but `caml_thread_join` calls `caml_threadstatus_wait`, which is a blocking section, so this should be all right.

otherlibs/systhreads/st_stubs.c
```diff
@@ -675,6 +712,10 @@ CAMLprim value caml_thread_exit(value unit)   /* ML */
 {
   struct longjmp_buffer * exit_buf = NULL;
 
+  /* we check if another domain was ever started */
+  if (caml_domain_is_multicore())
+    caml_fatal_error("Thread.exit: unsupported call under multiple domains");
```
xavierleroy: OK, but I'd rather call `caml_invalid_argument` so that the whole program is not aborted and the current thread gets a chance to react to the error (finalization).

xavierleroy (commented):
Looks good to me, with one suggested change below.

otherlibs/systhreads/st_stubs.c
abbysmal: Right.
One other thing that is worth pondering over a bit, is what happens if one such _rogue_ systhread was to spawn new systhreads. (while the initial thread tries to join.)

I'm mostly convinced it is ok and the way the chaining is handled and traversed makes it that such new threads would be waited upon organically, at they would be inserted _before_ `Current_thread` (in this case, the initial thread).

I do not think we can miss such a newly spawned thread. But more eyes thinking on this specifically would be reassuring.

abbysmal:
Thank you for your review: I switched the `fatal_error` to `invalid_argument` as suggested.

otherlibs/systhreads/st_stubs.c
xavierleroy: > I'm mostly convinced it is ok and the way the chaining is handled and traversed makes it that such new threads would be waited upon organically, at they would be inserted before Current_thread (in this case, the initial thread).

If the thread `th` that is being joined creates new threads before terminating, they will be inserted "between" `th` and `next`, so `th = next` will skip over them.  However, when`th` terminates, they will become reachable from the thread before `th`, so maybe they will be joined eventually.

I'm more worried about the thread `th->next` terminating while we're joining `th`.  Then, `th = next` will point `th` to freed memory...

abbysmal: Right, it is entirely possible that `th` ends up pointing to a freed memory space.
Maybe this loop is too simple for its own good and we would benefit from just inspecting the chaining always starting from the joining thread's entry?
Instead of looping `next`, we would always start from `Current_thread->next`, until it is `NULL`.
I think this would work fine (assuming the chaining is properly maintained by exiting threads) and would rule out possibles races.

xavierleroy: > we would always start from Current_thread->next, until it is NULL

You mean "until it is Current_thread" ?  Clever idea!  I think it should work fine.  The chaining of threads has to be maintained properly anyway, otherwise GC would crash when scanning their roots.

xavierleroy: Moreover, `Current_thread->next` cannot change while we're holding the master lock, so the test `Current_thread->next != Current_thread` is not racy.

abbysmal: Indeed, `Current_thread->next == Current_thread`, not `NULL`, my bad!
I just pushed the relevant change, as well as removing the assertion under this specific loop, that was not only invalid (`= NULL` rather than `== Current_thread`), but since this is also the precondition for exiting the loop now, it is not useful.

PR #745, opened 2021-11-16
Systhreads WG3 comments

abbysmal:
This PR addresses a few comments from the WG3 preliminary review cycle.

Commit names should be self-descriptive.

I felt confident with the suggested changes regarding some memory ordering and switch to non-atomic variables (especially for `m->busy`).

For the failure to allocate thread descriptors: I think raising OOM is fine there, but maybe we want these to be fatal errors?

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -370,6 +370,9 @@ static void caml_thread_initialize_domain()
   new_thread =
     (caml_thread_t) caml_stat_alloc_noexc(sizeof(struct caml_thread_struct));
 
+  if (new_thread == NULL)
+    caml_raise_out_of_memory();
+
   new_thread->descr = caml_thread_new_descriptor(Val_unit);
   new_thread->next = new_thread;
   new_thread->prev = new_thread;
@@ -527,6 +530,10 @@ CAMLprim value caml_thread_new(value clos)          /* ML */
 #endif
   /* Create a thread info block */
   th = caml_thread_new_info();
+
+  if (th == NULL)
+    caml_raise_out_of_memory();
+
   th->descr = caml_thread_new_descriptor(clos);
 
   th->next = Current_thread->next;
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -77,12 +77,12 @@ struct caml_thread_struct {
   value ** gc_regs_slot;     /* saved value of Caml_state->gc_regs_slot */
   void * exn_handler;        /* saved value of Caml_state->exn_handler */
 
-  #ifndef NATIVE_CODE
+#ifndef NATIVE_CODE
   intnat trap_sp_off;      /* saved value of Caml_state->trap_sp_off */
   intnat trap_barrier_off; /* saved value of Caml_state->trap_barrier_off */
   /* saved value of Caml_state->external_raise */
   struct caml_exception_context* external_raise;
-  #endif
+#endif
 
 };
 
@@ -180,11 +180,11 @@ void caml_thread_save_runtime_state(void)
   Current_thread->backtrace_pos = Caml_state->backtrace_pos;
   Current_thread->backtrace_buffer = Caml_state->backtrace_buffer;
   Current_thread->backtrace_last_exn = Caml_state->backtrace_last_exn;
-  #ifndef NATIVE_CODE
+#ifndef NATIVE_CODE
   Current_thread->trap_sp_off = Caml_state->trap_sp_off;
   Current_thread->trap_barrier_off = Caml_state->trap_barrier_off;
   Current_thread->external_raise = Caml_state->external_raise;
-  #endif
+#endif
 }
 
 void caml_thread_restore_runtime_state(void)
@@ -199,11 +199,11 @@ void caml_thread_restore_runtime_state(void)
   Caml_state->backtrace_pos = Current_thread->backtrace_pos;
   Caml_state->backtrace_buffer = Current_thread->backtrace_buffer;
   Caml_state->backtrace_last_exn = Current_thread->backtrace_last_exn;
-  #ifndef NATIVE_CODE
+#ifndef NATIVE_CODE
   Caml_state->trap_sp_off = Current_thread->trap_sp_off;
   Caml_state->trap_barrier_off = Current_thread->trap_barrier_off;
   Caml_state->external_raise = Current_thread->external_raise;
-  #endif
+#endif
 }
 
 /* Hooks for caml_enter_blocking_section and caml_leave_blocking_section */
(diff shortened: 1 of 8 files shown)


otherlibs/systhreads/st_stubs.c
```diff
@@ -376,13 +374,16 @@ static void caml_thread_initialize_domain()
   new_thread =
     (caml_thread_t) caml_stat_alloc_noexc(sizeof(struct caml_thread_struct));
 
+  if (new_thread == NULL)
+    caml_raise_out_of_memory();
```
ctk21: I don't think you can raise when `caml_thread_initialize_domain` is called from `caml_domain_start_hook`.
Or is there an argument where that is safe?

ctk21 (commented):
I think this does what is intended.

I need convincing that the raising of oom is safe when it is done via `caml_domain_start_hook`

otherlibs/systhreads/st_stubs.c
```diff
@@ -397,12 +397,12 @@ CAMLprim value caml_thread_yield(value unit);
...
 void caml_thread_interrupt_hook(void)
 {
+  uintnat is_on = 1;
   caml_domain_state *domain = Caml_state;
   atomic_uintnat* req_external_interrupt =
     (atomic_uintnat*)&domain->requested_external_interrupt;
 
-  if (atomic_load_acq(req_external_interrupt) == 1) {
-    atomic_store_rel(req_external_interrupt, 0);
+  if (atomic_compare_exchange_strong(req_external_interrupt, &is_on, 0)) {
```
xavierleroy, line 400: I was expecting `atomic_exchange`. The proposed code should also work, even though it's less obvious.

otherlibs/systhreads/st_stubs.c
xavierleroy: If it was safe, we could just use `caml_stat_alloc`, which is equivalent to `caml_stat_alloc_noexc` + call to `caml_raise_out_of_memory` if NULL...

I'm not convinced it is safe to raise exceptions here.  This said, the OCaml 4 implementation of systhreads calls `caml_stat_alloc` (which can raise) in similar places in st_stubs.c, so it's fishy in OCaml 4 too.

abbysmal: Does it sound reasonable to call `caml_fatal_error` in such circumstances then?
I am not a huge fan of raising `oom` in the event of the program actually trying to recover from it, and I remember the past literature surrounding the question of whether `out_of_memory` should be fatal or not to be quite dense.
I'm not super sure what would be the right call, and I'll proceed with `caml_fatal_error` if it sounds reasonable.

gadmm: > quite dense

It is.

It is reasonable to introduce a “small allocation rule” that says that below a certain size, it is assumed that an allocation never fails, and allow the program to crash if it does. (Any allocation from OCaml small enough to go on the minor heap (<256 words) can crash the program anyway if it triggers a GC that in turn tries and fails to extend the major heap, and this is reportedly hard to fix.)

Having looked closely at the problem of asynchronous exceptions in the runtime, it is best if the OOM exception is reserved for large (implicit or explicit) allocations (or whose size depends on user input), and turned into a synchronous exception. (The runtime does not subscribe to this rule currently, but it also has a lot of theoretical resource leaks due to the OOM exceptions that are going to be very difficult to fix otherwise.)

(Users who want to limit their memory usage can do so differently, such as subscribing to an asynchronous event signalling memory pressure using cgroups, and converting this event into a proper asynchronous exception, which has the advantage of also working on overcommitting systems.)

`caml_fatal_error` is used for OOM in other places but it now does `abort()` by default, which was reported not to be the best way to react to an allocation failure. But that's a separate problem.

So yes, it is reasonable to call `caml_fatal_error` here.

xavierleroy: I'm not happy with the caml_fatal_error solution, as I think we can report a proper `Out_of_memory` exception in this case.  I'll explain my idea next.

xavierleroy:
So, after reading and grepping through the whole code, I have a better understanding of how `caml_thread_initialize_domain` is used.

Currently, it's called from `domain_thread_func` in runtime/domain.c via `caml_domain_start_hook`.  It's called at a point where a failure cannot be handled.  I.e. `caml_domain_start_hook` could return an error code, but on seeing this error code, `domain_thread_func` would have to skip the callback into OCaml and do something, e.g. return the `Error Out_of_memory` value that the OCaml code would have returned had it encountered an OOM.  But the OCaml code does other mandatory work such as signalling domain termination to other domains, so it's not an option to skip the callback.

My proposal is to call `caml_thread_initialize_domain` from OCaml, more precisely in the `body` local function of `Domain.spawn`.  Then, `caml_thread_initialize_domain` can raise the `Out_of_memory` exception, and it will be handled gracefully.

To implement this cleanly, I'd suggest adding a "at start" facility to domains, just like there is an "at exit" facility and an "at first spawn" facility already.  Then the `body` local function of `Domain.spawn` looks like
```
  let body () =
    let result = match DLS.create_dls (); do_at_start(); f () with ...
```

Then, the systhreads library would register the external function `caml_thread_initialize_domain` as an "at start" function.

Note that the other use of `caml_thread_initialize_domain` is in `caml_thread_initialize`, which can safely raise exceptions.

xavierleroy:
While reading st_stubs.c, I found another potential OOM issue, but this one looks easier to fix.
https://github.com/ocaml-multicore/ocaml-multicore/blob/b5905dfd0037c1a4d2cc5a18ff41998d12c5b570/otherlibs/systhreads/st_stubs.c#L535-L536
`caml_thread_new_info` can return NULL, so we need to check for that and call `caml_raise_out_of_memory()`.  (It's safe to raise here.)

https://github.com/ocaml-multicore/ocaml-multicore/blob/b5905dfd0037c1a4d2cc5a18ff41998d12c5b570/otherlibs/systhreads/st_stubs.c#L255-L256
This is where `caml_thread_new_info` can return NULL, and it's OK.  However, a few lines down:

https://github.com/ocaml-multicore/ocaml-multicore/blob/b5905dfd0037c1a4d2cc5a18ff41998d12c5b570/otherlibs/systhreads/st_stubs.c#L261
`caml_alloc_main_stack` can return NULL on OOM, so this should be tested and NULL should be returned after having freed `th`.

abbysmal:
> To implement this cleanly, I'd suggest adding a "at start" facility to domains, just like there is an "at exit" facility and an "at first spawn" facility already.

I think this sounds sensible, and allow for a more graceful handling of such errors.
It is indeed fine to run in `caml_thread_initialize` since this code is already called from the `Thread` module toplevel.

I will work on implementing the proposed change, as well as addressing the other OOM issue you spotted.

Thanks!

abbysmal:
I rebased and pushed the suggested changes.
`Domain.at_startup` has been introduced to handle the domain's systhreads initialization code.
I also addressed the issue around `current_stack`'s allocation, thank you for spotting this.

Side note: I think with `Domain.at_startup`, and systhreads relying on it, we may actually remove the runtime C-side equivalent.
It was introduced to support systhreads in the first place.
I have no strong feeling on this point.

otherlibs/systhreads/st_stubs.c
xavierleroy: Now that it is safe to raise from `caml_thread_initialize_domain`, you can shorten this to
```
  new_thread =
    (caml_thread_t) caml_stat_alloc(sizeof(struct caml_thread_struct));
```

xavierleroy (approved):
Looks good to me.  Thanks!  A suggestion for a tiny simplification below.

abbysmal:
Indeed, thanks!

PR #752, opened 2021-11-22
document the current Multicore testsuite situation

### LINKED ISSUE #751: how does one run the Multicore testsuite?
I'm trying to review the Multicore tree, and noticed that running the testsuite appears to not work.

My process to run the testsuite is basically as follows:

```
make world.opt
cd testsuite
make parallel
```

It looks like many tests are failing on my machine, and one test appears to loop infinitely (the testsuite uses a timeout of 600 seconds by default, which is... rather long).

Non-terminating test (from the `testsuite` directory):

```
../ocamltest/ocamltest.opt -timeout 10 tests/tool-debugger/basic/debuggee.ml
```

List of failed tests -- only the tests that could be tried before I aborted the testsuite run due to non-termination.

```
List of failed tests:
    tests/statmemprof/'alloc_counts.ml' with 2 (bytecode) 
    tests/exotic-syntax/'exotic.ml' with 1 (native) 
    tests/statmemprof/'custom.ml' with 2 (bytecode) 
    tests/misc/'sorts.ml' with 1 (native) 
    tests/misc/'hamming.ml' with 1 (native) 
    tests/statmemprof/'exception_callback_minor.ml' with 2 (bytecode) 
    tests/statmemprof/'blocking_in_callback.ml' with 1.2 (native) 
    tests/lazy/'lazy3.ml' with 1 (native) 
    tests/statmemprof/'thread_exit_in_callback.ml' with 1.2 (native) 
    tests/statmemprof/'moved_while_blocking.ml' with 1.2 (native) 
    tests/match-exception/'streams.ml' with 1 (native) 
    tests/ephe-c-api/'test.ml' with 2 (bytecode) 
    tests/compatibility/'main.ml' with 2.2 (native) 
    tests/basic-modules/'recursive_module_init.ml' with 1 (native) 
    tests/basic/'patmatch.ml' with 1 (native) 
    tests/statmemprof/'arrays_in_minor.ml' with 1 (native) 
    tests/statmemprof/'lists_in_minor.ml' with 1 (native) 
    tests/basic-more/'robustmatch.ml' with 1 (native) 
    tests/statmemprof/'comballoc.ml' with 1 (bytecode) 
    tests/statmemprof/'comballoc.ml' with 2 (native) 
    tests/runtime-naked-pointers/'np1.ml' with 2 (native) 
    tests/statmemprof/'thread_exit_in_callback.ml' with 1.1 (bytecode) 
    tests/basic-more/'pr10338.ml' with 1 (native) 
    tests/lazy/'lazy4.ml' with 1 (native) 
    tests/statmemprof/'callstacks.ml' with 1.2 (bytecode) 
    tests/statmemprof/'minor_no_postpone.ml' with 2 (bytecode) 
    tests/lib-string/'binary.ml' with 1 (native) 
    tests/lazy/'minor_major_force.ml' with 1 (native) 
    tests/lazy/'lazy1.ml' with 1 (native) 
    tests/statmemprof/'moved_while_blocking.ml' with 1.1 (bytecode) 
    tests/statmemprof/'arrays_in_minor.ml' with 2 (bytecode) 
    tests/ephe-c-api/'test.ml' with 1 (native) 
    tests/lib-bytes/'binary.ml' with 1 (native) 
    tests/statmemprof/'intern.ml' with 2 (bytecode) 
    tests/statmemprof/'exception_callback.ml' with 2 (bytecode) 
    tests/statmemprof/'alloc_counts.ml' with 1 (native) 
    tests/statmemprof/'intern.ml' with 1 (native) 
    tests/backtrace/'backtrace2.ml' with 1 (native) 
    tests/backtrace/'lazy.ml' with 1 (native) 
    tests/compatibility/'main.ml' with 1.2 (native) 
    tests/runtime-naked-pointers/'np2.ml' with 2 (native) 
    tests/statmemprof/'callstacks.ml' with 1.1 (native) 
    tests/compatibility/'main.ml' with 2.1 (bytecode) 
    tests/effects/'test_lazy.ml' with 1 (native) 
    tests/statmemprof/'exception_callback_minor.ml' with 1 (native) 
    tests/lazy/'lazy2.ml' with 1 (native) 
    tests/statmemprof/'lists_in_minor.ml' with 2 (bytecode) 
    tests/letrec-compilation/'lazy_.ml' with 1 (native) 
    tests/statmemprof/'exception_callback.ml' with 1 (native) 
    tests/runtime-naked-pointers/'np1.ml' with 1 (bytecode) 
    tests/statmemprof/'blocking_in_callback.ml' with 1.1 (bytecode) 
    tests/compatibility/'main.ml' with 1.1 (bytecode) 
    tests/statmemprof/'arrays_in_major.ml' with 1 (native) 
    tests/statmemprof/'minor_no_postpone.ml' with 1 (native) 
    tests/runtime-naked-pointers/'np2.ml' with 1 (bytecode) 
    tests/statmemprof/'arrays_in_major.ml' with 2 (bytecode) 
    tests/lazy/'lazy6.ml' with 1 (native) 
    tests/statmemprof/'custom.ml' with 1 (native) 
```

Is my machine misconfigured? Are those failures intended? What is the recommended way to test the multicore branches?

gasche:
(closes #751)

With this PR the documentation is up-to-date, and calling `make all` or `make parallel` shows an error message.

Note that `make all-foo` (which runs all tests starting with `foo`) is renamed to `make really-all-foo`. The `all-enabled` naming makes the previous pattern unfortunate anyway.

### DIFF
--- a/testsuite/Makefile
+++ b/testsuite/Makefile
@@ -133,12 +133,12 @@ endif
 .PHONY: default
 default:
 	@echo "Available targets:"
-	@echo "  all             launch all tests"
-	@echo "  all-foo         launch all tests beginning with foo"
-	@echo "  all-enabled     launch all enabled tests"
-	@echo "  parallel        launch all tests using GNU parallel"
-	@echo "  parallel-foo    launch all tests beginning with foo using \
-	GNU parallel"
+	@echo "  all-enabled     launch all Multicore-enabled tests"
+	@echo "  really-all      launch all tests, including known-broken ones"
+	@echo "  really-all-foo  launch all tests beginning with foo, including known-broken ones"
+	@echo "  really-parallel     launch all tests using GNU parallel; this is broken with Multicore"
+	@echo "  really-parallel-foo launch all tests beginning with foo using \
+	GNU parallel; this is broken with Multicore"
 	@echo "  one TEST=f      launch just this single test"
 	@echo "  one DIR=p       launch the tests located in path p"
 	@echo "  one LIST=f      launch the tests listed in f (one per line)"
@@ -148,12 +148,17 @@ default:
 	@echo "  clean           delete generated files"
 	@echo "  report          print the report for the last execution"
 	@echo
-	@echo "all*, parallel* and list can automatically re-run failed test"
-	@echo "directories if MAX_TESTSUITE_DIR_RETRIES permits"
-	@echo "(default value = $(MAX_TESTSUITE_DIR_RETRIES))"
+	#@echo "all*, parallel* and list can automatically re-run failed test"
+	#@echo "directories if MAX_TESTSUITE_DIR_RETRIES permits"
+	#@echo "(default value = $(MAX_TESTSUITE_DIR_RETRIES))"
 
 .PHONY: all
 all:
+	@echo "'make all' is not supported on Multicore OCaml"
+	@echo "Please call 'make' to see the appropriate Multicore targets"
+
+.PHONY: really-all
+really-all:
 	@$(MAKE) --no-print-directory new-without-report
 	@$(MAKE) --no-print-directory report
 
@@ -188,7 +193,12 @@ all-enabled: lib tools
 	@$(MAKE) report
 
 .PHONY: all-%
-all-%: lib tools
+all-%:
+	@echo "'make all-*' targets are not supported on Multicore OCaml"
+	@echo "Please call 'make' to see the appropriate Multicore targets"
+
+.PHONY: really-all-%
+really-all-%: lib tools
 	@for dir in tests/$**; do \
 	  $(MAKE) --no-print-directory exec-one DIR=$$dir; \
 	done 2>&1 | tee $(TESTLOG)
@@ -219,7 +229,12 @@ all-%: lib tools
 # but the demangling separation is arguably nicer behavior that we might
 # want to implement at the exec-one level to also have it in the 'all' target.
 .PHONY: parallel-%
-parallel-%: lib tools
+parallel-%:
+	@echo "'make parallel-*' targets are not supported on Multicore OCaml"
+	@echo "Please call 'make' to see the appropriate Multicore targets"
+
+.PHONY: really-parallel-%
+really-parallel-%: lib tools
 	@echo | parallel >/dev/null 2>/dev/null \
 	 || (echo "Unable to run the GNU parallel tool;";\
 	     echo "You should install it before using the parallel* targets.";\
@@ -235,7 +250,12 @@ parallel-%: lib tools
 	@$(MAKE) report
 
 .PHONY: parallel
-parallel: parallel-*
+parallel:
+	@echo "'make parallel' targets are not supported on Multicore OCaml"
+	@echo "Please call 'make' to see the appropriate Multicore targets"
+
+.PHONY: really-parallel
+really-parallel: really-parallel-*
 
 .PHONY: list
 list: lib tools
--- a/testsuite/HACKING.adoc
+++ b/testsuite/HACKING.adoc
@@ -4,14 +4,10 @@
 
 == Useful Makefile targets
 
-`make parallel`::
-  runs the tests in parallel using the
-  link:https://www.gnu.org/software/parallel/[GNU parallel] tool: tests run
-  twice as fast with no difference in output order.
+Note: Multicore OCaml's testsuite is in flux, the upstream-OCaml test targets do not work.
 
-`make all-foo`, `make parallel-foo`::
-  runs only the tests in the directories whose name starts with `foo`:
-  `parallel-typing`, `all-lib`, etc.
+`make all-enabled`::
+  calls the whole testsuite (see `make` default output for other less-working options)
 
 `make one DIR=tests/foo`::
   runs only the tests in the directory `tests/foo`. This is often equivalent to


testsuite/Makefile
```diff
@@ -218,8 +228,13 @@ all-%: lib tools
 # I chose to keep the previous behavior exactly unchanged,
 # but the demangling separation is arguably nicer behavior that we might
 # want to implement at the exec-one level to also have it in the 'all' target.
-.PHONY: parallel-%
-parallel-%: lib tools
+.PHONY: paralell-%
```
abbysmal: A few of the same typos in the PR where `parallel` becomes `paralell`

testsuite/Makefile
```diff
@@ -218,8 +228,13 @@ all-%: lib tools
...
+paralell-%:
+	@echo "'make paralell-*' targets are not supported on Multicore OCaml"
```
abbysmal: ditto

testsuite/Makefile
```diff
@@ -234,8 +249,13 @@ parallel-%: lib tools
 	 | tee $(TESTLOG)
 	@$(MAKE) report
 
-.PHONY: parallel
-parallel: parallel-*
+.PHONY: paralell
```
abbysmal: ditto

testsuite/Makefile
```diff
@@ -234,8 +249,13 @@ parallel-%: lib tools
...
+paralell:
```
abbysmal: Here as well

abbysmal (changes requested):
Thank you for your PR!
This looks fine to me but is buggy at the moment because of a few typos. :)

I think though we should discuss how to address individually some of the outstanding issues.
Specifically the `parallel` target, which is hardly usable on Multicore because of the `parallel` testsuite.
Maybe the testsuite runner should be fed information about what tests directories are sequential and which one may execute multiple domains.
Going further, being able parametrize the multi-domains testcases would likely help as well relieve some pain in that area.

This is likely not the right place to discuss this, we should open a follow up issue after this one in order to make sure that we try to address a few of these at some point.

gasche:
Thanks for the feedback! I fixed the silly typos.

Re. `parallel`: a simple approach would to write a list of test directories that must not run in parallel, and have `make parallel` run only the other jobs in parallel, and those jobs in sequence after that (or before that). (The information could also be stored in each directory with more elaborate scripting.)

This being said, I think that the bigger-picture task is to cleanup the testsuite in the Multicore tree. If you want to start sending PRs to the upstream tree, you will need to send proper testsuite changes, and it's probably easier overall to work upfront to reduce the diff by cleaning up the testsuite in your tree.

abbysmal (approved):

abbysmal:
Thank you @gasche!

Agreed on having a way to distinguish sequential and parallel test directories. An alternative I'm thinking about could be to have a file (rather than storing information in individual directories) that describes what are the parallel test directories. (a bit like the `disabled` file in the Multicore tree handles disabled testcases.)
This would also lower the amount of scripting required.

I do agree the bigger picture involves some cleaning:
- A quick glance at the diff shows that some tests have non-required differences, I assume some changes that were lost in some of our rebases. These needs to be addressed.
- Do we want to keep the `disabled/enabled` mechanism? @ctk21 mentioned that the `skip` mechanism of `ocamltest` could be used instead of the disabled mechanism.
- @dra27 has some fixes for CI in #640 (I believe), including fixes for `parallel`. We should also integrate these.

Is there anything else that comes to your mind about the bigger picture for the testsuite story for 5.00?
Maybe we should devise proper RFCs to get a better view of the outstanding issues.

It does also seems that `CI` and `testsuite` are two separate concerns, and the the CI story for Multicore will need to be also a bit more clearly defined.
This stresses the previous point that we may want to submit a proposal/RFC for both CI and testsuite and make sure all concerns and outstanding issues are addressed and talked through properly.

gasche:
Note: I am not sure that you actually *need* specific care to avoid running certain tests in parallel. There are stress-tests that spawn multiple domains, but what if two of those tests run in parallel? At worst you are going to see less-deterministic scheduling, but that doesn't sound bad for testing, in fact possibly a plus. So I would first try to get `make parallel` to work on the Multicore tree (without forcing specific tests to be run alone), and see what happens.

> Do we want to keep the disabled/enabled mechanism?

It looks like a hack to me that you want to remove in favor of some upstream-supported way of skipping, indeed. 

If for some reason you found ocamltest not expressive enough for some testsuite features you would need, we can always improve ocamltest -- I would recommend opening an upstream issue, or hacking on ocamltest directly in the multicore repo if you feel like it.

> @dra27 has some fixes for CI in GitHub Actions for Windows #640 (I believe), including fixes for parallel. We should also integrate these.

Yes, I recommend integrating every @dra27 proposes for the testsuite :-) (And the present PR!)

> Is there anything else that comes to your mind about the bigger picture for the testsuite story for 5.00?

I think that the best would be for the multicore-ocaml testsuite to have a reasonable diff wrt. the trunk/4.14 testsuite by the time of the Multicore Merge, rather than 5.00. If you wait for later, potential problems include:
- if the Merge preserves the trunk testsuite, the tests may break due to tests you haven't seen in the multicore tree (we might even miss bugs if we promote test outputs without being careful)
- if the Merge does not include Multicore-specific tests, we end up with much weaker testsuite coverage of the concurrency features
- if the Merge moves the needle in the other direction of replacing the trunk testsuite by the multicore testsuite, we end up in a semi-broken testsuite state and developers/contributors/maintainers are going to complain (rightfully).

ctk21:
> > Do we want to keep the disabled/enabled mechanism?
> 
> It looks like a hack to me that you want to remove in favor of some upstream-supported way of skipping, indeed.
> 
> If for some reason you found ocamltest not expressive enough for some testsuite features you would need, we can always improve ocamltest -- I would recommend opening an upstream issue, or hacking on ocamltest directly in the multicore repo if you feel like it.
> 

Multicore introduced `testsuite/disabled` probably before ocamltest existed. Over time the number of tests needing to be disabled has fallen, hopefully it will fall further. 

My take is: ocamltest now has the `skip` feature and we should use it and remove the `testsuite/disabled` file.

gasche:
I was curious, so I checked: indeed the first ocamltest commit was in July 2017 ( 635b5edcc1a0ce40d974295a696fbb3dea2f0766 ), while `testsuite/disabled` was added in March 2017 ( 1148920a8ad8cae9f06e5d5d1c778a43584f3440 ) :-)

kayceesrk:
We now have https://github.com/ocaml-multicore/ocaml-multicore/pull/774 which uses ocamltest `skip` mechanism. @moyodiallo confirmed that `make parallel` also works.

ctk21:
I'm closing this PR, because the multicore testsuite now runs in the same way as the ocaml/ocaml.

PR #756, opened 2021-11-23
RFC: generalize the Domain.DLS interface to split PRNG state for child domains

gasche:
This is a "RFC" PR, intended to demonstrate an approach to implement a "proper" PRNG+Domains semantics where spawning a domain "splits" the PRNG state. This required changes to the Domain.DLS interface to support domain-local values that are "inherited" from the parent/spawning domain, with a user-provided function to derive the child state from the parent state.

### DIFF
--- a/testsuite/tests/lib-random/parallel.ml
+++ b/testsuite/tests/lib-random/parallel.ml
@@ -7,8 +7,14 @@ let f () =
   let c = Random.int 100 in
   (a, b, c)
 
-let domains = List.init 10 (fun _ -> Domain.spawn f)
+(* We intentionally spawn many more domains than hardware threads, to
+   ensure that we see computations scheduled in a a non-deterministic
+   order. With a small number of domains, it's hard to reliably
+   observe non-determinism. *)
+let domain_count = 1000
+let domains = List.init domain_count (fun _i -> Domain.spawn f)
 let results = List.map Domain.join domains
+let results = List.filteri (fun i _ -> i mod (domain_count / 10) = 0) results
 
 let () =
   results |> List.iter (fun (a, b, c) ->
--- a/testsuite/tests/lib-random/parallel.ml
+++ b/testsuite/tests/lib-random/parallel.ml
@@ -17,5 +17,5 @@ let () =
 
 let () =
   print_endline
-    "Note: the result of this test is currently BROKEN,\n\
-     the random numbers produced by child domain are highly correlated."
+    "Note: we observe in this output that the random numbers of each child domain\n\
+     appear uncorrelated, yet are produced deterministically."
--- a/testsuite/tests/lib-random/parallel.ml
+++ b/testsuite/tests/lib-random/parallel.ml
@@ -1,0 +1,21 @@
+(* TEST *)
+let () = Random.init 42
+
+let f () =
+  let a = Random.int 100 in
+  let b = Random.int 100 in
+  let c = Random.int 100 in
+  (a, b, c)
+
+let domains = List.init 10 (fun _ -> Domain.spawn f)
+let results = List.map Domain.join domains
+
+let () =
+  results |> List.iter (fun (a, b, c) ->
+    Printf.printf "%d %d %d\n%!" a b c
+  )
+
+let () =
+  print_endline
+    "Note: the result of this test is currently BROKEN,\n\
+     the random numbers produced by child domain are highly correlated."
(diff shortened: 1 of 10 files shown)


gasche:
In this version, some keys are "inherited" (when a `split_from_parent` argument is passed) and some are "local" (not inherited). The behavior of inherited and local keys differ:
- the value of inherited keys is computed *eagerly* when spawning the domain
- the value of local keys is computed *lazily* on the first `DLS.get`

One could implement lazy-initialization for inherited keys as well, with a slightly more complex implementation.

Note: there are two parts to the computation of an inherited key, a computation done by the parent domain and a computation done by the child domain. The parent domain computation can only done eagerly, only the child domain could be made lazy. (Trying to do the parent-domain computation when the child calls `DLS.get` would require synchronization or create races.)

Note: for PRNG splitting there is no difference in behavior between "eager" and "lazy" inheritance: the effectul part of the computation is done by the parent right before spawning, and the part of computation that is delayed is pure.

There is a performance and resource-usage benefit to lazy inheritance (if we spawn a domain that uses almost none of the parent keys), but this comes with a small constant overhead on `DLS.get`. It is also easy to implement on the user side, by building a `'a Lazy.t` key that is forced outside the DLS module.

(Note: there is no clear way to make local (non-inherited) keys "eager". In particular, it would not work to compute the value when `new_key` is called, that would be eager for the current domain but child domains would still wait for `get`. But we could make everything eager by making all keys inherited, with a trivial "re-initialize in the child" implementation for currently-local keys.)

gasche:
I pushed a new commit to make the computation on the child lazy for inherited keys. This incurs a small overhead for all calls to `DLS.get` (the value is stored in a `'a Lazy.t` instead of `'a`), but the behavior is more consistent with non-inherited keys, and the implementation is in fact simpler / more regular.

gasche:
@xavierleroy has done a pass to simplify the API and implementation, and we are both happy with the current result.

gasche:
The simplifications are as follows:

- We go back to the "strict" semantics again; in fact, the new value for a spawning domain is now computed solely on the parent domain, without delegating computation to the child domain.
- The implementation maintains a global list of inherited keys, instead tracking per-domain information; after an inherited key is created, all domains spawned in the future will compute a key, not just the descendants of domains that used the key.

testsuite/tests/lib-random/parallel.ml
```diff
@@ -0,0 +1,27 @@
...
+
+let f () =
+  let a = Random.int 100 in
+  let b = Random.int 100 in
+  let c = Random.int 100 in
+  (a, b, c)
+
+(* We intentionally spawn many more domains than hardware threads, to
+   ensure that we see scheduling effects that make the computations
+   are scheduled in a non-deterministic order. With a small number of
```
Octachron: The `are` is probably a typo?
Maybe just
```
that we see computations scheduled in a a non-deterministic order.
```
?

gasche: Thanks, fixed.

gasche:
I have a commit (not pushed in this PR's branch) that changes the global list to also store the key initialization function, so that the "key" passed around can be just the integer index: https://github.com/gasche/ocaml-multicore/commit/8ee8fc2e545b2f545cf9bee7b59e2af0e41814a9 (from https://github.com/gasche/ocaml-multicore/commits/multicore-random-variant ). @xavierleroy if this was to be done from scratch, which design would you prefer?

gasche:
I would be interested in a review of the Multicore devs who worked on DLS (one of @kayceesrk, @ctk21, @Sudha247 maybe?).

xavierleroy:
> I have a commit (not pushed in this PR's branch) that changes the global list to also store the key initialization function, so that the "key" passed around can be just the integer index:

I think I prefer the current "object-oriented" representation of keys as a (index, init function) pair, for the following not-very-strong reasons:
- Using a map makes computing the initial value of a key (in `get` and in `get_initial_keys`) a bit more expensive.
- It makes Maps a dependency of Domain.  What if Maps need Domain.DLS at some future point?
- Initialization functions can no longer be garbage-collected when the corresponding key becomes unreachable.  I know this should rarely happen, and I know that "split from parent" functions are never garbage-collected in my design, but still...

gasche:
Re-submitted as https://github.com/ocaml/ocaml/pull/10887 . Closing here.

Issue #750, opened 2021-11-21
Discussing the design of Lazy under Multicore 


gasche:
This issue is collecting my understanding of the design of Lazy under Multicore.

## Sequential Lazy

With Sequential OCaml, trying to force a value that is currently being forced raises Undefined. This can be caused by a thunk trying to force itself recursively, or two systhreads racing to force a value.

This is implemented by mutating the lacy block twice during forcing, once to replace the thunk closure with `fun () -> raise Undefined` before computation (the tag remains Lazy_tag), and once to set the result (the tag remains Lazy_tag in case of an exception, and becomes Forward_tag otherwise).

(Note: there is no allocation/yield between setting the thunk to `fun () -> raise Undefined` and invoking the thunk closure, so if two systhreads race to force a value, exactly one of them will get the real thunk closure.)

## Concurrency problems

With Multicore OCaml, synchronization was introduced to avoid two potential issues:
- duplicated computations: if two domains race to force a not-yet-computed thunk, they could both invoke the thunk closure
- memory unsafety: if a domain mutates a thunk with a computed value, racing with a domain that forces the thunk, the forcing domain could try to invoke the returned-value as a not-yet-computed-thunk, which is unsound

(This is in contrast with other mutable-state libraries like Stack where lack of synchronization can break programs but not break memory-safety.)

Note: it is clear that "memory unsafety" is a blocking issue, and that solving it requires some synchronization. However, whether "duplicated computations" is a blocking issue or not is a design choice, we could have decided that it is okay to have two domains racing. This being said, I don't think that relaxing this design constraint would lead to a substantially simpler implementation, as we need synchronization anyway to avoid memory unsafety.

## Current approach (November 2021)

The solution used in Multicore OCaml is basically to treat Lazy_tag values as atomic references, with the state indicated in the tag:
- Lazy_tag indicates (1) a thunk that is not yet forced, *or* (2) a thunk that was forced in the past and raised an exception, or whose forcing failed with `Undefiend`. The block argument stores a function closure that contains the thunk computation, or is of the form `fun () -> raise e` case (2).
- Forcing_tag indicates a thunk that is in the process of being forced by a thread. The block argument stores a marker of the forcing domain, to distinguish "recursive forcing" from "concurrent forcing by another domain".
- Forward_tag or any other tag is used for thunks that have been computed to a value (not an exception).

[...]

Issue #749, opened 2021-11-20
Potential bug on Forward_tag short-circuiting?


gasche:
## Context

When the minor GC sees a `Forward_tag` block, it can short-circuit the block by rewriting it to hold the value. (Forward_tag denotes a lazy value whose thunk has already been successfully forced; short-circuiting makes the memory representation more compact.)

Short-circuiting must be careful to not introduce confusions: it is only valid for values that cannot be mistaken with lazy thunks, and (if flat-float-array) it cannot introduce `float` values in the type of lazy values. To avoid confusions, short-circuiting is thus disabled on values of type `Forward_tag`, `Lazy_tag` and `Double_tag`. See the [minor_gc.c code](https://github.com/ocaml-multicore/ocaml-multicore/blob/4469f89/runtime/minor_gc.c#L363-L381):

## Potential bug

Multicore introduces a new tag for lazy thunks, `Obj.forcing_tag`, for values that are in the process of being forced (possibly concurrently). I believe that it is incorrect to shortcut `Forward_tag` on values of type `Obj.forcing_tag`, due to the following scenario:

```ocaml
let inner_thunk = lazy ...
let outer_thunk = lazy (some_allocating_code; inner_thunk)

let () = Lazy.force outer_thunk
  (* outer_thunk is now a Forward block to inner_thunk *)

let () = Domain.spawn (fun () ->
  ignore (Lazy.force inner_thunk)
  (* sets `inner_thunk` to Forcing_tag, then pause with a GC that short-circuits `outer_thunk` *))
let () = Domain.spawn (fun () ->
  ignore (Lazy.force outer_thunk)
  (* observes Forcing_tag as the tag of the short-circuited outer_thunk, and fails *) )
```

gasche:
The same issue may also exist in `Lazy.from_val`:

```ocaml
let from_val (v : 'arg) =
  let t = Obj.tag (Obj.repr v) in
  if t = Obj.forward_tag || t = Obj.lazy_tag || t = Obj.double_tag then begin
    make_forward v
  end else begin
    (Obj.magic v : 'arg t)
  end
```

I believe that `|| t = Obj.forcing_tag` should also be added there for similar reasons.

kayceesrk:
This is a good catch. Both of these cases need to be fixed.

kayceesrk:
Fixed by #760.

PR #760, opened 2021-11-29
Simplify lazy semantics

kayceesrk:
Simplify lazy and fix a bug based on discussions from https://github.com/ocaml-multicore/ocaml-multicore/issues/749 and https://github.com/ocaml-multicore/ocaml-multicore/issues/750. 

* Mark lazy as not concurrency (domains, systhreads, fibers) safe.
* Remove RacyLazy exception. Any concurrent use of lazy value may raise Undefined exception.
* Remove domain-local id to distinguish concurrent and recursive forcing.
* Remove try_force.

The commit also fixes a bug to prevent short-circuiting forcing_tag values.

We need to bootstrap to remove references to `RacyLazy` from the bootstrap binaries. After bootstrap, the primitive
`caml_ml_domain_unique_token` should be removed from the runtime.


kayceesrk:
@gasche would appreciate your review.

runtime/domain.c
```diff
@@ -817,8 +807,7 @@ CAMLprim value caml_ml_domain_id(value unit)
 
 CAMLprim value caml_ml_domain_unique_token (value unit)
 {
-  CAMLnoalloc;
-  return Caml_state->unique_token_root;
+  return Val_unit;
```
gasche, line 806: I suppose that the function here is left around to avoid having to do a bootstrap?

runtime/obj.c
```diff
@@ -259,39 +260,25 @@ CAMLprim value caml_lazy_read_result (value v)
...
     } else {
-      CAMLassert (tag == Lazy_tag);
-      /* We may reach here if the tag of [v] has been reset to [Lazy_tag] using
-       * [caml_lazy_reset_to_lazy] since our own attempt at updating the tag of
-       * [v] in [obj_update_tag] failed. The compare and swap in
-       * [obj_update_tag] failed since it must have observed [Forcing_tag].
-       * This represents a racy update of the tag of [v] by another mutator. */
+      CAMLassert (tag == Forward_tag);
       return Val_int(2);
     }
```
gasche: The logic here is simpler, I don't think the change introduces any bug.

Two questions, unrelated to this PR but to the function itself:
- Do we gain anything by distinguish Forward from any other shortcut tag? I would have expected the first check to be just on `tag != Lazy_tag && tag != Forcing_tag`. (Things are different after the CAS, where we know that the GC has not run since the tag check, so we have fine-grained knowledge that shortcutting hasn't happened which justifies the more precise `CAMLassert(tag == Forcing_tag)`.)
- if the CAS fails, is it really the case that we can assume `tag != Lazy_tag`? I would think that a concurrently-forcing domain could still revert the tag to `Lazy` if the thunk fails with an exception.

runtime/sys.c
```diff
@@ -148,18 +148,21 @@ CAMLexport void caml_do_exit(int retcode)
...
-                      (intnat)s.forced_major_collections);
+        "forced_major_collections: %"ARCH_INTNAT_PRINTF_FORMAT"d\n",
+        (intnat)s.forced_major_collections);
       caml_gc_message(0x400, "heap_words: %"ARCH_INTNAT_PRINTF_FORMAT"d\n",
                       heap_words);
       caml_gc_message(0x400, "top_heap_words: %"ARCH_INTNAT_PRINTF_FORMAT"d\n",
                       top_heap_words);
-      caml_gc_message(0x400, "mean_space_overhead: %lf\n", caml_mean_space_overhead());
+      caml_gc_message(0x400, "mean_space_overhead: %lf\n",
+                      caml_mean_space_overhead());
```
gasche: I'm not sure why these changes are in the PR, maybe some typo-check fix or automated formatting. No problem with having them either.

runtime/minor_gc.c
```diff
@@ -337,7 +337,8 @@ static void oldify_one (void* st_v, value v, value *p)
       ft = Tag_val (get_header_val(f) == 0 ? Field(f, 0) : f);
     }
 
-    if (ft == Forward_tag || ft == Lazy_tag || ft == Double_tag) {
+    if (ft == Forward_tag || ft == Lazy_tag ||
+        ft == Forcing_tag || ft == Double_tag) {
```
gadmm, line 341: Nice catch! There is similar logic in `runtime/{extern,major_gc,weak}.c` which needs to be fixed in the same way.

stdlib/camlinternalLazy.ml
```diff
@@ -59,83 +54,73 @@ external reset_to_lazy : Obj.t -> unit = "caml_lazy_reset_to_lazy" [@@noalloc]
...
-  | _ ->
-      (* Other cases indicate racy access to lazy for the same reasons outlined
-       * in [force_gen_lazy_block]. *)
-      None
-
 (* [force_gen ~only_val:false] is not used, since [Lazy.force] is
    declared as a primitive whose code inlines the tag tests of its
    argument, except when afl instrumentation is turned on. *)
-
 let force_gen ~only_val (lzv : 'arg lazy_t) =
```
gadmm, line 83:
```suggestion
let force_gen ~only_val (lzv : 'arg lazy_t) =
  let lzv = Sys.opaque_identity lzv in
```
The above fix was introduced in trunk to prevent flambda from forwarding the tag information from a previous read (which can become invalid with short-circuiting). (See https://github.com/ocaml/ocaml/issues/7301 and https://github.com/ocaml/ocaml/issues/713.) Do you need something similar? Are there risks that similar issues happen with the new tag Forcing_tag? (Note: this can be addressed in a separate PR if needed.)

gasche (approved):
Thanks for the ping! (I don't get notifications for ocaml-multicore/)

The change looks correct to me. It does simplify the implementation.

The documentation is changed so that Lazy.force is "unspecified" (but safe) in the case of concurrent forcing. I think we could be more precise, and will try to make some proposals.

The PR comes with some changes in the testsuite, some for backward-compatibility, some (dividing iteration counts by 10 or removing some `cpu_relax()` call) that seem to come from performance considerations, I'm not sure which/why.

gasche:
... after thinking more about it, I think that the current non-documentation of concurrent-forcing behavior is okay.

stdlib/lazy.mli
```diff
@@ -16,82 +16,59 @@
...
-   Note: if the program is compiled with the [-rectypes] option,
-   ill-founded recursive definitions of the form [let rec x = lazy x]
-   or [let rec x = lazy(lazy(...(lazy x)))] are accepted by the type-checker
-   and lead, when forced, to ill-formed values that trigger infinite
-   loops in the garbage collector and other parts of the run-time system.
-   Without the [-rectypes] option, such ill-founded recursive definitions
-   are rejected by the type-checker.
+(** A value of type ['a Lazy.t] is a deferred computation, called a suspension,
+    that has a result of type ['a].  The special expression syntax [lazy
+    (expr)] makes a suspension of the computation of [expr], without computing
```
gadmm: Beware of line breaks inside `[]` code blocks, they do not display as you want in the documentation.

stdlib/camlinternalLazy.ml
```diff
@@ -59,83 +54,73 @@ external reset_to_lazy : Obj.t -> unit = "caml_lazy_reset_to_lazy" [@@noalloc]
...
   let x = Obj.repr lzv in
-  (* XXX No safe points start *)
   let t = Obj.tag x in
-  if t = Obj.forward_tag then (Obj.obj (Obj.field x 0) : 'arg)
-  (* XXX No safe points end *)
-  else if t <> Obj.lazy_tag && t <> Obj.forcing_tag then (Obj.obj x : 'arg)
+  (* START no safe points. If a GC occurs here, then the object [x] may be
```
gadmm, line 86: I think the boundary of the critical section was clearer before: if there is a polling after Obj.tag returns, then there is a bug.
I was looking into expressing this with the `[@poll error]` attribute (to be sure flambda understands this critical section), but it appears unsufficiently expressive. We need one of the extension that I suggested (or Leo's variant of it). The problem is already there in trunk, so I'll see if I have time to open an issue upstream about it.

stdlib/camlinternalLazy.ml
```diff
@@ -59,83 +54,73 @@ external reset_to_lazy : Obj.t -> unit = "caml_lazy_reset_to_lazy" [@@noalloc]
...
+     short-circuited, and getting the first field of [x] would get us the wrong
+     value. Luckily, the compiler does not insert GC safe points at this place,
+     so it is ok. *)
+  if t = Obj.forward_tag then
+    (Obj.obj (Obj.field x 0) : 'arg)
```
gadmm, line 92: I am trying to convince myself that this is correct according to OCaml's memory model, but I do not see what ensures that the field is seen as set despite the tag being seen as a forward_tag. Don't you need a load/load barrier before reading the field, or reading the tag with an acquire load, or something stronger? (To my knowledge, OCaml's memory model does not have relaxed atomics yet, so furthermore only a plain atomic read of the tag makes sense here yet.) What makes this correct (and the similar code inlined by lambda)?

gadmm (commented):
I have more questions about this code now that I understand it better. Some comments are not about behaviour introduced by this PR, feel free to address them later/separately.

stdlib/camlinternalLazy.ml
gasche: My understanding is that the memory model gives a message-passing semantics to non-atomic writes+reads that are synchronized through an atomic at another location.

In the operational semantics of the memory-model paper: if I write to an atomic location, the location's frontier is updated with my frontier, including non-atomic writes that I did before (in program order). Then when the other domain reads from this atomic location, its frontier is updated by my frontier, so it also sees my non-atomic writes to other locations.

runtime/obj.c
gadmm: I think you are right (the assertion `tag == Forward_tag` can fail), but this is related to this PR.

stdlib/camlinternalLazy.ml
gadmm: Sure, but for this you need an atomic read as I wrote

stdlib/camlinternalLazy.ml
kayceesrk: I'll merge this suggestion. And I think the rest of the discussion may happen at ocaml/ocaml after the MVP merge to trunk, before the 5.00 release. The MVP will not have flambda support. If the discussion and the PR happens after merge, we will have more eyes on the flambda support as a whole.

runtime/domain.c
kayceesrk: Yes. That's right. Once the changes are ok. I'll do a bootstrap commit and remove this definition.

runtime/obj.c
kayceesrk: Firstly, my overall idea with this function is to leave some of the functionality intact even though `camlinternalLazy.ml` only cares about `0` and `not 0` cases. The idea is that future implementation of concurrency safe lazy may need to distinguish the various cases. Hence, my choice was to leave the structure of the function as is. 

It is useful to distinguish `Forward` from other shortcut tags; if the result is former and you have the assurance that the GC hasn't occurred, which you currently, the code in `camlinternalLazy.ml` can read the first field of the `Forward` object and return it as a result. 

As @gadmm, the assertion is a bug. 

My assessment is that, rather than think about tricky corner cases, we may simplify the function such that it just returns `0` if the function succeeds in atomically updating the tag from `Lazy` to `Forcing`, and `1` in all the other cases. This way we get to a state where we can merge multicore MVP trunk, and further development can happen on ocaml/ocaml. Thoughts?

runtime/sys.c
kayceesrk: `check-typo` fixes.

stdlib/lazy.mli
kayceesrk: Good point. Let me undo it.

stdlib/camlinternalLazy.ml
kayceesrk: I'll move it to where it was. `[@poll error]` is not sufficient as it cannot reason about calls to functions IINM. For example, The analysis would encounter the function call `Obj.obj` and report that it is not safe. Is that correct @sadiqj?

stdlib/camlinternalLazy.ml
kayceesrk: We should be using C11 memory model to be reasoning about this as the tag update `update_to_forward` and `Obj.tag` are both external calls implemented in C. `update_to_forward` uses `atomic_compare_and_exchange_strong`, but `Obj.tag`, which calls `caml_obj_tag` uses non-atomic reads. My suggestion is to replace the `Tag_val` in `obj.c:obj_tag` with an acquire load. This should be sufficient to ensure that the message-passing semantics works between the tag update and the field write. @stedolan?

kayceesrk:
Also, thanks @gasche and @gadmm for your high-quality reviews.

runtime/minor_gc.c
kayceesrk: Done.

runtime/obj.c
kayceesrk: Simplified `update_to_forcing` in the latest commits.

stdlib/camlinternalLazy.ml
kayceesrk: Moved.

stdlib/camlinternalLazy.ml
kayceesrk: Updated the tag read to be load acquire. See https://github.com/ocaml-multicore/ocaml-multicore/pull/760/commits/57137194f2be0caec3697a5980d0851763649d34#diff-5dfaf7e6dc66b87cf561b38c634d235d2e5d747c2a9e27df18f290bc223533cfR43-R44.

gasche: I'm a bit nervous about the performance implications of making `Obj.tag` an atomic read all over the place, "just" for the need of lazy values -- I hope it is uncommon to concurrently change a value tag, and it's not clear that we should pay the price for all tag lookups. Wouldn't it be more prudent to expose an `obj_tag_acquire` variant (or `obj_tag_atomic`) and use it?

runtime/obj.c
gasche: The simplified version is very nice, thanks!

stdlib/camlinternalLazy.ml
gasche: Note: `Obj.tag` is already known as a not-so-efficient way to read tags, but currently there is no faster/simpler primitive (so it remains used in all tag-manipulating OCaml code). And whether or not `Obj.tag` is slow, I think that it may be beneficial to force ourselves to express the intent that the block we are accessing may be concurrently mutated.

kayceesrk: Added `caml_obj_tag_atomic` for this. I've unconditionally used this in lazy implementation, as I'd rather err on the side of safety than performance for lazy for MVP.

runtime/obj.c
kayceesrk: It needs one extra check to ensure that the value we are attempting to update the tag of is actually a block. I've added this check in `update_to_forcing`.

kayceesrk:
The failures are due to line number differences in the backtrace tests. I've already fixed these twice in this PR :-| I'll wait until the review on the changes reaches a fixed point, and then I shall fix the failing tests.

stdlib/camlinternalLazy.ml
gadmm: Let's remind ourselves to keep the original comment describing the problem when merging upstream, and to go back to it later.

stdlib/camlinternalLazy.ml
gadmm: Trying on trunk (similar code), the poll-error attribute does not see Obj.tag or Obj.obj as polling, but it does see Obj.field as polling (whether it actually polls or not on trunk, there is something to do about it).

You still need a notation to allow a call to a polling function as suggested here: https://github.com/ocaml/ocaml/pull/10462#issuecomment-868457097 (namely the call to `force_gen_lazy_block`) because there is no obvious way to refactor it (apart maybe by conveying the need to evaluate the thunk using an exception?).

stdlib/camlinternalLazy.ml
gadmm: Don't forget that there is similar code in `lambda/matching.ml` which will need an acquire load as well.

Alternatively, you only need a load/load barrier (e.g. `dmb ishld` on aarch64) before reading the field, which removes the synchronisation cost for short-circuited blocks (though it is likely very small).

Eventually it would be nice to have acquire/release atomics in the OCaml memory model, or barriers to reason about non-atomic locations like here, because it looks complicated to reason about mixing two memory models (and especially that this does not introduce any catch-fire semantics).

Note that since the header is used as an atomic location by the runtime, the tag must always be read from C with at least a `relaxed` atomic read (or else it can catch fire).

gadmm: To clarify, my comment about mixing memory models concerns the inlined forcing code at `matching.ml`.

Acquire/release fences are the most natural in my alternative solution at https://gitlab.com/gadmm/stdlib-experiment/-/blob/master/other/sync_lazy.ml (which also avoids mutating the tag concurrently); and an OCaml version of a release fence might only need a `dmb st` fence.

kayceesrk: I've already put in the inlined fixes for `lambda/matching.ml`. See https://github.com/ocaml-multicore/ocaml-multicore/pull/760/files#diff-059d6544bdc63e5f278818ea9f50bef7aa3f855c4d5ce6fd726abee18b187f55R1860-R1861. Am I missing something?

kayceesrk: I'm not comfortable using raw fences in the OCaml. Our PLDI'18 paper does not include them, and we'd have to extend the work to talk about the guarantees, and in particular, preserving LDRF and their interaction with the rest of the memory model. I wouldn't think of adding fences lightly. 

I would like to aim for a working implementation in this PR, even if it is not the most efficient, and save the optimisations for after multicore MVP is merged with OCaml.

gadmm: To be more precise, `inline_lazy_force_switch` in `matching.ml` does not use `prim_obj_tag` so I assume that it makes an unsynchronized read. This is the one used to compile native code (for efficiency). Possible solutions:
1. replace it with a call to `inline_lazy_force_cond` (remove optimisation)
2. add an argument to Lswitch to read-acquire the tag (heavy solution with a single use-case)
3. add a load-load fence before reading the field (only minimal working solution of a fence for internal use needed in a first time, more carefully design can come later if programmers need it)
4. use my variant above that stores the forward bit in the thunk field rather than with the tag, and atomically read this field.

If you do not want a fence, then 1. and 4. are the simplest I see (and we could do 1. first and experiment with 4. later).

gadmm: I understand the “future work” paragraph p.14 of your paper to refer to adding acquire/release primitives in the future, where the mentioned Kang et al. paper considers fences (IIANM), so I was exploring this possibility. But whatever suitable acquire/release primitives would be better than SC atomics in the current use-case. (Any idea @stedolan?)

Anyway my idea was not to suggest that doing this work is any urgent, but that for now it would be useful to have something equivalent to what you can already do in the C runtime but without incurring a C call, reserved for internal use. I think that the LDRF property is trivially preserved if you see acquire/release atomics as non-atomic, and fences as noops (of course refining the LDRF property and proving that the extended memory model has nice properties is substantially more work).

gasche:
My high-level impression is that we should try to merge a "good first part" version of this PR, that simplifies the current implementation and does not introduce any new bugs, and then have a separate, second discussion about how to safely read concurrently-modified tags. (I suspect we may not be out of surprises there: many other parts of the runtime read tags, how safe are they with respect to concurrent mutation?)

stdlib/camlinternalLazy.ml
xavierleroy: > Added caml_obj_tag_atomic for this.

Apologies for yet another dissenting opinion, but I don't like the idea of having multiple versions of C functions that differ only in the memory model used.

First, it's a software engineering nightmare: code duplication + tiny changes...  We already have enough smelly C code to maintain, thank you very much for not adding more.

Second, the performance degradation that @gasche is worried about has not been demonstrated.  It will be null on x86 (all loads have acquire behaviors) and (in my guess) negligible on other platforms, since `caml_obj_tag` is quite expensive already.

Third, I think the C primitives should be made to fit the Multicore OCaml memory model, and not the other way around.  Meaning that if a C11 nonatomic load in a C primitive is a source of concern, it makes perfect sense to make this load more memory coherent for all uses.  And perhaps other primitives from Obj should be scrutinized from this angle too.

gasche: Here is a proposal for `inline_lazy_force_switch`:

```diff
diff --git a/lambda/matching.ml b/lambda/matching.ml
index 985f86e25..d996489d2 100644
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -1943,12 +1943,12 @@ let inline_lazy_force_switch arg loc =
         ( Lprim (Pisint, [ varg ], loc),
           varg,
           Lswitch
-            ( varg,
-              { sw_numconsts = 0;
-                sw_consts = [];
-                sw_numblocks = 256;
+            ( Lprim (Pccall prim_obj_tag, [ varg ], loc),
+              { sw_numblocks = 0;
+                sw_blocks = [];
+                sw_numconsts = 256;
                 (* PR#6033 - tag ranges from 0 to 255 *)
-                sw_blocks =
+                sw_consts =
                   [ (Obj.forward_tag, Lprim (Pfield(0, Pointer, Mutable),
                                              [ varg ], loc));
```

Unfortunately, the compiler generates not-so-good code in this case (it handles poorly a sparse switch with only integer constants). I have some matching-specific ideas to improve it, but it should be discussed with matching experts and is somewhat irrelevant to the current PR discussion. I believe that this version is better than reverting to `inline_lazy_force_cond`, so it would be a good first step for that function.

gasche:
To continue on my previous line of thought: I think we could cut the current PR to the  commit before "Add new primitive to atomically read tag", and get this merged right away, and then have a dedicated discussion on the memory-safety to make sure that we properly address the issue.

(@kayceesrk : just checking that you know about `make promote DIR=...` to auto-fix line numbers and in generate test outputs in the testsuite.)

gadmm:
I think we're there right after changing `lazy_..._switch` to `lazy_..._cond` or to what @gasche proposed, and changing obj_tag to address @xavierleroy's concern. It's an easy fix so no need to delay the last part of this PR.

stdlib/camlinternalLazy.ml
gasche: @xavierleroy I'm nervous because `Obj.tag`-like functionality is used for pattern-matching, so actually quite heavily used in OCaml programs. Something smells bad about the idea that the dirty tag tricks we use for lazy thunks spreads overhead into all OCaml programs, even those that never use Lazy. (Maybe the cost is neglectible but, if we are going to pay it, we should measure it!)

More importantly, we should think hard about whether we can actually safely mutate tags in this way for lazy thunks, and which parts of the runtime are affected. Adding a barrier to `Obj.tag` is not going to make the problem go away. Having two different primitives, and forcing us to think about which one to use where, is a way to record our decisions/understanding about this problem. Maybe in the end we would find out that we only use the atomic version, and we can simplify again, but in the meantime we should have a clear way to tell which parts of the codebase have to account for concurrent tag modifications, and which do not.

xavierleroy: > I'm nervous because Obj.tag-like functionality is used for pattern-matching

Can you be more precise?  Does the pattern-matching compiler generate calls to `caml_obj_tag` or similar C primitives?  The "get tag" operations produced to implement Lswitch, etc, are under the control of the OCaml compiler, so they obey the Multicore OCaml memory model.

gasche: I reviewed the use of `Obj.tag` in the compiler codebase; @xavierleroy is correct (of course!) that there is mostly nothing performance-sensitive there, and that adding an atomic load to all `Obj.tag` users would be fine. The use-cases of `caml_obj_tag`  in `lambda/matching.ml` come from the compilation of `lazy` patterns, and I guess it's okay if those suffer from overhead. The compilation of non-lazy patterns uses a different `get_tag` operation.

> The "get tag" operations produced to implement Lswitch, etc, are under the control of the OCaml compiler, so they obey the Multicore OCaml memory model.

Right but, if I understand correctly, as @gadmm pointed out the Multicore OCaml memory model says little about what you get if you non-atomically read a field (as `get_tag` does) that was atomically-written to by a Lazy mutation (as `Lazy.force` does). My understanding is that calling `get_tag` to learn that a block has one of the lazy tags (Lazy, Forcing, Forward) *would* be racy and not guarantee anything about the block arguments. We are safe because we never do this -- and my unboxed-constructors work is also safe, as `Lazy.t` is known to have any shape due to Forward shortcutting.

gasche:
What about the lazy-tag checks in other parts of the runtime?

- compare.c, hash.c: I think they need an atomic read to handle Forward_tag safely
- extern.c: this logic can run concurrently with another mutator, but iirc. we decided that it is unspecified what happens in this case (but no crashes); I think the current code will not crash if it reads the tags wrong, at worst it will incorrectly shortcut a pointer in the serialized result. No change seems required.
- finalise.c: no change is required
- minor_gc.c, major_gc.c, weak.c: no change required as iirc. no mutators can run concurrently with the GC

bluddy:
> compare.c, hash.c: I think they need an atomic read to handle Forward_tag safely

Why? Worst case you get the wrong result (which is fine), but do you get a crash?

gasche:
I think you are correct: we can get garbage in this case, but no crash, because the block arguments of a lazy value remains well-formed at any point. (With the new policy, specified in the current PR, that lazy values are not thread-safe, this is fine. Previously we tried to make them shareable, as long as only there is no racy force.)

kayceesrk:
Based on the recommendations, I have 

1. Removed the new primitive to atomically read tag and made `obj_tag` load the tag using acquire.
2. Incorporated @gasche's suggestion for `inline_lazy_force_switch`

>  just checking that you know about make promote DIR=... to auto-fix line numbers and in generate test outputs in the testsuite.

I did not know this. Thanks for pointing this out.

stdlib/camlinternalLazy.ml
gadmm: It is true that the memory model does not say what happens when treating an atomic location as non-atomic (maybe the model can trivially be extended to say what happens in this case), but I was more pointing out that would be racy anyway (and the alternative of a strong atomic load is ~not nice~ totally worth exploring).

runtime/obj.c
```diff
@@ -33,12 +33,15 @@
...
 {
+  header_t hd;
+
   if (Is_long (arg)){
     return 1000;   /* int_tag */
   }else if ((long) arg & (sizeof (value) - 1)){
     return 1002;   /* unaligned_tag */
   }else{
-    return Tag_val(arg);
+    hd = (header_t)atomic_load_acq(Hp_atomic_val(arg));
```
gadmm, line 46:
```suggestion
    /* The acquire load ensures that reading the field of a Forward_tag
       block has appropriate synchronization. */
    hd = (header_t)atomic_load_acq(Hp_atomic_val(arg));
```

gadmm (approved):
LGTM. I would just add a comment (below).

gasche:
(Note: my inline_lazy_switch suggestion requires massaging from the matching+backend people; if we merge as-is, as I think is reasonable, I will open an issue to track this.)

gadmm:
> (Note: my inline_lazy_switch suggestion requires massaging from the matching+backend people; if we merge as-is, as I think is reasonable, I will open an issue to track this.)

Or that could be used as an incentive to explore an implementation that does not mutate tags!

runtime/obj.c
kayceesrk: Good point. Added comment in https://github.com/ocaml-multicore/ocaml-multicore/pull/760/commits/c45feb0653bf56adac0ccd5568b035b87b13d47b that also says where the field read occurs.

kayceesrk:
Will merge when the CI turns green. Thanks @gasche, @gadmm and @xavierleroy.

Issue #742, opened 2021-11-16
Minor tasks from asynchronous review


kayceesrk:
This is a list of minor tasks from the asynchronous review phase. Major tasks will have their own issues. The multicore team members should feel free to start picking up these tasks and working on it. Please put your GitHub handle beside the task if you've picked up a task to avoid duplicating the work.

- [x] Please move runtime/gen_sizeclasses.ml to tools/ or remove it entirely.  As far as I could tell, it's not used in normal builds of the system.
- [x] The function build_frame_descriptors should be static or should be renamed caml_build_frame_descriptors.
- [x] For runtime/lf_skiplist.c I trust Luc's extensive review.  Just one cosmetic suggestion:
```
    (see also William Pugh, "Skip lists: a probabilistic alternative to
    balanced binary trees", Comm. ACM 33(6), 1990). */
```
Replace with
```
    A sequential implementation of skip lists is in file skiplist.c and is based on the paper by
    William Pugh, "Skip lists: a probabilistic alternative to
    balanced binary trees", Comm. ACM 33(6), 1990). */
```
- [x] caml_extern_state could be allocated on first use of the marshaling functions in a domain (and then stored in Caml_state).  This would make domain creation faster in the common case where a domain will not use the marshaler.
- [x] runtime/frame_descriptors.[ch] need copyright headers.
- [x] The function build_frame_descriptors should be static or should be renamed caml_build_frame_descriptors.
- [x]  indentation in caml_process_pending_signals_exn needs fixing.
- [x] sync.c  needs copyright header
- [x] sync.c line 191  line too long

damiendoligez:
- [x] Revisit the capping of work-to-do. https://github.com/ocaml-multicore/ocaml-multicore/blob/a3832eeaae08ec103d2b4377b5624c081e2782c2/runtime/major_gc.c#L411 https://github.com/ocaml-multicore/ocaml-multicore/blob/a3832eeaae08ec103d2b4377b5624c081e2782c2/runtime/major_gc.c#L420 In the sequential GC, the work was capped because you can't do more than 1/3 of a GC cycle in one slice. This turned out to cause problems. Here, you have an accumulated work-to-do, which is not subject to the constraint, so I think it's better not to cap it (never mind doing it twice...). I think it would be better to cap `computed_work` here: https://github.com/ocaml-multicore/ocaml-multicore/blob/a3832eeaae08ec103d2b4377b5624c081e2782c2/runtime/major_gc.c#L471
[...]

PR #766, opened 2021-12-01
Signals changes from sync review and WG3

sadiqj:
This PR implements the three changes coming out of @xavierleroy 's review of the signal code and the comments from the runtime working group:

1. We now block all signals before spawning a domain and then unblock afterwards immediately and when safe to do so for the parent and child respectively. This is https://github.com/ocaml-multicore/ocaml-multicore/commit/3b1db34d0f84ac44d5382bcb638f1e482cc4c7f5
2. We have removed the `total_signals_pending`. This makes some things a little more expensive but the intention is we could potentially optimise these with bitvector tricks or suchlike at a later stage. This is https://github.com/ocaml-multicore/ocaml-multicore/commit/7b389fa4b980732825d4b0fc67b9a18b4190934e
3. We now coalesce signals by signal number. This is https://github.com/ocaml-multicore/ocaml-multicore/commit/5dc58d93ac82021f6771d9cfb9725f8fe46302f7
4. Suspicious indentation in `caml_process_pending_signals_exn` is fixed. This is https://github.com/ocaml-multicore/ocaml-multicore/commit/8de958914bea7188b9a56a158a2762f81f518305 (https://github.com/ocaml-multicore/ocaml-multicore/issues/742)

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -590,6 +590,10 @@ struct domain_startup_params {
   struct domain_ml_values* ml_values;
   dom_internal* newdom;
   uintnat unique_id;
+  #ifdef POSIX_SIGNALS
+  /* signal mask to set after it is safe to do so */
+  sigset_t mask;
+  #endif
 };
 
 static void* backup_thread_func(void* v)
@@ -661,6 +665,9 @@ static void* backup_thread_func(void* v)
 static void install_backup_thread (dom_internal* di)
 {
   int err;
+  #ifdef POSIX_SIGNALS
+  sigset_t mask, old_mask;
+  #endif
 
   if (di->backup_thread_running == 0) {
     CAMLassert (di->backup_thread_msg == BT_INIT || /* Using fresh domain */
@@ -673,8 +680,18 @@ static void install_backup_thread (dom_internal* di)
       caml_plat_lock (&di->domain_lock);
     }
 
+    #ifdef POSIX_SIGNALS
+    /* No signals on the backup thread */
+    sigfillset(&mask);
+    pthread_sigmask(SIG_BLOCK, &mask, &old_mask);
+    #endif
+
     atomic_store_rel(&di->backup_thread_msg, BT_ENTERING_OCAML);
     err = pthread_create(&di->backup_thread, 0, backup_thread_func, (void*)di);
+
+    #ifdef POSIX_SIGNALS
+    pthread_sigmask(SIG_SETMASK, &old_mask, NULL);
+    #endif
     if (err)
       caml_failwith("failed to create domain backup thread");
     di->backup_thread_running = 1;
@@ -733,6 +750,10 @@ static void* domain_thread_func(void* v)
 
   if (domain_self) {
     install_backup_thread(domain_self);
+    #ifdef POSIX_SIGNALS
+    /* it is now safe for us to handle signals */
+    pthread_sigmask(SIG_SETMASK, &p->mask, NULL);
+    #endif
     caml_gc_log("Domain starting (unique_id = %"ARCH_INTNAT_PRINTF_FORMAT"u)",
                 domain_self->interruptor.unique_id);
     caml_domain_set_name(T("Domain"));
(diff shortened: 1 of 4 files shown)


runtime/domain.c
```diff
@@ -760,7 +784,21 @@ CAMLprim value caml_domain_spawn(value callback, value mutex)
...
+#ifdef POSIX_SIGNALS
+/* we block all signals while we spawn the new domain, this is because
+   pthread_create inherits the current signals set and we want to avoid
+   a signal handler being triggered in the new domain before
+   the domain_state is fully populated. */
+  sigfillset(&mask);
+  pthread_sigmask(SIG_BLOCK, &mask, &old_mask);
+  p.mask = old_mask;
+#endif
   err = pthread_create(&th, 0, domain_thread_func, (void*)&p);
```
xavierleroy: It's sad to see so many  `#ifdef POSIX_SIGNALS`.  If you have POSIX threads, you have POSIX signals.  Since you're calling `pthread_create` directly, you have POSIX threads...

The only platform that does not have POSIX_SIGNALS is Win32.  On the one hand it's good to make provisions for a future Win32 port.  On the other hand a lot of the current code assumes POSIX threads and doesn't work with Win32.

runtime/signals.c
```diff
@@ -94,16 +86,13 @@ if (!check_for_pending_signals())
     if( specific_signal_pending > 0 ) {
       if( !atomic_compare_exchange_strong(
             &caml_pending_signals[i],
-             &specific_signal_pending, specific_signal_pending - 1) ) {
+             &specific_signal_pending, 0) ) {
         /* We failed our CAS because another thread beat us to processing
            this signal. Try again to see if there are more of this signal
            to process. */
         goto again;
       }
```
xavierleroy, line 94: You might be able to simplify the whole `again: ... goto again` by using a single atomic_exchange, to read the old value and `caml_pending_signals[i]` and set it to 0 atomically.  On the other hand, the current code might be more efficient if `caml_pending_signals[i]` is zero most of the time, which it should be in practice.  Your choice.

xavierleroy (commented):
The switch to Boolean-valued signals and the removal of `total_signals_pending` look good to me.  I need to re-read the signal-blocking material in context, because I got lost just reading the diff.

runtime/domain.c
sadiqj: Yes, this is a good point. So much of `domain.c` will have to change for Win32 that it's better to just assume POSIX_SIGNALS. I have removed the ifdefs.

runtime/signals.c
sadiqj: Exchange does simplify the code quite nicely but having a ponder I think CAS is better here. I worry about multiple sockets playing cache line ping-pong as a few hundred threads come out of a stop-the-world pause and all race to deliver a waiting signal.

abbysmal:
The changes looks good to me, there's however one omission I believe: `caml_thread_new` should be masking during thread creation as well. (in `st_stubs.c`).

Otherwise everything looks good to me.

sadiqj:
> The changes looks good to me, there's however one omission I believe: `caml_thread_new` should be masking during thread creation as well. (in `st_stubs.c`).

Good catch. Will fix that.

sadiqj:
@Engil this should block for `caml_thread_new` as well now. When we add win32 support we should consider moving the logic to `st_create_thread`.

abbysmal:
LGTM now, thanks.
@xavierleroy you mentioned wanting to re-read the signal-blocking material in context, should we merge this PR as in (and have a wider look at it during merge) or wait for further review from your part?

abbysmal (approved):

runtime/signals.c
```diff
@@ -129,9 +118,6 @@ CAMLexport void caml_record_signal(int signal_number)
   atomic_fetch_add_explicit
     (&caml_pending_signals[signal_number], 1, memory_order_seq_cst);
```
xavierleroy: As far as I can see, it's enough to atomically store 1 in `caml_pending_signals[signal_number]`.  Plus, this avoids the highly theoretical risk of overflow and wrapping around to 0...

xavierleroy (approved):
I downloaded the branch and read the changes again but with more context this time.  It all looks good to me.  Just a suggestion for one more simplification below.

runtime/signals.c
sadiqj: Yes, this makes sense. Changing.

abbysmal:
Everything seems fine now, thank you everyone for your reviews.
Merging.

PR #767, opened 2021-12-01
relaxed -> acquire in minor_gc header read

sadiqj:
relaxed -> acquire in minor_gc header read because we read the forwarding pointer straight afterward.

This reimplements https://github.com/ocaml-multicore/ocaml-multicore/pull/365 but against 5.00


ctk21:
LGTM, merging

PR #768, opened 2021-12-03
Make intern not invoke the GC

kayceesrk:
This PR brings the intern implementation closer to the implementation of trunk OCaml. In particular, intern no longer triggers GC. 

If the unmarshaled object fits within the minor heap, then just like trunk OCaml, this PR allocates a single object that is then carved out for individual allocations. If the unmarshalled object does not fit within the minor heap, trunk OCaml allocates a block, which is then carved out for individual object allocation, which is finally attached to the list of heap blocks. Multicore cannot do this since the allocator uses size-segmented free list. Instead, each allocation goes to the major heap individually. 

Here is the performance result on a [simple binary-tree benchmark](https://gist.github.com/kayceesrk/bf2c5b373da15ac747927924e4b27389). 

N | OCaml trunk | This PR | Slowdown
-- | -- | -- | --
2 | 1.20E-07 | 1.20E-07 | 0.00%
4 | 3.10E-07 | 3.20E-07 | 3.23%
8 | 9.10E-06 | 1.40E-05 | 53.85%
16 | 2.60E-03 | 3.90E-03 | 50.00%
20 | 4.60E-02 | 6.40E-02 | 39.13%
22 | 2.20E-01 | 2.70E-01 | 22.73%
24 | 1.10E+00 | 1.20E+00 | 9.09%
25 | 1.90E+00 | 2.10E+00 | 10.53%


kayceesrk:
@xavierleroy I would appreciate your review on this.

gadmm:
Regarding performance, you could do something a bit similar to what trunk does, as follows: for each block size that is allocated a lot, allocate an empty pool, and allocate from it initially using bump-pointer allocation. Once you are done, make it look like a normal pool and store it with the other pools.

However I think this is not necessarily related to unmarshalling. The bump-pointer optimisation is a known technique in allocators. [Leijen, Zorn and De Moura (§2.2)](https://www.microsoft.com/en-us/research/publication/mimalloc-free-list-sharding-in-action/) say bad things about it (a (small) performance degradation attributed to bad branch prediction). However I think that their program characteristics and OCaml's are very different: the major GC allocates and deallocates in large batches, which is likely to show better branch prediction, and there are ideas worth exploring to make deallocation bump-pointer-optimisation-friendly since it is done by sweeping. Have you already experimented with the bump-pointer optimisation?

Something else important to measure is the impact of cache locality for the imported values (e.g. for workflows such as Coq which uses unmarshalling to load object files). It is possible that it is more advantageous to allocate new pools than filling up old fragmented pools.

(post-MVP ideas obviously!)

xavierleroy:
The description and performance figures look quite nice!  Will review ASAP, I have a bit of a backlog right now.

kayceesrk:
@gadmm We have not experimented with bump-pointer allocation in the major heap. It is not immediately clear to me how to determine "allocated a lot", which depends on the nature of the program. This is something for post-5.00.

runtime/domain.c
```diff
@@ -1272,6 +1266,7 @@ static void domain_terminate()
   // run the domain termination hook
   caml_domain_stop_hook();
   caml_stat_free(domain_state->ephe_info);
+  caml_free_intern_state();
   caml_free_extern_state();
   caml_free_intern_state(domain_state->intern_state);
```
xavierleroy: `caml_free_intern_state` is called  twice with different arities...  This can't be right!

runtime/caml/intext.h
```diff
@@ -109,8 +109,7 @@ void caml_output_val (struct channel * chan, value v, value flags);
   /* Output [v] with flags [flags] on the channel [chan]. */
 
 struct caml_intern_state;
-struct caml_intern_state* caml_alloc_intern_state (void);
-void caml_free_intern_state (struct caml_intern_state*);
+void caml_free_intern_state ();
```
xavierleroy: Please declare as
```
void caml_free_intern_state (void);
```
so that the C compiler can shout at you if you call it with arguments...

runtime/caml/intext.h
```diff
@@ -144,7 +143,7 @@ CAMLextern value caml_input_value_from_malloc(char * data, intnat ofs);
      to the beginning of the buffer, and [ofs] is the offset of the
      beginning of the externed data in this buffer.  The buffer is
      deallocated with [free] on return, or if an exception is raised. */
-CAMLextern value caml_input_value_from_block(const char * data, intnat len);
+CAMLextern value caml_input_value_from_block(char * data, intnat len);
```
xavierleroy: Why the removal of `const`?  `caml_input_value_from_block` is not supposed to change the bytes pointed to by the `data` argument.

runtime/meta.c
```diff
@@ -46,7 +46,7 @@ CAMLprim value caml_get_global_data(value unit)
 CAMLprim value caml_get_section_table(value unit)
 {
   if (caml_params->section_table == NULL) caml_raise_not_found();
-  return caml_input_value_from_block(caml_params->section_table,
+  return caml_input_value_from_block((char*)caml_params->section_table,
```
xavierleroy: Likewise, `caml_params->section_table` is `const` and should not be made non-const, it's the new declaration for `caml_input_value_from_block` that is wrong.

xavierleroy (commented):
Some minor remarks.  I have to go now but will review runtime/intern.c later today.

runtime/intern.c
```diff
@@ -239,446 +284,395 @@ static void readfloats(struct caml_intern_state* s,
...
-#define INTERN_STACK_MAX_SIZE (1024*1024*100)
-typedef value intern_stack_item[STACK_NFIELDS];
-struct intern_stack {
-  struct caml__roots_block caml__roots_stack;
+static void intern_stack_overflow(struct caml_intern_state* s)
+{
+  caml_gc_message (0x04, "Stack overflow in un-marshaling value\n");
+  intern_free_stack(s);
+  caml_raise_out_of_memory();
+}
```
xavierleroy: This code is essentially the same that's in OCaml trunk today.  Yet, I wonder if `intern_free_stack(s)` should not be `intern_cleanup(s)`, so that the obj table and possibly the input block are freed too.  Then the odd comment `Same, then raise Out_of_memory` should be removed.

runtime/intern.c
```diff
@@ -239,446 +284,395 @@ static void readfloats(struct caml_intern_state* s,
...
+    CAMLassert ((value*)s->intern_dest >= d->young_start &&
+                (value*)s->intern_dest < d->young_end);
+    p = s->intern_dest;
+    *s->intern_dest = Make_header (wosize, tag, 0);
+    s->intern_dest += 1 + wosize;
+  } else {
+    p = caml_shared_try_alloc(d->shared_heap, wosize, tag, 0 /* not pinned */);
+    d->allocated_words += Whsize_wosize(wosize);
+    if (p == NULL) {
+      caml_raise_out_of_memory();
```
xavierleroy, line 388: We should call `intern_cleanup(s)` before raising, right?

runtime/intern.c
```diff
@@ -239,446 +284,395 @@ static void readfloats(struct caml_intern_state* s,
...
+        /* Read a value to *dest, then offset *dest by ofs */
+        PushItem(s);
+        sp->dest = dest;
+        sp->op = OShift;
+        sp->arg = ofs;
+        ReadItems(s, dest, 1);
+        continue;  /* with next iteration of main loop, skipping *dest = v */
+      case CODE_CUSTOM:
+      case CODE_CUSTOM_LEN:
+      case CODE_CUSTOM_FIXED: {
```
xavierleroy: In trunk there's code marked "deprecated" that handles the `CODE_CUSTOM` case specially.  I don't see this code here.  Either we handle `CODE_CUSTOM` or we don't...

runtime/intern.c
```diff
@@ -239,446 +284,395 @@ static void readfloats(struct caml_intern_state* s,
...
-      break;
-    default:
-      CAMLassert(0);
     }
   }
-  stack_free(&S);
-  s->stack = NULL;
-  CAMLreturn(result);
+  /* end of case OReadItems */
+  *dest = v;
```
xavierleroy, line 665: Just to make sure I understand: the direct assignment `*dest = v;` is safe because
- either `dest` points within the minor heap for the current domain;
- or `dest` points within a major-heap block freshly allocated by `caml_shared_try_alloc` but not yet visible to the GC, and if `v` is a block it's in the major heap too, so no back-reference is created.

runtime/intern.c
```diff
@@ -239,446 +284,395 @@ static void readfloats(struct caml_intern_state* s,
...
+  do {                                                                  \
+    if (_n > 0) {                                                       \
+      PushItem(s);                                                      \
+      sp->op = OReadItems;                                              \
+      sp->dest = _dest;                                                 \
+      sp->arg = _n;                                                     \
+    }                                                                   \
+  } while(0)
+
+static void intern_alloc(struct caml_intern_state* s,
```
xavierleroy: Perhaps `intern_alloc` should be renamed `intern_alloc_workspace` or `intern_alloc_storage` to better distinguish it from the new `intern_alloc_obj` function?

runtime/intern.c
```diff
@@ -239,446 +284,395 @@ static void readfloats(struct caml_intern_state* s,
...
-static void stack_push(struct intern_stack* s, value v,
-                       int field, int op, intnat arg) {
-  if (Is_block(v)) {
-    CAMLassert(field < Wosize_hd(Hd_val(v)));
-  }
-  if (s->sp == s->len - 1) {
-    stack_realloc(s, v);
+  if (wosize <= Max_young_wosize && wosize != 0) {
+    Alloc_small(v, wosize, String_tag,
+                { caml_handle_gc_interrupt_no_async_exceptions(); } );
```
xavierleroy: I bite: why not calling `caml_alloc_small` ?  

Answering my own question: because `caml_alloc_small` is `caml_alloc` in Multicore OCaml, so we'd pay a bit extra?

Replying to own answer: however, with String_tag, `caml_alloc` doesn't initialize the object, so it's quite cheap indeed.

At any rate, it's OK to use the `Alloc_small` macro.

xavierleroy (commented):
Note to other reviewers: the diff for runtime/intern.c is hard to read, I found it easier to diff the new intern.c in this PR and the one in OCaml 4 trunk.

At any rate, the new intern.c looks very good to me.  It's kind of the minimal adaptation of the one in OCaml 4 that makes it compatible with the Multicore OCaml major heap structure.  

I think I spotted one missing `intern_cleanup` call.  The other comments are just suggestions or me mumbling to myself.

runtime/intern.c
kayceesrk: That is correct. I'll add a comment in the code. And using `caml_modify` would be incorrect since `*dest` is uninitialised memory, and the deletion barrier in `caml_modify` shouldn't be allowed to mark the old value.

runtime/intern.c
kayceesrk: Yes. That's correct. Fixing..

runtime/intern.c
kayceesrk: Good point. It makes sense to call `intern_cleanup` here.

runtime/intern.c
kayceesrk: Removed `CODE_CUSTOM` which we don't handle anymore.

runtime/intern.c
kayceesrk: It seems cleaner to call `caml_alloc_small`. Fixing..

runtime/caml/intext.h
kayceesrk: Fixed..

runtime/meta.c
kayceesrk: Fixed.

xavierleroy (approved):
Looks good to me.

I understand why the old `CODE_CUSTOM` encoding, with the size left implicit, cannot be supported at all in this new implementation of the unmarshaler.  Consequently, one more suggestion: in runtime/caml/intext.h, replace
```
#define CODE_CUSTOM 0x12 /* deprecated */
```
by
```
/* #define CODE_CUSTOM 0x12  -- no longer supported */
```

xavierleroy:
Re: the use of `caml_alloc_small`, see also #786 .

kayceesrk:
I've commented out `CODE_CUSTOM` as suggested. Thanks for the thorough review @xavierleroy.

PR #769, opened 2021-12-03
move frame descriptors header and fix typos

sadiqj:
This PR moves the frame descriptors header from `runtime` to `runtime/caml` and ifdefs with `CAML_INTERNALS`. Adds a check for null that could happen if code is compiled without -g. Also a few sacrifices to check-typo.


abbysmal:
LGTM, merging.

PR #776, opened 2021-12-07
Allow Dynlink only on Domain 0.

abbysmal:
As per WG3 synchronous reviews, this PR will prevent any non-Domain 0 from using Dynlink.
This implementation may be a bit heavy handed, I am open to a better approach.

### DIFF
--- a/otherlibs/dynlink/dynlink_common.ml
+++ b/otherlibs/dynlink/dynlink_common.ml
@@ -20,6 +20,14 @@
 
 open! Dynlink_compilerlibs
 
+(* Dynlink is only allowed on the main domain.
+   Entrypoints to public functions should check for this. *)
+let is_dynlink_allowed () =
+  if not (Domain.is_main_domain ()) then
+    failwith "Dynlink can only be called from the main domain."
+  else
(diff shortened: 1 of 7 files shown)


kayceesrk:
Do we need a new primitive? Can we not use `Domain.get_id` and check that it is `0`? `is_main_domain` is essentially doing the same.

kayceesrk (approved):

kayceesrk:
It was agreed that we will retain the new primitive for `is_main_domain` as there will be a change to the domain interface which will make the domain ids an abstract type rather than a private int as they are now.

xavierleroy:
> t was agreed that we will retain the new primitive for is_main_domain as there will be a change to the domain interface which will make the domain ids an abstract type rather than a private int as they are now.

This will be discussed later, but even if domain.mli exports an abstract type, domain.ml can define it as `int` and can define `is_main` as equality with 0.  Fewer C primitives is always better.

abbysmal:
This is indeed a much simpler approach, I will address this in a subsequent PR, thank you for pointing it out.

PR #777, opened 2021-12-09
Make Domain.id abstract

abbysmal:
This PR makes `Domain.id` abstract, as discussed during WG2. (noted as a pre-MVP item)

This PR also take this opportunity to simplify #776's implementation of `Domain.is_main_domain`.

Does the API surrounding `id` sounds reasonable at this point? Should be provide as well an equality story (`Domain.compare_id`)? (since we do not want users to reason about the underlying implementation?)

### DIFF
--- a/testsuite/tests/lazy/lazy3.ml
+++ b/testsuite/tests/lazy/lazy3.ml
@@ -2,8 +2,10 @@
    ocamlopt_flags += " -O3 "
 *)
 
+let get_domain_id () = Domain.self () |> Domain.string_of_id |> int_of_string
+
 let f count =
-  let _n = (Domain.self ():> int) in
+  let _n = get_domain_id () in
   let r = ref 0 in
   for i = 1 to count do
     incr r;
@@ -14,7 +16,7 @@ let main () =
   let l = lazy (f 1_000_000_000) in
   let d1 =
     Domain.spawn (fun () ->
-        let _n = (Domain.self ():> int) in
+        let _n = get_domain_id () in
         Lazy.force l)
   in
   let n2 = Lazy.force l in
--- a/testsuite/tests/lazy/lazy5.ml
+++ b/testsuite/tests/lazy/lazy5.ml
@@ -1,6 +1,9 @@
 (* TEST
    ocamlopt_flags += " -O3 "
 *)
+
+let get_domain_id () = Domain.self () |> Domain.string_of_id |> int_of_string
+
 let rec safe_force l =
   try Lazy.force l with
   | Lazy.Undefined ->
@@ -8,7 +11,7 @@ let rec safe_force l =
       safe_force l
 
 let f count =
-  let _n = (Domain.self ():> int) in
+  let _n = get_domain_id () in
   let r = ref 0 in
   for i = 1 to count do
     incr r;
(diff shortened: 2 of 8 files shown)


abbysmal:
I did forget to address the testsuite, my bad. Will mark this as WIP.

testsuite/tests/lazy/lazy3.ml
```diff
@@ -2,8 +2,10 @@
    ocamlopt_flags += " -O3 "
 *)
 
+let get_domain_id () = Domain.self () |> Domain.string_of_id |> int_of_string
```
kayceesrk, line 5: It looks like this test does not use the domain id. Let's remove this definition and its uses.

testsuite/tests/lazy/lazy5.ml
```diff
@@ -1,14 +1,17 @@
 (* TEST
    ocamlopt_flags += " -O3 "
 *)
+
+let get_domain_id () = Domain.self () |> Domain.string_of_id |> int_of_string
```
kayceesrk, line 5: Remove the function definition and uses.

testsuite/tests/lazy/lazy7.ml
```diff
@@ -4,14 +4,16 @@
 
 let num_domains = 4
 
+let get_domain_id () = Domain.self () |> Domain.string_of_id |> int_of_string
```
kayceesrk, line 7: Remove the function definition and uses.

testsuite/tests/parallel/domain_id.ml
```diff
@@ -12,9 +12,9 @@ let id () = ()
 
 let newdom_id () =
   let d = Domain.spawn id in
-  let n = Domain.get_id d in
+  let n = Domain.get_id d |> string_of_id |> int_of_string in
```
kayceesrk, line 15:
```ocaml
Domain.get_id d |> string_of_id |> int_of_string
```

I'm not comfortable with this pattern. `string_of_id` doesn't guarantee that it returns the stringified version of an integer. If all the uses need this pattern, we really should have `int_of_id` instead of `string_of_id`. We can guarantee uniqueness in the API doc

```
For two different domains that are currently running, the integer form of their
domain ids will be distinct. That is, if [id1] and [id2] the domain ids of the
current running domains, then [not (int_of_id id1 = int_of_id id2)] holds.
```

kayceesrk:
`domain_id.ml` test is not relevant as it is testing a very specific implementation detail of how ids are implemented. I would prefer it if we remove this test.

xavierleroy:
All right, maybe we need to step back and accept that domain IDs are integers, like in the original API.  If we have abstract IDs + a projection to integers, it's very much like having `type id = private int` like in the current API, only more clumsy...

The reason `type id = private int` made me nervous is a (wrong?) parallel with the type `Unix.file_descr`, which must be abstract because it has two very different implementations (POSIX and Win32).  However, a better parallel is with the `Thread` module, where `Thread.t` is abstract (again because the implementation may be very system-dependent) but `Thread.id` returns a concrete integer identifying the thread.  For printing and logging purposes, `Thread.id` could have returned a string instead, but an integer is nicer as an index into data structures, and other similar uses.

So, I'm OK with dropping my suggestion to make `Domain.id` abstract.

kayceesrk:
Thanks @xavierleroy. It makes sense to retain `Domain.id` as an integer.

abbysmal:
I will then close this PR, and import the simplification of `is_main_domain` into a separate branch. (see #792).

PR #780, opened 2021-12-10
Backport of pr770

kayceesrk:
Bring root management of backtrace_last_exn in systhreads closer to trunk. 

There are unrelated formatting edits to pass the hygine check.


kayceesrk:
The macos run got stuck on `threadsigmask.ml` test. This failure is unrelated to this PR, and I suspect that this may have been fixed on the `5.00` branch but not backported. I'm merging this irrespective of this failure.

PR #783, opened 2021-12-10
Cherry pick: Merge pull request #701 from ocaml-multicore/really_flush

bikallem:
Really flush output when pre-defined formatters are used in parallel.


ctk21:
LGTM, merging

PR #786, opened 2021-12-11
Reimplement `caml_alloc_small` like in OCaml 4.x

xavierleroy:
The following idiom is here to stay (it's documented in the OCaml/C API, even though it's in the "low-level interface" section):

    v = caml_alloc_small(sz, tag);
    Field(v, 0) = ... ;
    Field(v, 1) = ... ;

At some point in the development of Multicore OCaml, `caml_alloc_small` was made to just call into `caml_alloc`.

If `sz <= Max_young_wosize`, as it should, everything works fine; it's just a bit less efficient than in OCaml 4.x, owing to the pre-initialization of the fields of the new value.

However, if the programmer made a mistake and `sz` is bigger than `Max_young_wosize`, the OCaml 4.x implementation in debug mode catches the problem with an assertion, while the `caml_alloc` implementation says nothing and returns a block allocated in the major heap.  Then, the direct assignments `Field(v, i) = ...` mess up the GC invariants, but this will not be caught easily.

Hence: the OCaml 4.x implementation of `caml_alloc_small` is preferable, and this PR brings it back.

### DIFF
--- a/runtime/alloc.c
+++ b/runtime/alloc.c
@@ -158,7 +158,15 @@ CAMLexport value caml_alloc_N (mlsize_t wosize, tag_t tag, ...)
 
 CAMLexport value caml_alloc_small (mlsize_t wosize, tag_t tag)
 {
-  return caml_alloc(wosize, tag);
+  value result;
+
+  CAMLassert (wosize > 0);
+  CAMLassert (wosize <= Max_young_wosize);
+  CAMLassert (tag < 256);
+  CAMLassert (tag != Infix_tag);
+  Alloc_small (result, wosize, tag,
+               { caml_handle_gc_interrupt_no_async_exceptions(); });
+  return result;
 }
 
 /* [n] is a number of words (fields) */


kayceesrk (approved):

kayceesrk:
The change looks good to me. 

`caml_alloc_small` was made to call `caml_alloc` in this commit https://github.com/ocaml-multicore/ocaml-multicore/commit/7824259b2e84d00a6faaed190a18cef94d1edc7c. IINM the change was necessary for the earlier concurrent version of the minor GC where the `Field` accesses may trigger a read fault, which may, in turn, trigger a minor GC on the initialising domain (if a remote domain wants to promote an object from the minor heap of the initialising domain). This no longer happens in multicore since we moved to the parallel minor GC. 

Similarly, multicore is initialising the object on `caml_alloc_shr` (https://github.com/ocaml-multicore/ocaml-multicore/blob/5.00/runtime/memory.c#L370-L379) while trunk doesn't https://github.com/ocaml/ocaml/blob/trunk/runtime/memory.c#L499-L506. The initialisation can be omitted. 

CC @ctk21 @sadiq.

ctk21:
Motivation for the change is a good one. I believe the change to be correct and shouldn't be a problem.

kayceesrk:
I'm going to ignore the hygiene check failure as the target branch already has the necessary fixes.

PR #790, opened 2021-12-14
add caml_check_pending_actions, caml_process_pending_actions

sadiqj:
This PR adds `caml_check_pending_actions` and `caml_process_pending_actions` which are part of the C API that is currently missing.

There's also `caml_process_pending_actions_exn` which I will create an issue for in a minute.

### DIFF
--- a/runtime/io.c
+++ b/runtime/io.c
@@ -118,7 +118,7 @@ static void check_pending(struct channel *channel)
     /* Temporarily unlock the channel, to ensure locks are not held
        while any signal handlers (or finalisers, etc) are running */
     Unlock(channel);
-    caml_handle_gc_interrupt();
+    caml_process_pending_actions();
     Lock(channel);
   }
 }
--- a/runtime/caml/domain.h
+++ b/runtime/caml/domain.h
@@ -37,7 +37,6 @@ int caml_reallocate_minor_heap(asize_t);
 
 int caml_incoming_interrupts_queued(void);
 
-int caml_check_pending_actions();
 void caml_handle_gc_interrupt(void);
 void caml_handle_gc_interrupt_no_async_exceptions(void);
 void caml_handle_incoming_interrupts(void);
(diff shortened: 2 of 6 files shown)


runtime/io.c
```diff
@@ -118,7 +118,7 @@ static void check_pending(struct channel *channel)
     /* Temporarily unlock the channel, to ensure locks are not held
        while any signal handlers (or finalisers, etc) are running */
     Unlock(channel);
-    caml_handle_gc_interrupt();
```
damiendoligez, line 121: There's a good number of calls to `caml_handle_gc_interrupt` in the runtime and I'd like to know why you chose to rename this one and not, for example: https://github.com/ocaml-multicore/ocaml-multicore/blob/a0a8583af15f75e974a3eb121feab83189b7fb4f/runtime/signals_nat.c#L92

It would make sense to decide that `caml_process_pending_actions` is a backward-compatibility stub, and thus never use it in the runtime. Or you might want to do a global rename of `caml_handle_gc_interrupt` to `caml_process_pending_actions`. Using two names for the same thing is a bit strange.

damiendoligez (commented):
Looks good except for one point.

runtime/io.c
sadiqj: The original intention was that `caml_process_pending_actions` and `caml_handle_gc_interrupt` do different things, the former is also supposed to process pending signals (but, in a mistake on my part, does not at the moment).

Really, any `caml_handle_gc_interrupt` / `caml_process_pending_signals` pairs should be replaced by `caml_process_pending_actions`.

sadiqj: Latest commit should address this.

I'm still not 100% satisfied because `caml_check_pending_actions` can return false when there is actually a pending action to do (in that there's a signal to process, where the signal handler records just before a store to young_limit). That only leads to delayed delivery of a signal though.

The interaction between signals and GC interrupts needs to be reworked. Right now they feel like two independent mechanisms. The replumbing that will be required to support https://github.com/ocaml-multicore/ocaml-multicore/issues/791 is probably the best place to do it.

kayceesrk:
The outstanding task has been forked off into issue #806. Merging this.

PR #792, opened 2021-12-15
stdlib: simplify is_main_domain

abbysmal:
This is a followup to #777.
It makes `is_main_domain` simpler and remove the underlying primitive.


kayceesrk:
LGTM.

PR #798, opened 2021-12-16
Revert asmgen testsuite and ocamltest to trunk

abbysmal:
This shaves the diff off by a sizeable amount.
Gist of it was: there was a missing stub for `caml_call_realloc_stack` in `asmgen/main.c`, causing the linker to fail when using trunks code.
Multicore version was linking the program properly with the runtime, actually resolving the symbols, when trunk just stubs these out. (as they don't matter for these testcases.)


kayceesrk:
Good catch. LGTM.

PR #800, opened 2021-12-17
document which Gc stats are global and which are per-domain

sadiqj:


ctk21 (commented):
Changes look good. 

Should we also clarify (domain vs whole program) for the other stat functions exposed in `gc.mli`:
 `stat`, `quick_stat`, `counters`, `minor_words`, `print_stat`, `allocated_bytes`

sadiqj:
`stat`/`quick_stat` should be covered by this but it doesn't hurt to re-iterate.

I'll do the others in a sec.

sadiqj:
Good call @ctk21 .

So the other functions in the Gc module have rather .. complicated documentation at the minute. The per-domain counters are actually per-domain _slot_, they don't get reset during spawning. We can't reset them because otherwise stats won't at present. We're going to need a separate set of counters that keep a track of the per-domain counters.

The documentation as it stands reflects what is there right now, I think post-MVP we're going to need to rework the stats collection so that they are a little more useful.

ctk21:
Thanks, good to flush that out. The changes addressed my concerns.

kayceesrk:
The comments capture the state of the world, and I agree that there is room for improvement post-MVP. The comments look good to me.

PR #807, opened 2021-12-17
Make sure variables that are not explicitly initialized duri…

abbysmal:
…ng create_domain or any sub function are initialized there.

### DIFF
--- a/runtime/domain.c
+++ b/runtime/domain.c
@@ -432,9 +432,31 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
       goto alloc_main_stack_failure;
     }
 
+    domain_state->c_stack = NULL;
+    domain_state->exn_handler = NULL;
+
+    domain_state->gc_regs_buckets = NULL;
+    domain_state->gc_regs = NULL;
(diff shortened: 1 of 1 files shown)


runtime/domain.c
```diff
@@ -432,17 +432,40 @@ static void create_domain(uintnat initial_minor_heap_wsize) {
...
     caml_register_generational_global_root(&domain_state->backtrace_last_exn);
 
+    domain_state->compare_unordered = 0;
+    domain_state->oo_next_id_local = 0;
+
+    domain_state->requested_major_slice = 0;
+    domain_state->requested_minor_gc = 0;
+    domain_state->requested_external_interrupt = 0;
+
+    domain_state->parser_trace = Val_bool(0);
```
ctk21:
```suggestion
    domain_state->parser_trace = 0;
```

ctk21: there might be a better way here; I think the way we are using `domain_state->parser_trace` is bugged

kayceesrk:
LGTM. Merging.

PR #808, opened 2021-12-17
signal_alloc testcase fix

abbysmal:
@moyodiallo noticed that #722 failed to actually re-enable the `signals_alloc` testcase, and that it was broken on bytecode.
This likely happened on the recent rebase of 722, and was not caught because the testcase got lost in the midst of the `disabled` file removal.
This PR addresses both issues.

### DIFF
--- a/runtime/interp.c
+++ b/runtime/interp.c
@@ -80,7 +80,7 @@ sp is a local copy of the global variable Caml_state->extern_sp. */
 #define Restore_after_gc \
   { sp = domain_state->current_stack->sp; accu = sp[0]; env = sp[1]; sp += 3; }
 #define Enter_gc \
-  { Setup_for_gc; caml_handle_gc_interrupt(); \
+  { Setup_for_gc; \
     caml_process_pending_actions();         \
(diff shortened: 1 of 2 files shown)


runtime/interp.c
```diff
@@ -81,7 +81,7 @@ sp is a local copy of the global variable Caml_state->extern_sp. */
   { sp = domain_state->current_stack->sp; accu = sp[0]; env = sp[1]; sp += 3; }
 #define Enter_gc \
   { Setup_for_gc; caml_handle_gc_interrupt(); \
-    caml_check_for_pending_signals();         \
+    caml_process_pending_actions();         \
```
sadiqj, line 84: `caml_process_pending_actions` already has a call for to `caml_handle_gc_interrupt` (actions is just gc_interrupt+pending_signals), so probably worth reverting this change or removing the `caml_handle_gc_interrupt`.

sadiqj (changes requested):

sadiqj (approved):

PR #809, opened 2021-12-19
Finish off tools/check-typo on the repo

dra27:
This can be checked commit-by-commit and the last commit can be ignored (or patdiff can be used to limit its noise).

Two things remain:
- I was too lazy to remind myself how to break Python lines properly, so `tools/eventlog_to_latencies.py` and `tools/gdb_ocamlrun.py` still contain over-long lines
- There are `FIXME` comments of @kayceesrk's in `runtime/caml/stack.h` which I figured might want removing, rather than line-breaking?


kayceesrk:
You are correct @dra27. `Callback_link` is no longer used and can be removed. Other architectures should define the macros defined under `amd64`.

ctk21:
Many thanks for these cleanups, appreciated. LGTM, merging!

PR #819, opened 2021-12-24
Do not initialise in caml_alloc_shr

kayceesrk:
Make array.c use non-initialising allocation instead of the initialising caml_alloc.

### DIFF
--- a/runtime/alloc.c
+++ b/runtime/alloc.c
@@ -49,6 +49,9 @@ CAMLexport value caml_alloc (mlsize_t wosize, tag_t tag)
     }
   } else {
     result = caml_alloc_shr (wosize, tag);
+    if (tag < No_scan_tag) {
+      for (i = 0; i < wosize; i++) Field (result, i) = Val_unit;
+    }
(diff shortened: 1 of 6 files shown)


kayceesrk:
The changes are made in such a way that the match trunk. Came out of the review of https://github.com/ocaml/ocaml/pull/10831.

sadiqj (approved):
This looks good and should be a good win for numerical code. (It also matches what `caml_alloc_shr` does on trunk).

I did notice a couple tests (globrootsprim and alloc_async_stubs) that seem to use `caml_alloc_shr` without initialising subsequently that might need to be fixed.

kayceesrk:
Thanks for the review @sadiqj. 

`globrootsprim` use is ok since the tag is `String_tag`. 

In `alloc_async_stubs` (which is currently skipped on multicore), I've changed `caml_alloc_shr(1000, 0)` to `caml_alloc_shr(1000, String_tag)`. This is ok since the result of the allocation is not used.

PR #822, opened 2021-12-31
Revert array.c to trunk. Introduce word-sized memmove.

kayceesrk:
The word-sized `memmove` is only used when there is more than 1 domain running. The commit includes other minor changes.

~The PR assumes that it is running on 64-bit (which is true under multicore), where `sizeof(double) == sizeof(value)`.~

For reviewing the PR, it would be useful to compare the new `array.c` file with the one in trunk rather than one in the PR.

### DIFF
--- a/runtime/array.c
+++ b/runtime/array.c
@@ -155,6 +155,9 @@ static value caml_array_unsafe_set_addr(value array, value index,value newval)
 }
 
 /* [ floatarray -> int -> float -> unit ] */
+/* [MM]: [caml_array_unsafe_set_addr] has a fence for enforcing the OCaml
+   memory model through its use of [caml_modify]. [caml_floatarray_unsafe_set]
+   will also need a similar fence in [Store_double_flat_field]. */
 CAMLprim value caml_floatarray_unsafe_set(value array, value index,value newval)
 {
   intnat idx = Long_val(index);
@@ -191,9 +194,9 @@ CAMLprim value caml_floatarray_create(value len)
     caml_invalid_argument("Float.Array.create");
   else {
     result = caml_alloc_shr (wosize, Double_array_tag);
-    result = caml_check_urgent_gc (result);
   }
-  return result;
+  /* Give the GC a chance to run */
+  return caml_check_urgent_gc (result);
 }
 
 /* [len] is a [value] representing number of words or floats */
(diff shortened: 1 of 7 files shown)


runtime/array.c
```diff
@@ -303,24 +306,81 @@ CAMLprim value caml_make_array(value init)
...
+
 CAMLprim value caml_floatarray_blit(value a1, value ofs1, value a2, value ofs2,
                                     value n)
 {
-  memmove((double *)a2 + Long_val(ofs2),
-          (double *)a1 + Long_val(ofs1),
-          Long_val(n) * sizeof(double));
+  CAMLassert (sizeof(double) == sizeof(value));
+  wo_memmove(&Field(a2, Long_val(ofs2)), &Field(a1, Long_val(ofs1)),
+             Long_val(n));
```
xavierleroy: The address computations are wrong for 32 bit platforms!  Since we have not yet formally ruled them out, please use the same address calculations as in the current trunk, `(double *) a1 + Long_val(ofs1)`.  

More generally: please stop changing things irrelevant to multicore "because you can". it risks introducing new bugs, and it makes reviewing more painful than strictly necessary.

runtime/array.c
```diff
@@ -370,20 +434,41 @@ static value caml_array_gather(intnat num_arrays,
...
     /* Array of values, small enough to fit in young generation.
        We can use memcpy directly. */
     res = caml_alloc_small(size, 0);
     for (i = 0, pos = 0; i < num_arrays; i++) {
-      caml_blit_fields(arrays[i], offsets[i], res, pos, lengths[i]);
+      /* [res] is freshly allocated, and no other domain has a reference to it.
+         Hence, a plain [memcpy] is sufficient. */
+      memcpy(&Field(res, pos),
+             &Field(arrays[i], offsets[i]),
+             lengths[i] * sizeof(value));
```
xavierleroy, line 476: `memcpy` -> `wo_memcpy` ?

runtime/array.c
```diff
@@ -303,24 +306,81 @@ CAMLprim value caml_make_array(value init)
...
+  } else {
+    if (dst < src) {
+      /* copy ascending */
+      for (i = 0; i < nvals; i++) dst[i] = src[i];
+    } else {
+      /* copy descending */
+      for (i = nvals; i > 0; i--) dst[i-1] = src[i-1];
+    }
+  }
+}
```
xavierleroy: A good C compiler might try to use vector instructions to move several words at a time, resulting in mixed-size concurrent accesses...  However, if I read `caml_modify` correctly, you need to use atomic stores of the "release" kind, which should prevent the writes from being coalesced into vector writes.

xavierleroy (commented):
Looks good to me, modulo one point where I prefer the code in the current trunk.

Concerning `memcpy` and size of accesses, note that blit operations on strings and byte arrays may have similar issues, in that `memcpy` will be performed not byte-per-byte but word-per-word as soon as possible.  I suggest we find a way to live with it, and don't try to force byte-per-byte copying, because performance of blitting on byte arrays is crucial to the efficient implementation of string operations.

runtime/array.c
kayceesrk: `memcpy` is ok here since `res` is freshly allocated and no other domain yet has a reference to it. See the comment above.

xavierleroy: Right, sorry for having missed this point.

runtime/array.c
```diff
@@ -308,12 +308,14 @@ CAMLprim value caml_make_array(value init)
...
-
-   [MM]: TODO enforcing the memory model also requires fences. See note in
-   memory.c */
+   time.
+
+   Since the [memmove] implementation does not guarantee that the writes are
+   always word-sized, we explicitly perform word-sized writes of the release
+   kind to avoid mixed-mode accesses. Performing release writes should be
+   sufficient to prevent smart compilers from coalesing the writes into vector
+   writes, and hence prevent mixed-mode accesses. [MM].
```
xavierleroy, line 317: According to the comments in memory.c, you might also need a memory barrier.  Presumably one barrier before the whole loop?

kayceesrk: That's exactly what I was thinking over dinner. Adding a barrier now.  

As an aside, I'm adding `[MM]` in comments to those places in code with changes relevant to the memory model. These will need to be sanity checked for the arm64 backend.

kayceesrk:
The barrier has been added now, and I believe that all the concerns have been addressed. I'll merge this PR later today if there aren't further concerns before then. 

> blit operations on strings and byte arrays may have similar issues, in that memcpy will be performed not byte-per-byte but word-per-word as soon as possible

Agree with not implementing byte-by-byte copying as this alone is not sufficient to prevent mixed-mode access for strings and byte arrays. Even without blit, byte arrays and strings permit mixed-mode access as they allow reading and writing integers into byte arrays and strings. See https://github.com/ocaml-multicore/ocaml-multicore/blob/5.00/stdlib/bytes.mli#L575. 

We may document the fact that with racy access byte arrays and strings won't crash, but the memory model does not provide any guarantees for values observed.

xavierleroy:
OK for merging and for doing nothing special with byte arrays.  For single bytes, mixed-size accesses can be expected to not produce values out of thin air (unlike say a 64-bit float being written with two 32-bit writes in sequence).  

Big arrays is another can of worms...  Just like byte arrays, they are widely used to interface with C code, and said C code doesn't obey the Multicore memory model to begin with, so it feels like a lost battle to try to implement the MMM for byte arrays and bigarrays.

kayceesrk:
Thanks. I'll start working on the documentation post-merge.

runtime/array.c
```diff
@@ -303,24 +306,112 @@ CAMLprim value caml_make_array(value init)
...
+  if (caml_domain_alone ()) {
+    memmove (dst, src, nvals * sizeof(double));
+  } else {
+    /* See memory model [MM] notes in memory.c */
+    atomic_thread_fence(memory_order_acquire);
+    if (dst < src) {
+      /* copy ascending */
+      for (i = 0; i < nvals; i++)
+        atomic_store_explicit(&((atomic_double*)dst)[i], src[i],
+                              memory_order_release);
```
xavierleroy, line 364: Apologies for noticing after this PR was merged, but: this use of `atomic_double` does not work on ARM 32 bits and may not work on other platforms.  On ARM 32 bits, GCC implements `_Atomic double` as a double protected by a mutex.  So, it doesn't have the memory layout of a `double`, and atomic operations call into library functions from the `-latomic` library, which call  POSIX thread mutex operations.

What this means is that the OCaml runtime will not even compile on ARM32 in bytecode-only mode, and if we add `-latomic` the blit operation over float arrays will crash or hang or write garbage.

I think it would be  preferable to leave open the memory model issues involving non-word-sized accesses and discuss them thoroughly later rather than trying quick "fixes" that don't quite work.

kayceesrk: Thanks for noticing this. I agree with not inserting quick fixes for memory model. 

How should we proceed with this now? I can remove `atomic_double` and revert `caml_floatarray_blit` to trunk version with a comment about the mixed-mode accesses? It may be useful to retain the fence before `memmove` since that would anyway be needed to enforce the memory model.

xavierleroy: For the time being, maybe you could keep the `memmove`-based implementation and just add a comment saying "not consistent with the memory model, to be discussed later".

Later, if we agree to keep supporting 32-bit platforms but only in single-domain mode, we could use `wo_memmove` for 64-bit platforms and `memmove` for 32-bit platforms.  But maybe we could finally agree to drop all support for 32-bit platforms...

kayceesrk: To clarify, it is only the `caml_floatarray_blit` that is problematic. `wo_memmove` is not used on float arrays, and IINM is safe on 32-bit platforms. Please correct me if I am wrong. 

> For the time being, maybe you could keep the memmove-based implementation and just add a comment saying "not consistent with the memory model, to be discussed later".

With that, are you suggesting that I revert the `caml_floatarray_blit` implementation to trunk version and add the suggested comment?

xavierleroy: > With that, are you suggesting that I revert the caml_floatarray_blit implementation to trunk version and add the suggested comment?

Yes, that's my suggestion, but feel free to disagree :-) 

My reasoning is that we need to discuss more globally the memory model for byte arrays, flat float arrays, and bigarrays, and agree on what we want before writing code.

kayceesrk: I’ve opened https://github.com/ocaml-multicore/ocaml-multicore/pull/831 that reverts the changes.

PR #825, opened 2022-01-04
stdlib: make Atomic.t covariant

bikallem:
Make type parameter 'a' in 'a Atomic.t covariant as well as injective,
i.e. `+!'a Atomic.t`. Consider a common use case in OCaml where
concurrency libraries are functorised over an IO module:

module type IO = sig
  type +'a t
end

If we were to use 'a Atomic.t as follows,
module Io : IO = struct
  type +'a t = 'a Atomic.t
end

The compilation fails with the following error:

38 |   type +!'a t = 'a Atomic.t
       ^^^^^^^^^^^^^^^^^^^^^^^^^
Error: In this definition, expected parameter variances are not satisfied.
       The 1st type parameter was expected to be injective covariant,
       but it is injective invariant.

Making type parameter 'a' covariant allows the above code to
build/compile successfully.

Signed-off-by: Bikal Lem <gbikal+git@gmail.com>

### DIFF
--- a/stdlib/atomic.mli
+++ b/stdlib/atomic.mli
@@ -27,7 +27,7 @@
     import additional compatibility layers. *)
 
 (** An atomic (mutable) reference to a value of type ['a]. *)
-type !'a t
+type +!'a t
 
 (** Create an atomic reference. *)
 val make : 'a -> 'a t
--- a/stdlib/camlinternalAtomic.ml
+++ b/stdlib/camlinternalAtomic.ml
@@ -16,7 +16,7 @@
 
 (* CamlinternalAtomic is a dependency of Stdlib, so it is compiled with
    -nopervasives. *)
-type !'a t
+type +!'a t
 
 (* Atomic is a dependency of Stdlib, so it is compiled with
    -nopervasives. *)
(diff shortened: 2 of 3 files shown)


xavierleroy:
This would be very clearly unsound.  Atomic references are invariant, like ordinary references.  Consider:
```
B is a strict subtype of A
a : A
b : B
let r : B ref = ref b in (r :> A ref) := a; !r
```
The result is value `a` with type `B`, which it doesn't have.

Please close this PR.

bikallem:
Thanks for the illustration. Closing this PR.

bikallem:
@xavierleroy I tried your example concretely as below. However, it seems the compiler is clever enough to not allow unsound casting. Specifically, the OCaml compiler infers that `r :> a ref` is invalid. Trying below in OCaml toplevel:
```
type a = [ `A | `B ]
type b = [ `A ]  (* b is a subtype of a *) ;;

let a:a = `B;;
let b:b = `A;;

let r: b ref = ref b;;
(r :> a ref) := a;;
```
The last expression `(r :> a ref)` results in error as below:
```
Error: Type b ref is not a subtype of a ref
       The first variant type does not allow tag(s) `B
```
I am treading at the limits of my comprehension of OCaml type machinery, but shouldn't this assuage us that `+!'a Atomic.t` will not result in unsoundness?

xavierleroy:
> Specifically, the OCaml compiler infers that r :> a ref is invalid.

Yes, because `ref` is invariant, not covariant !  If `ref` was declared covariant, as you proposed  to do with `Atomic.t`, the coercion would be accepted, but it would be unsound.

PR #828, opened 2022-01-05
initialise extern_flags to 0 on extern stat init

avsm:
Without this the value may potentially contain junk values and is subsequently used in reachable_words, which results in
memory corruption. Fixes Obj.reachable_words on platforms which return junk memory from malloc (e.g. FreeBSD/OpenBSD but also potentially some libcs on Linux)

Fixes https://github.com/ocaml-multicore/ocaml-multicore/issues/824

PR #829, opened 2022-01-05
fix pthread name setting on FreeBSD/OpenBSD/NetBSD

avsm:
There is an unfortunate portability mess around setting pthread names, so we now use the various permutations on FreeBSD
and OpenBSD (which require another header include as well) and NetBSD (which is similar to Linux, but with a different error
code return).

The code previously used GNU_SOURCE to detect Linux, which isnt much use as GNU_SOURCE is unconditionally defined at the top of unix.c, so we just assume that the 'default' platform is Linux.  pthread_setname_np is only defined in glibc 2.12 or higher, but looking for that specifically makes it complex to also detect musl, so it's simplest to leave it as the default case.

This (along with #828) lets FreeBSD pass the test suite, and for OpenBSD to build. There is one more patch to follow and OpenBSD will also pass the test suite.

### DIFF
--- a/runtime/unix.c
+++ b/runtime/unix.c
@@ -50,6 +50,9 @@
 #include <mach-o/dyld.h>
 #endif
 #include <pthread.h>
+#if defined(__OpenBSD__) || defined(__FreeBSD__)
+#include <pthread_np.h>
+#endif
 #include "caml/fail.h"
 #include "caml/memory.h"
 #include "caml/misc.h"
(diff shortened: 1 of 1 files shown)


avsm:
I'll defer this one to ocaml/ocaml post merge, as BSD support wasn't in the MVP

avsm:
closed in favour of https://github.com/ocaml/ocaml/pull/10872

PR #830, opened 2022-01-05
remove Int/Long/Bool_field macros

avsm:
These were originally introduced (and were more complex) when we had a read barrier. Now only the Long_field macro is used, so we can reduce the size of the diff to trunk by removing them and just using Long_val(Field()) instead.


kayceesrk:
LGTM.

